PR #66, opened 2015-02-25
Block Tests

winsvega:
Block Rlps with invalid transactions


CJentzsch:
All those tests have an invalid format. If they are valid, the they should have a blockHeader and uncleHeaders object, and when they are invalid, then there should be no transaction object.

PR #71, opened 2015-03-04
Tests Changes To compile without errors

winsvega:
Gas cost fees increased
Some tests are still incorrect and need review


CJentzsch:
I updated the tests from the fillers you pushed to the cpp-repo.

PR #75, opened 2015-03-12
Introduce ECIES crypto test

romanman:


BasicTests/crypto.json
```diff
@@ -1,8 +1,16 @@
-[
-    {
+{
+    "aes1": {
```
subtly: Good to see these tests :)  AES-CTR should probably include an IV in order to confirm the functionality of an IV is available. We omit the IV for streams but I believe it maybe used for ECIES.

romanman:
@subtly  : well yeah I just started this thread by very basic 
examples that go/cpp/java already support we will expand them 
to the moon

PR #79, opened 2015-03-16
Move hash value into Block header structure

romanman:


wanderer:
@romanman thanks but I really need to totally regenerate this test since we have new blockheader fields for POC9

PR #93, opened 2015-07-20
Changed block number to fit uint32

arkpar:


arkpar:
cpp-ethereum currently rejects block numbers that overflow 32 bits

PR #102, opened 2015-11-20
Add Homestead test for OOG when returning a data after contract creation

Nashatyrev:
This is the test I've created for testing Homestead updates (1) & (3) from https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.mediawiki

It tries to create the following Solidity contract, but has the gas limit 1 gas lower than required. Post state checks that no new accounts were created. The block number is set to 666000

```
         contract Test {
            uint a = 256;

            function set(uint s) {
                a = s;
            }

            function get() returns (uint) {
                return a;
            }
         }
```

 Put this to `stExample.json` as a draft

### DIFF
--- a/StateTests/stExample.json
+++ b/StateTests/stExample.json
@@ -61,5 +61,50 @@
             "to" : "095e7baea6a6c7c4c2dfeb977efac326af552d87",
             "value" : "0x0186a0"
         }
+    },
+
+    "homesteadContractCreateOOG": {
+        "env": {
+            "currentCoinbase": "2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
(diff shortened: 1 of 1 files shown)


CJentzsch:
Thanks for contributing.
I think we should have a homestead branch for this. And when homestead is official, I will make develop to Frontier, and homestead to develop.
Could you make commit to a new Homestead branch instead of develop?
Thank you!

Nashatyrev:
Sure, just let me know when the branch is ready and what json file is more appropriate for the test. 
Or you may just copy/paste the test and close this PR.

winsvega:
outdatedd

PR #113, opened 2016-05-20
Update bcInvalidRLPTest.json

NikVolf:
is this some kind of typo?


winsvega:
yes. some of this tests were created manually. 
however to fix this you will need to create a PR into libethereum (I'll managed that)
as tests get refilled from fillers in libethereum
https://github.com/ethereum/libethereum/blob/develop/test/libethereum/BlockchainTestsFiller/bcInvalidRLPTest.json

PR #116, opened 2016-07-20
linted json files

wanderer:
This cleans up the VMtests by adding persistent formating to the storage fields. It also includes the linter.

### DIFF
--- a/fixVMtests.js
+++ b/fixVMtests.js
@@ -1,0 +1,109 @@
+const fs = require('fs')
+const path = require('path')
+const ethUtil = require('ethereumjs-util')
+const BN = ethUtil.BN
+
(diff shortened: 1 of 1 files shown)


winsvega:
this tests are created from fillers 
https://github.com/ethereum/libethereum/tree/develop/test/libevm/VMTestsFiller
we'll have to run this script every time after test being refilled. (on each commit to this test repository)

what was the issue in the first place?

wanderer:
@winsvega because the storage values are not correct. Storage Key and Values are always 32 bytes.

winsvega:
Such format is not comfortable to read. 
We just assume that this values padded to 32 bytes.

PR #120, opened 2016-10-05
Test prefunded contract creations that go OOG

Arachnid:
In this situation, the prefunded balance should remain in the account, not be zeroed out.


winsvega:
added
https://github.com/ethereum/tests/commit/29bedd9c05e0a1cefb9421ef24f32fc93c627945

PR #145, opened 2017-03-08
Rpc tests

winsvega:


RPCTests/modules/async.js
```diff
@@ -0,0 +1,5289 @@
+(function (global, factory) {
```
chriseth: Please don't copy foreign code into our repository, especially not without copyright notice. Isn't there an NPM package that does this?

RPCTests/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "homepage": "",
+  "standard": {
+    "ignore": [
+      "node_modules/*",
+      "build/*",
+      "test/resources/*"
+    ],
+    "parser": "babel-eslint"
+  },
+  "browserify": {
```
yann300: we can delete this, (this tool is not meant to be run on the browser)

RPCTests/package.json
```diff
@@ -0,0 +1,33 @@
...
+    
+  },
+  "homepage": "",
+  "standard": {
+    "ignore": [
+      "node_modules/*",
+      "build/*",
+      "test/resources/*"
+    ],
+    "parser": "babel-eslint"
```
yann300: same as below

RPCTests/scripts/addPeer.js
```diff
@@ -0,0 +1 @@
+web3.admin.addPeer("enode://" + args[0], function(err, res){ callback(err, res); })
```
yann300: addPeer.js is never used. is it ?

RPCTests/scripts/tests/MineBlocks2/step00_node02_mineBlocks.js
```diff
@@ -0,0 +1,31 @@
...
+
+   //wait for block being mined and propagated
+   sleep(1000).then(() => {
+	web3.eth.getBlockNumber(function(err, res){ onGetBlockNumber(err, res); })
+   });
+}
+
+onGetBlockNumber = function (err, res)
+{
+   if (res == latestBlock + 1)
```
yann300, line 26: instead of check for failure like this, we could use `tape`
in `package.json` add  "tape": "^4.6.0",
in this file: 
`var tape = require('tape')`

and

```
tape('test suite', function (t) {
  t.test('test name', function (st) {
    st.equal(res, latestBlock + 1)
  })
}

winsvega: how can I refere to tape function in other part of the code? 
(to use it as a callback)

winsvega: this code is executed on console environment. so require does not present

PR #146, opened 2017-03-15
Add missing callcreates key to test fixture

pipermerriam:
### What was wrong?

One of the test fixtures in the arithmetic tests didn't have the `callcreates` key.

### How was it fixed?

I added it.

#### Cute animal picture

lvcsv2i


pipermerriam:
After reading http://www.ethdocs.org/en/latest/contracts-and-transactions/ethereum-tests/vm_tests/index.html I realize this isn't a bug.

PR #147, opened 2017-03-21
REVERT opcode tests

gumb0:
Some simple tests for REVERT


winsvega:
just from the sources tests wont be executed. 
you need to run 
./testeth -t GeneralStateTests/stRevertTest -- --filltests --checkstate 

and commit the actual tests as well

gumb0:
Ah, I see

PR #165, opened 2017-04-28
Fix storage position to be valid quantities

gsalgado:
According to the spec[1], quantities should not have leading zeroes
after the 0x prefix. Although most clients don't seem to care about
that, pyethapp does.

[1] https://github.com/ethereum/wiki/wiki/JSON-RPC


winsvega:
I will have to change testeth for that. as the tests are being overwritten

winsvega:
there is a discussion about this issue 
https://github.com/ethereum/EIPs/issues/633

pirapira:
The storage index is a 256-bit word, not an integer.

pirapira:
Closing as this PR edits automatically generated files manually.

PR #171, opened 2017-05-19
Adding final test cases for elliptic curves

pirapira:
I also renamed `paring` into `pairing`.


pirapira:
I created a new branch in `ethereum/tests`.

PR #174, opened 2017-05-24
Adding tests for RETURNDATACOPY and RETURNDATASIZE

pirapira:


pirapira:
This works with `cpp-ethereum:develop`.   I guess the first step is to merge this into `tests:develop`?

Issue #169, opened 2017-05-18
Travis CI seems to be failing for all PRs


pirapira:
Currently, Travis CI seems to fail on all PRs like this: https://travis-ci.org/ethereum/tests/builds/233281939?utm_source=github_status&utm_medium=notification

@winsvega agreed that the test script should check the grammar of the final tests (without `Filler.json` suffix), but not necessarily on the fillers.

winsvega:
should be working now

pirapira:
I noticed this started working a while ago.

PR #178, opened 2017-06-05
#169 Update travis.yml with Filler exclude

maaktweluit:
- Excluded *Filler.json from checking
- Excluded *Filler.json from linting
- Updated check output to print all "lines not starting with "./" and the line before" containing the affected file and error close together.


maaktweluit:
Proposal to fix #169

winsvega:
Thank you !!!

Issue #177, opened 2017-06-05
Link to testFillers dead in README.md


maaktweluit:
Hi Ethereum Testers,

When starting to get to know the project i found the link to the testFillers ( the second in the README.md ) is dead, i think they have moved in cpp-ethereum.

I'd love to fix it but i do not know where they have moved.

I've checked open issues and PR's and do not see any other report like this.

Have a good one!

Maaktweluit

maaktweluit:
Maybe they live here now: https://github.com/ethereum/cpp-ethereum/tree/develop/test/tools/jsontests

PR #179, opened 2017-06-05
#177 Updated README to the moved Fillers

maaktweluit:


maaktweluit:
Based on the possible solve in #177 i've updated the README.md

PR #190, opened 2017-06-21
Add missing '0's in calldata.  The hex strings were of odd length

pirapira:


winsvega:
I should check that after this change test is working

pirapira:
I realize I forgot the --checkstate option when I did filling.

pirapira:
I'm going to file a related cpp pull-request.  So you might want to wait for that.

pirapira:
There are many more similar odd-length strings.

pirapira:
https://github.com/ethereum/cpp-ethereum/pull/4185 is the corresponding cpp-ethereum PR.

PR #193, opened 2017-06-22
GeneralTestFiller formatting

pirapira:
This PR
* splits generalstatetest fillers that contain more than one tests and
* fixes odd-length hex strings that represent code.


pirapira:
https://github.com/ethereum/cpp-ethereum/pull/4196 is the corresponding cpp-ethereum branch.

PR #195, opened 2017-06-23
correct staticcall tests

winsvega:


pirapira:
https://github.com/ethereum/cpp-ethereum/pull/4200 is the corresponding cpp-ethereum PR.

src/GeneralStateTestsFiller/stStaticCall/static_call_value_inheritFiller.json
```diff
@@ -0,0 +1,70 @@
...
+            "a94f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
+                "balance" : "1000000000000",
+                "code" : "",
+                "nonce" : "0",
+                "storage" : {
+                }
+            },
+            "094f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
+                "balance" : "1",
+                "code" : "{ [[0]] (STATICCALL 50000 0xc94f5374fce5edbc8e2a8697c15331677e6ebf0b 0 0 0 32) [[1]] (MLOAD 0) }",
```
winsvega, line 40: if something goes wrong int the static call. (MLOAD would still return 0)
thats why I think using 
(if (= 0 (CALLVALUE)) (MSTORE 1 1) (SSTORE 1 2) ) 
is better. cause this way static call will return 0 if callvalue != 0  inside a staticcall

src/GeneralStateTestsFiller/stStaticCall/static_call_value_inherit_from_callFiller.json
```diff
@@ -0,0 +1,76 @@
...
+            },
+            "c8b90f8d999d9d9988209481999a9889098a89a0" : {
+                "balance" : "0",
+                "code" : "{ (CALL 100000 0x094f5374fce5edbc8e2a8697c15331677e6ebf0b 10 0 0 0 0) }",
+                "nonce" : "0",
+                "storage" : {}
+            },
+            "c94f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
+                "balance" : "1",
+                "code" : "{ (MSTORE 0 (CALLVALUE)) (RETURN 0 32) }",
```
winsvega, line 54: this return value would be returned to c8b90f8d999d9d9988209481999a9889098a89a0 and wont be stored cause memory is not allocated.   `10 0 0 0 0) `
094f5374fce5edbc8e2a8697c15331677e6ebf0b  will always return 0 for (MLOAD 0)

src/GeneralStateTestsFiller/stStaticCall/static_callToDelCallOpCodeCheckFiller.json
```diff
@@ -24,7 +24,7 @@
         "pre" : {
             "1000000000000000000000000000000000000000" : {
                 "balance" : "1000000000000000000",
-		"code" : "Static call success",
+		"code" : "Static call fails because SSTORE is attempted",
```
winsvega: this comment is not an error. static call below MUST return 1 otherwise test fails

src/GeneralStateTestsFiller/stStaticCall/static_call_value_inherit_from_callFiller.json
pirapira: The return value will be returned to "094f5374fce5edbc8e2a8697c15331677e6ebf0b".

src/GeneralStateTestsFiller/stStaticCall/static_call_value_inheritFiller.json
pirapira: If something goes wrong in the static call, the `[[0]]` storage slot would receive 0, and we can detect that.

winsvega: well by wrong I mean if you made a typo or something 
or if the code wont work as you expect it to work

src/GeneralStateTestsFiller/stStaticCall/static_call_value_inherit_from_callFiller.json
winsvega: I see. I usually put the contracts in the call order. 

1)  calls 2
2)  calls 3
3)  do smth 

also addresses like 0x1000000000
0x20000000

are easier to read.

PR #196, opened 2017-06-26
Add test cases for RETURNDATACOPY after a successful CREATE should no…

pirapira:
…t throw if size = offset = zero

### DIFF
--- a/src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_0_0_following_successful_createFiller.json
+++ b/src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_0_0_following_successful_createFiller.json
@@ -20,6 +20,9 @@
                         "storage" : {
                             "0x00" : "0x01"
                         }
+                    },
+                    "945304eb96065b2a98b57a48a06ae28d285a71b5" : {
+                        "shouldexist" : false
                     }
                 }
             },
@@ -34,6 +37,11 @@
                         "storage" : {
                             "0x00" : "0x00"
                         }
+                    },
+                    "945304eb96065b2a98b57a48a06ae28d285a71b5" : {
+                        "storage" : {
+                            "0x00" : "0x01"
+                        }
                     }
                 }
             }
@@ -41,7 +49,8 @@
         "pre" : {
             "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6" : {
                 "balance" : "0x00",
-                "code" : "{ (create (STOP)) (RETURNDATACOPY 0 0 0) (SSTORE 0 (MLOAD 0)) }",
+                "code" : "create should succeed. RETURNDATACOPY should also succeed in Metropolis.",
+                "code" : "{ (create 0 0 (lll {(SSTORE 0 1)} 0)) (RETURNDATACOPY 0 0 0) (SSTORE 0 0) }",
                 "nonce" : "0x00",
                 "storage" : {
                     "0x00" : "0x01"
--- a/GeneralStateTests/stReturnDataTest/returndatacopy_0_0_following_successful_create.json
+++ b/GeneralStateTests/stReturnDataTest/returndatacopy_0_0_following_successful_create.json
@@ -2,7 +2,7 @@
     "returndatacopy_0_0_following_successful_create" : {
         "_info" : {
             "comment" : "",
-            "filledwith" : "cpp-1.3.0+commit.5ad949e6.Linux.g++",
+            "filledwith" : "cpp-1.3.0+commit.29c877a3.Linux.g++",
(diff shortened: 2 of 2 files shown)


src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_0_0_following_successful_createFiller.json
```diff
@@ -0,0 +1,82 @@
...
+                "balance" : "0x00",
+                "code" : "{ (seq (MSTORE 0 0x0000111122223333444455556666777788889999aaaabbbbccccddddeeeeffff) (RETURN 0 32)) }",
+                "code" : "0x7d111122223333444455556666777788889999aaaabbbbccccddddeeeeffff60005260206000f3",
+                "nonce" : "0x00",
+                "storage" : {
+                }
+            },
+            "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6" : {
+                "balance" : "0x00",
+                "code" : "{ (seq (create (STOP)) (RETURNDATACOPY 0 0 0) (SSTORE 0 (MLOAD 0)) )}",
```
winsvega: what does (seq (create (STOP))   do ?

src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_0_0_following_successful_createFiller.json
```diff
@@ -0,0 +1,82 @@
...
+                            "0x00" : "0x00"
+                        }
+                    }
+                }
+            }
+        ],
+        "pre" : {
+            "0x0aabbccdd5c57f15886f9b263e2f6d2d6c7b5ec6" : {
+                "balance" : "0x00",
+                "code" : "{ (seq (MSTORE 0 0x0000111122223333444455556666777788889999aaaabbbbccccddddeeeeffff) (RETURN 0 32)) }",
```
winsvega: is here a collision with this contract address?

src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_0_0_following_successful_createFiller.json
winsvega: I think now when we are working together we should write more comments to the tests. so it would be clear what a particular test case do and what results to expect and why. 
a comments could be added to the filler in doubled fields. 

a generela description for other developers could be added into _info fields

pirapira: The `seq` contains `(create ...) (RETURNDATACOPY ...) (SSTORE ...)`.

src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_0_0_following_successful_createFiller.json
pirapira: No here is no collision.

winsvega: I dont see where this account gets executed

pirapira: I will remove this account.

src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_0_0_following_successful_createFiller.json
pirapira: But since there is `{}` I don't need the `seq`.  I'll remove it.

pirapira:
@winsvega I addressed your comments.

src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_0_0_following_successful_createFiller.json
```diff
@@ -0,0 +1,74 @@
...
+                            "0x00" : "0x00"
+                        }
+                    }
+                }
+            }
+        ],
+        "pre" : {
+            "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6" : {
+                "balance" : "0x00",
+                "code" : "{ (create (STOP)) (RETURNDATACOPY 0 0 0) (SSTORE 0 (MLOAD 0)) }",
```
winsvega: you could use sstore  instead of stop and see the new contract in the post state

pirapira: I did, and added an expectation about the storage.

PR #207, opened 2017-07-03
Remove compiled code

pirapira:
Before this PR, some filler files contained
```
"code": "<LLL code>"
"code": "<compiled hex code>"
```

This was a workaround when the LLL compiler was not available.  This PR removes the compiled hex code from such filler files.

Mysteriously, this changed some tests.


winsvega:
are you sure that those opcodes are implemented in your version of lllc ?

pirapira:
Yes, I was able to fill these tests.

winsvega:
thats strange. maybe we changed something in the implementation since then? 
the result bytecode looks the same but the hash is different.

ah wait it is not. the code has changed.

PR #208, opened 2017-07-03
Add a test RETURNDATACOPY at the beginning of a program should throw …

pirapira:
…an exception if offset + size = 2^256

### DIFF
--- a/src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_initial_256Filler.json
+++ b/src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_initial_256Filler.json
@@ -27,7 +27,7 @@
         "pre" : {
             "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6" : {
                 "balance" : "0x0de0b6b3a7640000",
-                "code" : "{ (CALLDATACOPY 0 0 32) (RETURNDATACOPY (- 0 (MLOAD 0)) 0 0x64) (MSTORE 0 0x112233445566778899aabbccddeeff) (SSTORE 0 (MLOAD 0)) }",
+                "code" : "{ (RETURNDATACOPY (- 0 (CALLDATALOAD 0)) 0 0x64) (MSTORE 0 0x112233445566778899aabbccddeeff) (SSTORE 0 (MLOAD 0)) }",
                 "nonce" : "0x00",
                 "storage" : {
                     "0x00" : "0x01"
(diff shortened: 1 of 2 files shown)


winsvega:
you could also check gas usage for throw scenarious

pirapira:
If the gas usage is different, the root state hash will be different.

src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_initial_256Filler.json
```diff
@@ -0,0 +1,60 @@
...
+                            "0x00" : "0x01"
+                        }
+                    }
+                }
+            }
+        ],
+        "pre" : {
+            "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6" : {
+                "balance" : "0x0de0b6b3a7640000",
+                "code" : "{ (MSTORE 0 0x112233445566778899aabbccddeeff) (RETURNDATACOPY (- 0 32) 0 32) (SSTORE 0 (MLOAD 0)) }",
```
winsvega: if you test the edge condition you could also check 
that test fails   on    X  case.   but it works on X-1 case. and fails on X+1 case. 
you could pass different value through transaction data

pirapira: Added.

src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_initial_256Filler.json
```diff
@@ -27,7 +27,7 @@
         "pre" : {
             "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6" : {
                 "balance" : "0x0de0b6b3a7640000",
-                "code" : "{ (MSTORE 0 0x112233445566778899aabbccddeeff) (RETURNDATACOPY (- 0 32) 0 32) (SSTORE 0 (MLOAD 0)) }",
+                "code" : "{ (CALLDATACOPY 0 0 32) (RETURNDATACOPY (- 0 (MLOAD 0)) 0 0x64) (MSTORE 0 0x112233445566778899aabbccddeeff) (SSTORE 0 (MLOAD 0)) }",
```
winsvega: you could also use (CALLDATALOAD 0)

pirapira: I think that's better.

pirapira: Done.

PR #209, opened 2017-07-03
Remove shallowStackOK.json because shallowStackOKFiller.json does not exist

pirapira:
The test also caused a formatting error in https://travis-ci.org/ethereum/go-ethereum/jobs/249117383


pirapira:
@fjl this was the test without 0x prefix.

PR #212, opened 2017-07-05
Fix typos

ligi:


winsvega:
You will also have to change the source Fillers for that tests in the same way.

TransactionTests/ttTransactionTest.json
```diff
@@ -243,7 +243,7 @@
         "blocknumber" : "0",
         "rlp" : "0xf880800182520894095e7baea6a6c7c4c2dfeb977efac326af552d87a1010000000000000000000000000000000000000000000000000000000000000000801ba048b55bfa915ac795c431978d8a6a992b628d557da5ff759b307d495a36649353a0efffd310ac743f371de3b9f7f9cb56c0b28ad43601b4ab949f53faa07bd2c804"
     },
-    "TransactionWithHihghGas" : {
+    "TransactionWithHighGas" : {
```
winsvega, line 246: src/TransactionTests/ttTransactionTestFiller.json

has to have the same test name.

ligi: Thanks - will adapt that later - was not aware of this - I am using these test-cases a bit differently and just noticed the typos this way: https://github.com/walleth/kethereum/blob/master/functions/src/test/java/org/kethereum/functions/transactions/TransactionTestData.kt

ligi:
@winsvega pushed changes as requested

pirapira:
Since @winsvega is on vacation, I just merge.

Issue #204, opened 2017-07-01
JSON schema for state tests


cdetrio:
Create a [JSON schema](http://json-schema.org/) for the General State Tests format https://github.com/ethereum/EIPs/issues/176. Also add a step in [travis](https://github.com/ethereum/tests/blob/develop/.travis.yml) to validate the state tests against the schema.

See also https://github.com/ethereum/tests/issues/203

jwasinger:
I can take on this issue.  Translating EIP#176 into JSON schema looks straightforward.

cdetrio:
test JSON format is also being discussed at https://github.com/ethereum/EIPs/issues/633 and https://github.com/ethereum/cpp-ethereum/pull/4263.  Per https://github.com/ethereum/cpp-ethereum/pull/4263, state test format is frozen until after Metropolis.

winsvega:
@jwasinger do we need this issue open?

jwasinger:
@winsvega this is old.  Closing it.

cdetrio:
for reference, schemas were merged here https://github.com/ethereum/tests/pull/213  with all schemas passing at the time.

PR #213, opened 2017-07-06
JSON schema for state tests

jwasinger:
I'm opening this pull request for feedback on what I have implemented so far regarding #204 

I didn't originally intend to write the schema validation script in JavaScript.  Unfortunately, the only library I could find that supported limiting allowed JSON object properties via `"additionalProperties": "false"` was implemented using Node.

### DIFF
--- a/JSONSchema/schema.json
+++ b/JSONSchema/schema.json
@@ -5,10 +5,12 @@
             "pattern": "^$"
         },
         "HexData": {
+            "description": "Hex data. see https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding",
             "pattern": "^0x([0-9a-fA-F][0-9a-fA-F])*$",
             "type": "string"
         },
         "HexQuantity": {
+            "description": "Hex quantity. see https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding",
             "pattern": "(^0x0$)|(^0x[1-9a-fA-F][0-9a-fA-F]*$)",
             "type": "string"
         },
@@ -91,9 +93,18 @@
             "additionalproperties": true,
             "properties": {
                 "balance": { "$ref": "#/definitions/ConfusedHexType" },
-                "code": { "type": "string" },
+                "code": { "$ref": "#/definitions/HexMaybePrefixOrEmpty" },
                 "nonce": { "$ref": "#/definitions/ConfusedHexType" },
-                "storage": { "type": "object" }
+                "storage": {
+                    "type": "object",
+                    "additionalProperties": false,
+                    "patternProperties": {
+                        "^0x[0-9a-f]+": {
+                            "description": "storage key with 0x. data is HexData",
+                            "$ref": "#/definitions/HexData"
+                        }
+                    }
+                }
             },
             "required": [
                 "balance",
--- a/JSONSchema/filler-schema.json
+++ b/JSONSchema/filler-schema.json
@@ -5,10 +5,12 @@
             "pattern": "^$"
         },
         "HexData": {
+            "description": "Hex data. see https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding",
             "pattern": "^0x([0-9a-fA-F][0-9a-fA-F])*$",
             "type": "string"
         },
         "HexQuantity": {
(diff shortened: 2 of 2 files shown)


JSONSchema/schema.json
```diff
@@ -0,0 +1,168 @@
...
+                            "type":"string",
+                            "pattern":"^0x[0-9a-f]*$"
+                        },
+                        "currentDifficulty":{
+                            "type":"string",
+                            "pattern":"^0x[0-9a-f]*$"
+                        },
+                        "currentGasLimit":{
+                            "type":"string",
+                            "pattern":"^0x[0-9a-f]*$"
```
sdtsui: Is there a way the schema can enforce the `not larger then maxGasLimit = 0x7fffffffffffffff` rule in the EIP?
Haven't done this in awhile.

JSONSchema/schema.json
```diff
@@ -0,0 +1,168 @@
...
+                        },
+                        "secretKey":{
+                            "type":"string",
+                            "pattern":"^0x[0-9a-f]*$"
+                        },
+                        "to":{
+                            "type":"string",
+                            "pattern":"^0x[0-9a-f]*$"
```
cdetrio: the `to` field can be empty, for a contract creation tx. [example](https://github.com/ethereum/tests/blob/3b90fbfaf0b6180be9276bbd050ac2475b95d949/GeneralStateTests/stInitCodeTest/StackUnderFlowContractCreation.json#L82)

JSONSchema/schema.json
```diff
@@ -0,0 +1,168 @@
...
+                                "type": "integer"
+                            }
+                        },
+                        "additionalProperties": false
+                    }
+                },
+                "additionalProperties": false
+            }
+        },
+        "HexString": {
```
cdetrio: any reason why HexString isn't being used?

cdetrio:
Great!

JSONSchema/schema.json
jwasinger: As far as I know, there is not going to be an easy way to enforce this rule against string values.  I'm guessing this could be done using a regex and I will look into it further.

JSONSchema/schema.json
jwasinger: That's an oversight on my part.  It's being deleted and replaced with several different sub-schema definitions.

JSONSchema/schema.json
jwasinger: Good to know.  I'll add this too.

JSONSchema/schema.json
```diff
@@ -0,0 +1,185 @@
...
+                            "type":"string",
+                            "pattern":"^0x[0-9a-f]*$"
+                        },
+                        "secretKey":{
+                            "type":"string",
+                            "pattern":"^0x[0-9a-f]*$"
+                        },
+                        "to":{
+                            "type": "string",
+                            "pattern":"^(0x([0-9a-f]*){10})?$"
```
cdetrio: [nitpick] can we use `oneOf` to allow for the empty string? will be more readable than a regex

jwasinger: Agreed.  I'll make the change.

jwasinger:
@cdetrio I think this is almost ready to be merged.  I just want to add some more linting around allowed hex values for certain fields first (e.g. a byte field cannot have a string of more than length 2, an address cannot be more than length 40, etc.).  I'm guessing this will probably reveal some additional flawed tests.

I'll try to have these changes ready within a few hours.

pirapira:
Shall we fix the Travis build before we merge?

jwasinger:
Agreed.  I'll fix those and then we can merge this.

cdetrio:
Added a schema for state test fillers. The schemas are quite permissive, only 15 state tests are failing now, all because their `transaction.data` fields are missing a `0x` prefix. The corresponding filler files are also failing for the same reason.

If the `0x` prefix is added to the failing files, and https://github.com/ethereum/tests/pull/250 is merged (with a couple other minor fixes), then the travis build should be green. At least then the formats shouldn't get any worse.

The schemas can be gradually made more strict, and the files fixed over time. Or alternatively, we can add a rule in the schema to detect newly generated files, and enforce a strict schema on those.

pirapira:
Merging while it's passing.

PR #214, opened 2017-07-06
Add an explanation

pirapira:
I'm trying to a comment @winsvega wanted.

### DIFF
--- a/src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_following_failing_callFiller.json
+++ b/src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_following_failing_callFiller.json
@@ -35,6 +35,7 @@
             },
             "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6" : {
 				"balance" : "0x00",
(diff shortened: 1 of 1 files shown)


src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_following_failing_callFiller.json
```diff
@@ -35,6 +35,7 @@
             },
             "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6" : {
 				"balance" : "0x00",
+                "code" : "CALL should fail because of an invalid instruction, and RETURNDATACOPY should also fail because of out-of-bound access.  SSTORE should not be reached",
```
winsvega, line 38: so if you want a CALL to fail, you could make it like this 

(CALL  (CALLDATALOAD 0)  0x11111... 0 0 0 0 0)  

and pass the amount of gas of this subcall via data of transaction. 
this way you could get a subcall go OOG and not going OOG depending on transaction data 
and check out of bounds RETURNDATACOPY  in case if call fails and not fails (not oog)

pirapira: This filler is only about the failing call.

PR #221, opened 2017-07-10
Expect zeros in the storage of an overwritten contract

pirapira:
https://github.com/ethereum/cpp-ethereum/pull/4244 is the corresponding cpp-ethereum commit.


pirapira:
Travis runs fine on the cpp-ethereum PR, which has already been merged.

PR #244, opened 2017-08-04
Change tests so that the storage is cleared before, not after contract initialization

pirapira:
This should fix the problem @vbuterin saw on `createJS_ExampleContract` case.

The corresponding cpp-ethereum PR: https://github.com/ethereum/cpp-ethereum/pull/4305


winsvega:
where is the change to expect section  ?

pirapira:
@winsvega I didn't change the filler.  No filler checks this.

pirapira:
@winsvega only some "post" changes after filling.

PR #245, opened 2017-08-04
blockhash test on Byzantine transition

winsvega:


pirapira:
What was wrong before this PR?

winsvega:
nothing just adding blockhash code check on Byzantine transition

pirapira:
OK.

Issue #216, opened 2017-07-06
Malformed VMTest?


ehildenb:
VM Test `loop_stacklimit_1021` seems to be malformed (see file https://github.com/ethereum/tests/blob/develop/VMTests/vmIOandFlowOperationsTest.json).

It has an `expect` node which indicates that the account in question should have something it its storage at the end of execution, but there are no opcodes in the test that can store anything to account storage.

pirapira:
I agree that the case is malformed.  `expect` sections should not exist in VMTests.  Instead, `post` section should be present.

winsvega:
Maybe it was filled wrong. The VM tests fillers could have expect section but it should be erased from the final test if the test generation went without errors.

pirapira:
@ehildenb the test in question has changed.  Please check if your problem still stays.

ehildenb:
@pirapira indeed the expect node is removed. However, in file `vmEnironmentalInfoTest.json` there is still an `expet` node and an `export` node. I suspect similar issues with these tests.
https://github.com/ethereum/tests/blob/a3880f58b3862ef25e9973b1e1043685ed45fc47/VMTests/vmEnvironmentalInfoTest.json#L396
https://github.com/ethereum/tests/blob/a3880f58b3862ef25e9973b1e1043685ed45fc47/VMTests/vmEnvironmentalInfoTest.json#L814

pirapira:
@ehildenb thank you.  I'm starting a relevant cleanup: instead of copying the filler and modifying it, building the output from the empty JSON.

pirapira:
@ehildenb I searched for `expet` and `export`, and now they are gone from VMTests.

I'd appreciate if you try again.

PR #246, opened 2017-08-04
Remove "expect" sections from VM tests

pirapira:
The corresponding C++ PR is https://github.com/ethereum/cpp-ethereum/pull/4306


pirapira:
After this is merged, we should investigate if #216 is fixed.

PR #250, opened 2017-08-05
fix invalid json state test fillers

cdetrio:
Did not try filling with these fixes, but all filler files in src/GeneralStateTests are now valid json.


winsvega:
so now all of this comments will end up getting into the final tests
wouldn't it be confusing ?

cdetrio:
Yea we don't want that. Perhaps `testeth` can ignore keys that begin with `//`?

src/GeneralStateTestsFiller/stZeroKnowledge/pairingTestFiller.json
```diff
@@ -130,7 +130,7 @@
 		"[1] vk.B + proof.B + negateG1(proof.Bp) + P2",
 		"[2] proof.C + vk.C + negateG1(proof.Cp) + P2",
 		"[3] proof.K + vk.gamma + negateG1(addG1(vkx, addG1(proof.A, proof.C))) + vk.gammaBeta2 + negateG1(vk.gammaBeta1) + proof.B",
-		"[4] addG1(vkx, proof.A) + proof.B + negateG1(proof.H) + vk.Z +	negateG1(proof.C) + P2",
+		"[4] addG1(vkx, proof.A) + proof.B + negateG1(proof.H) + vk.Z + negateG1(proof.C) + P2",
 		"[5] proof.A + vk.A + proof.Ap + P2"
 	    ],
             "data" : [
```
pirapira, line 136: To me, the field `"data"` seems to be duplicated.  Is this still a valid JSON?

pirapira:
I created an issue about removing `\\` fields https://github.com/ethereum/cpp-ethereum/issues/4309

src/GeneralStateTestsFiller/stZeroKnowledge/pairingTestFiller.json
pirapira: I opened an issue about this https://github.com/ethereum/tests/issues/252

cdetrio: technically duplicate keys are valid JSON. I had to look it up: "The names within an object SHOULD be unique." So duplicate keys are not recommended but not forbidden.

Luckily the schema validator we're using only validates the last key/value when there are duplicates.

PR #255, opened 2017-08-08
Add a test case that RETURNDATACOPYs REVERTed data

pirapira:


pirapira:
Should I create a cpp PR?

PR #261, opened 2017-08-14
Add State Tests for Unused Opcodes

jwasinger:
Based on an issue discovered by @cdetrio (https://gist.github.com/cdetrio/2776a57db1594f679ceef11d05bfc9a9). This pull adds tests to make sure that clients aren't implementing unused opcodes: SSIZE, SLOADBYTES, SSTOREBYTES.


src/GeneralStateTestsFiller/stBadOpcode/badOpcodeSLOADBYTESFiller.json
```diff
@@ -0,0 +1,65 @@
...
+                "indexes" : {
+                    "data" : -1,
+                    "gas" : 0,
+                    "value" : -1
+                },
+                "network" : ["ALL"],
+                "result" : {
+                    "095e7baea6a6c7c4c2dfeb977efac326af552d87" : {
+                        "code": "SLOADBYTES: key, mstart, msize",
+                        "code" : "0x 6001 6000 6001 e1 6001 6002 55",
```
winsvega: I thought we agreed to make correct json schemas. so instead name the fields like 
``` "//code1" :  ```
or 
``` "//comment1" : ```

jwasinger: Done

cdetrio:
We should just have tests for all possible opcodes, or 256 test cases. Could do it in a single json file by using an array of 256 elements for the `data` field. Something like `["0x00", "0x01", "0x02", ...]`, though it might tricky to get contract code to execute bytecode passed in as calldata. An easier way would be to use contract creation tx's that fail when the opcode is invalid, and succeed when its valid.

winsvega:
make a transaction with init code 

push 01 push 01 push 01 push 01 push 01 push 01 push 01 (some opcode)  (call to 0x11111)

if opcode is ok, then call to 0x11111 succeeded and could write a storage value

jwasinger:
I like these ideas.  I'm working on generating the tests.
EDIT: messed up this branch a bit so still WIP until I fix the issues.

pirapira:
@jwasinger have you fixed the issues you mentioned above?

winsvega:
cpp PR merged

PR #266, opened 2017-08-22
Refilled tests again so that 0x is in front of byte sequences

pirapira:


pirapira:
The problem is I don't understand why some `0x`s were not added last time.  Maybe because I `ctrl-c`ed `testeth`.

PR #268, opened 2017-08-22
Add blockchain test linting

jwasinger:
**Status:** Work in progress
**Depends On:** https://github.com/ethereum/tests/pull/275 
Adding linting for blockchain fillers and generated tests will help to prevent issues like https://github.com/ethereum/cpp-ethereum/issues/4369


JSONSchema/bc-schema.json
```diff
@@ -0,0 +1,390 @@
...
+                "transactions": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/Transaction"
+                    }
+                },
+                "uncleHeaders": {
+                    "type": "array"
+                },
+                "expectExceptionHomestead": {
```
winsvega: this field should be only in fillers

JSONSchema/bc-schema.json
```diff
@@ -0,0 +1,390 @@
...
+                "expectExceptionALL": {
+                    "type": "string"
+                },
+                "expectExceptionByzantium": {
+                    "type": "string"
+                },
+                "expectExceptionConstantinople": {
+                    "type": "string"
+                },
+                "blocknumber": {
```
winsvega: this field is optional but also could be forced to happen only in fillers

JSONSchema/bc-schema.json
```diff
@@ -0,0 +1,390 @@
...
+                    "$ref": "#/definitions/IntegerOrEmptyOrConfusedHex"
+                },
+                "chainname": {
```
winsvega: optional

JSONSchema/bc-schema.json
```diff
@@ -0,0 +1,390 @@
...
+                    "type": "string"
+                },
+                "chainnetwork": {
```
winsvega: optional

JSONSchema/bc-schema.json
jwasinger: I'm seeing it in quite a lot of filled tests.

jwasinger:
@winsvega right now all block fields are optional in the schema because there are tests with and without them.

winsvega:
the optional fields in blockchain test are not needed in the final test. the client should be able to input a block by its RLP. actually those fields are confusing client developers so we should remove it from filled tests.

@pirapira

PR #273, opened 2017-08-24
Fix the filler and fill vmIOandFlowOperationsTest

pirapira:
This PR changes a VMTest filler so that `testeth -t VMTests --filltests` succeeds.  This PR also contains the filled VM test files.


src/VMTestsFiller/vmIOandFlowOperationsTestFiller.json
```diff
@@ -1489,7 +1489,6 @@
         "expect" : {
             "0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6" : {
                 "storage" : {
-                    "0x02" : "0x23"
```
winsvega, line 1492: why this disappeared?

pirapira: I don't know, maybe something changed in cpp-ethereum.

pirapira: I'll investigate this https://github.com/ethereum/cpp-ethereum/issues/4395

pirapira:
I guess we should not just merge this.

pirapira:
I investigated the issue.  https://github.com/ethereum/cpp-ethereum/issues/4395

The filler was wrong, and a recent change started to detect the mistake in the filler.  I believe this PR is the right fix.  @winsvega

PR #276, opened 2017-08-26
Refill walletReorganizeOwnersFiller.json

pirapira:
walletReorganizeOwner.json contained an empty state in a post state in Byzantium, which is clearly wrong.

I had to remove expectations because the Filler was failing for Frontier.
This issue will be investigated separately at https://github.com/ethereum/cpp-ethereum/issues/4409


winsvega:
is it ready?

pirapira:
Yes.

PR #277, opened 2017-08-26
Add more opcodes to badOpcodes test

jwasinger:
This PR adds more opcodes to 'badOpcodes' test to make sure that all invalid opcodes are included in the test.


jwasinger:
@pirapira @winsvega I think this one is ready for merging.

src/GeneralStateTestsFiller/stBadOpcode/badOpcodesFiller.json
```diff
@@ -30,49 +30,138 @@
                 "nonce": "0",
                 "storage": {
                 }
+            },
+            "1000000000000000000000000000000000000001" : {
```
winsvega: where is it used?

src/GeneralStateTestsFiller/stBadOpcode/badOpcodesFiller.json
```diff
@@ -30,49 +30,138 @@
...
-		"0x60016001600160014b",
-		"0x60016001600160014c",
-		"0x60016001600160014d",
-		"0x60016001600160014e",
-		"0x60016001600160014f",
-		"0x60016001600160015c",
-		"0x60016001600160015d",
-		"0x60016001600160015e",
-		"0x60016001600160015f"
+"0x60016001600160010c60006000f3",
```
winsvega: you could add one valid opcode to check that a new contract is actually being created this way

jwasinger: ~~Is there a way to do this without adding an expect section for each index in the data~~ ~~array?  I am only familiar with the `"field": "-1" ` syntax which applies to all indices.~~

~~I'm not sure how to make the expect section different for a single case (or a subset of cases).~~
~~^Maybe this could be a useful feature?~~
Okay. just discovered array syntax `"data" : [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16 ... ]`.

src/GeneralStateTestsFiller/stBadOpcode/badOpcodesFiller.json
jwasinger: Unused.  Removing it

src/GeneralStateTestsFiller/stBadOpcode/badOpcodesFiller.json
jwasinger: Okay.  just discovered array syntax `"data" : [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16 ... ]`.

winsvega: this is also true for network field

winsvega:
could also run testeth -t StateTestsGeneral/stBadOpcodes -- --fillchain
testeth -t StateTestsGeneral/stBadOpcodes -- --filltests

to add this tests execution on hive

jwasinger:
Okay.  Currently waiting on the Blockchain tests to finish filling before adding them.  It's taking a very long time.

winsvega:
make sure you run only your test case 
```testeth -t StateTestsGeneral/stBadOpcodes -- --fillchain```
it should not be too long for one case

jwasinger:
@winsvega The reason it took so long was because there are many test cases (one for each possible invalid opcode) contained in the file.  Test generation complete now so I will push the results.

winsvega:
like 30 data fields * 5 forks = 150 blocks to mine.

jwasinger:
There are ~120 data fields now.

winsvega:
one hell of a tests

PR #292, opened 2017-08-30
Structural overhaul of the whole docs / Updated state test section / info box on test type status

holgerd77:
Hi,
this is still a "get things going" and not a "everything is accurate now" PR! :-)

I've done a lot of structural work on the docs, should be much easier to read now and also internally easier to edit for doc updates.

I updated the state test section with the latest structure from EIP 176, this is still in a bit raw version, would be glad anyway if this could be gradually updated/enhanced in later PRs.

For every test type there is now an info box at the beginning of the section, containing the linked location, the supported forks/networks and the current status (Actively supported / Outdated). Would be glad if this could be reviewed.

Cheers
Holger

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -8,8 +8,7 @@ Ethereum Tests
 |
 
 .. note::
-   Larger parts of the test docs are currently outdated (see :ref:`contribute` on how to
-   help).
+   See :ref:`contribute` if you want to help improve this documentation.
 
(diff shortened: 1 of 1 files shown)


holgerd77:
Added a dev version of the changes on RTD: http://ethereum-tests-dev.readthedocs.io/

This is stupidly reviewed in the raw version.

PR #299, opened 2017-09-04
Re-fill stBadOpcodes

jwasinger:
Somehow this state test was not filled correctly.


winsvega:
the issue is that when you run testeth --filltests --fillchain  
the state tests being filled as empty file again. it is a refactoring issue in testeth @pirapira

pirapira:
@winsvega though a similar issue was known from the past https://github.com/ethereum/cpp-ethereum/issues/4258.

PR #302, opened 2017-09-04
Add a test containing a call transaction that reverts

pirapira:


pirapira:
Yes, I commented in the filler https://github.com/ethereum/tests/pull/302/files#diff-492eaf2c2bf18025722448b6bb5e46bcR4

PR #303, opened 2017-09-05
Add a Blockchaintest where create transaction fails for a REVERT inst…

pirapira:
…ruction in the init code

The status code in the transaction receipt is the focus of interest.


BlockchainTests/bcValidBlockTest/createRevert.json
```diff
@@ -0,0 +1,500 @@
...
+        "lastblockhash" : "0x031366bc6868ea40e9c7a8a3ea4d95e5d46b12f03fc1b0fe012eb7e0adeb7de3",
+        "network" : "Constantinople",
+        "postState" : {
+            "0x00000000000000000000000000000000000000f0" : {
+                "balance" : "0x00",
+                "code" : "",
+                "nonce" : "0x00",
+                "storage" : {
+                }
+            },
```
pirapira: I see an empty account.  Weird.

winsvega: so this looks like a Constantinople only issue. should we merge the test?

pirapira: the next commit removed the empty account.

PR #304, opened 2017-09-05
[WIP] add more issue tests from the fuzz server

winsvega:
tests are not analyzed with geth yet. this PR is for geth debugging the tests


src/GeneralStateTestsFiller/stRandom/randomStatetest646Filler.json
```diff
@@ -0,0 +1,53 @@
+{
+    "randomStatetest646" : {
+        "_info" : {
+            "comment" : "Geth Failed this test on all networks"
```
winsvega, line 4: ok this test is fixed. the issue was in cpp applying zeroAddress rules on all forks.

PR #328, opened 2017-09-18
Add a test that fails before PR 4482

pirapira:
This PR adds a test on ecpairing where g1 is invalid while g2 is zero.

cpp-ethereum before PR 4482 accidentially returned some answer in this case.  It should throw.


src/GeneralStateTestsFiller/stZeroKnowledge/ecpairing_one_point_with_g2_zero_and_g1_invalidFiller.json
```diff
@@ -0,0 +1,275 @@
...
+                    "90f0b1ebbba1c1936aff7aaf20a7878ff9e04b6c": {
+                        "storage": {},
+                        "nonce": "0",
+                        "balance": "1000000000000000000",
+                        "code": "0x"
+                    }
+                }
+            }
+        ],
+        "explanation": "Puts the given data into the ECPAIRING precompile"
```
winsvega: we agreed to use //for comments in the fillers
and if you want the comment to end up in generated test fill the description into
```
"_info" : {
            "comment" : ""
}
```

_info.comment should be copied to the final test when filling

pirapira: OK.

pirapira:
@winsvega I added a comment in `_info`.

GeneralStateTests/stZeroKnowledge/ecpairing_one_point_with_g2_zero_and_g1_invalid.json
```diff
@@ -0,0 +1,364 @@
+{
+    "ecpairing_one_point_with_g2_zero_and_g1_invalid" : {
+        "_info" : {
+            "comment" : "",
```
winsvega, line 4: this field should be filled by testeth if there is one in a filler file

pirapira: testeth does not do that.  So testeth has a bug.

pirapira: https://github.com/ethereum/cpp-ethereum/issues/4454

PR #339, opened 2017-09-22
regenerate customMainNet tests

winsvega:


pirapira:
I noticed a comment is missing, so I created this issue: https://github.com/ethereum/tests/issues/340

PR #341, opened 2017-09-25
refill mainnetwork tests with byzantium at 4.37M

winsvega:


holiman:
Regression: https://github.com/ethereum/tests/pull/341/files#diff-972a1fed71f05b9b9ba157fcd60370c1R6 - switched back to not using hashes

winsvega:
sorry was using wrong executable. check it now

PR #343, opened 2017-09-27
Add a test where RETURNDATACOPY with a nonzero size fails after a CAL…

pirapira:
…L with insufficient balance


src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_following_too_big_transferFiller.json
```diff
@@ -0,0 +1,70 @@
...
+                            "0x00" : "0x01"
+                        }
+                    }
+                }
+            }
+        ],
+        "pre" : {
+            "0x0aabbccdd5c57f15886f9b263e2f6d2d6c7b5ec6" : {
+				"balance" : "0x00",
+                "code" : "{ (seq (MSTORE 0 0x0000111122223333444455556666777788889999aaaabbbbccccddddeeeeffff) (RETURN 0 32)) }",
```
winsvega, line 33: do ve have a test where return data is checked when the call fails on other reasons? 
like with REVERT? 

or where this subcall has another subcall that sets return data but then the first subcall fails 
and thus all first subcall should be reverted and in this scenario we should check the return data in the original call.

pirapira: I'll add these to the spreadsheet, and write a comment here again.

pirapira: Case 1. https://docs.google.com/spreadsheets/d/1xat7UI8GtB4ZGVdlK5_XQSHJZaMThi4SrlcL8XMZb5Q/edit#gid=1605675414&range=C36

pirapira: I didn't understand the second case you described.

winsvega: TRANSACTION -> CALL -> (CALL -> CALL RETURN | OOG) | check return data

-> mean call
| finish call and go one level up

PR #355, opened 2017-10-13
fuzztests bytecode

winsvega:


src/GeneralStateTestsFiller/stRandom2/randomStatetest647Filler.json
```diff
@@ -0,0 +1,61 @@
+{
+    "randomStatetest647" : {
+        "_info" : {
+            "comment" : "Malicious bytecode found by fuzztest tool"
```
holiman: You could make a comment that this basically only does `returndatacopy(0,-1)`

PR #356, opened 2017-10-13
Add a test case about transaction calls A (CALL B(CALL C(RETURN) OOG)…

pirapira:
… 'check buffers')

https://docs.google.com/spreadsheets/d/1xat7UI8GtB4ZGVdlK5_XQSHJZaMThi4SrlcL8XMZb5Q/edit#gid=1605675414&range=C37


src/GeneralStateTestsFiller/stReturnDataTest/returndatasize_after_oog_after_deeperFiller.json
```diff
@@ -0,0 +1,98 @@
...
+            "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
+                "balance" : "0x100000000000",
+                "code" : "{ (seq (CALL 100000 0xbb00000000000000000000000000000000000000 0 0 0 0 0) (while 1 (SSTORE 0 1)) )}",
+                "nonce" : "0x00",
+                "storage" : {
+                }
+            },
+            "0xbb00000000000000000000000000000000000000" : {
+                "balance" : "0x00",
+                "code" : "{ (seq (STOP) )}",
```
winsvega: this should return some non empty data. via revert or return

pirapira: That makes sense.  Will do.

pirapira: Done.

PR #359, opened 2017-10-17
Refill one test that was outdated

pirapira:
This fixes the `testeth --all` run.


pirapira:
This is a duplicate of #358.

PR #365, opened 2017-10-24
returndatacopy bug

winsvega:


chfast:
Try removing `--depth` option. I will work on better solution separately.

PR #371, opened 2017-11-22
[WIP] zero signature test fillers

winsvega:
just keep the test fillers for zero signature tests


pirapira:
Are these the same tests as those removed earlier?

winsvega:
yes. I remember notifying you.

pirapira:
In the last coredev call, it was confirmed that EIP-86 will be probably replaced with something else.

https://github.com/ethereum/EIPs/blob/master/EIPS/eip-145.md will be in more likely.

PR #374, opened 2017-12-04
General Transaction Tests

winsvega:


pirapira:
Did Travis in cpp-ethereum already try the new tests?

winsvega:
yes

winsvega:
its this one https://github.com/ethereum/cpp-ethereum/pull/4664

pirapira:
Why are 10k lines removed while only 8k are added?

winsvega:
I've removed unneeded copies of the tests. a general test describe filler once and then the final test has results on each fork

before I had to copy each test case for new fork. now all testcases are executed on all forks by default

PR #375, opened 2017-12-06
Adding a new test about returndatacopy following a revert in an init code

pirapira:
Implementing this case: https://docs.google.com/spreadsheets/d/1xat7UI8GtB4ZGVdlK5_XQSHJZaMThi4SrlcL8XMZb5Q/edit#gid=1605675414&range=C55


src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_following_revert_in_createFiller.json
```diff
@@ -0,0 +1,74 @@
...
+                            "0x00" : "0x0000111122223333444455556666777788889999aaaabbbbccccddddeeeeffff"
+                        }
+                    }
+                }
+            }
+        ],
+        "pre" : {
+            "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6" : {
+				"balance" : "0x00",
+                "code" : "{ (seq (CREATE 0 0 (lll (seq (MSTORE 0 0x0000111122223333444455556666777788889999aaaabbbbccccddddeeeeffff) (REVERT 0 32)) 0)) (RETURNDATACOPY 0 0 32) (SSTORE 0 (MLOAD 0)) )}",
```
winsvega, line 44: how about putting the same but inside transaction init code?

pirapira: That doesn't work.  I want to do returndatacopy.

winsvega: I mean you could also put CREATE opcode inside transaction INIT code. 
and that new contract that being created from transaction init code would run the same as in your test here. and then you could check returndatacopy of that contract while being in transaction init code.

winsvega: but thats a separate testcase, yes.

pirapira: I added a cell in the spreadsheet https://docs.google.com/spreadsheets/d/1xat7UI8GtB4ZGVdlK5_XQSHJZaMThi4SrlcL8XMZb5Q/edit#gid=1605675414&range=C59

PR #376, opened 2017-12-06
Document RPC test

pirapira:
Gitter ethereum/tests:
```
Yoichi Hirai @pirapira 17:25
Can you add a document how to run it?

Eth-Gitter-Bridge @Eth-Gitter-Bridge 17:25
[W Dimitry] you need node
[Yoichi Hirai] ~$ node --version
v6.10.2

Eth-Gitter-Bridge @Eth-Gitter-Bridge 17:27
[W Dimitry] https://github.com/ethereum/cpp-ethereum/blob/develop/scripts/runalltests.sh#L38

Yoichi Hirai @pirapira 17:28
ok
```

### DIFF
--- a/RPCTests/README.md
+++ b/RPCTests/README.md
@@ -54,7 +54,7 @@ TEST_getBlockHashOnNode1 OK
 (node:30406) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Callback was already called.
 ```
 
-Answer: everybody experiences these failures now.
+Answer: everybody experiences these failures now. They are being tracked in [issue 377](https://github.com/ethereum/tests/issues/377).
 
 
 Do these failures indicate bugs in cpp-ethereum or in the test?


winsvega:
well just running the script is fine. it will ask for cpp client path and print help

another issue is that this tool has a variable with blockinfo. and this arguments need manual
like this one: 

`args["node1_lastblock"].hash`

this tests currently has an issue. for some reason author is not being set when running cpp client from console for the first time. in a fresh environment. @gumb0  is aware of that.

gumb0:
There are also some dependency node.js modules that need to be installed before it is able to run.

And in the end I didn't succeed with running it without errors at all, not setting the author was only one of the issues.

pirapira:
I'm adding FAQs now.

pirapira:
I couldn't answer these two questions:

* Do these failures indicate bugs in cpp-ethereum or in the test?
* Has any other clients been tested with this?

@winsvega any ideas?

winsvega:
this tool is cpp-client only. it could be updagraded into multi client tool prettu easily if rpc interface is the same.
yes it indicates a bug in cpp-ethereum. 
it starts two nodes and perform an rpc requests to both of the nodes. connecting it to each other on a private network. basic tests are  (addPeer. check that peer is connected. mine a block. check that block is seen by the other peer. generate a transaction...)

@gumb0 
yes I remember now that it require a specific nodejs module to be installed. I guess there was a command to download the dependencies.

gumb0:
> yes it indicates a bug in cpp-ethereum.

I'm not so sure in the correctness of the tests

pirapira:
I created an issue for the errors https://github.com/ethereum/tests/issues/377

PR #381, opened 2017-12-13
Test RETURN for big offset

hugo-dc:
This is a new test case created to test a problem in bn.js (used in ethereumjs-vm)
https://github.com/indutny/bn.js/blob/master/lib/bn.js#L2145 

Currently when the RETURN fee for a very high offset is going to be added to the totalGas used, the bn.js can throw an assertion when the return fee is very high.

This code https://github.com/ethereumjs/ethereumjs-vm/blob/master/lib/runCall.js#L165 checks when the return fee is higher than the gasLimit and sets the returnFee as gasLimit + 1 , this way the return fee remained lower than the bn.js assertion.

In this new test case I added a higher gasLimit and this gasLimit is higher than the number validated in the bn.js assertion (0x4000000). I used this test case locally and it fails in ethereumjs-vm.


src/GeneralStateTestsFiller/stCreateTest/CREATE_ContractRETURNBigOffsetFiller.json
```diff
@@ -0,0 +1,44 @@
...
+            "a94f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
+                "balance" : "100000000000",
+                "code" : "",
+                "nonce" : "0",
+                "storage" : {
+                }
+            }
+        },
+        "transaction" : {
+            "data" : "{ (RETURN 0x074ac2 0x0d15bc) }",
```
winsvega: multiple json comment fields should be prefixed with //

pirapira: @winsvega do you want `//`s before we merge this PR?

winsvega: well actually this should breake json scheme check in the test repo. but it is green.
@jwasinger ? 

the test itself is ok.

hugo-dc: thanks for the feedback @winsvega @pirapira , I'll add `//` before it is merged.

hugo-dc: @pirapira @winsvega I added the prefix `//`to the comment fields

src/GeneralStateTestsFiller/stCreateTest/CREATE_ContractRETURNBigOffsetFiller.json
```diff
@@ -0,0 +1,44 @@
...
+                "nonce" : "0",
+                "storage" : {
+                }
+            }
+        },
+        "transaction" : {
+            "//data" : "{ (RETURN 0x074ac2 0x0d15bc) }",
+            "//data" : "{ (RETURN 0x074ac2 0x010000) }",
+            "//data" : "{ (RETURN 0x074ac2 0x051eb9) }",
+            "//data" : "{ (RETURN 0x074ac2 0x051eb8) }",
```
pirapira: The ordering of these lines does not correspond to the ordering in the actual `data` field.

winsvega:
you could now use yaml format for the tests
https://github.com/ethereum/tests/issues/385

hugo-dc:
@winsvega @pirapira used yaml format and ordered the data comments.

pirapira:
Thank you!

PR #382, opened 2017-12-13
Test with CALLDATACOPY where call data offset > call data length, length to copy > 0

jwasinger:
The resulting behavior is that `length` amount of memory is zeroed out.

In ethereumjs-vm I was able to completely clear memory in this case without causing test failures: https://github.com/ethereumjs/ethereumjs-vm/pull/174/commits/5366bc4254b72523f8347bf68d0cd494212173ce

PR #388, opened 2018-01-04
add filler for codeCopyOffset, change location of the test to stMemoryTest

jwasinger:
Corrects an error with this PR: https://github.com/ethereum/tests/pull/382


GeneralStateTests/stMemoryTest/codeCopyOffset.json
```diff
@@ -45,7 +89,7 @@
             },
             "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee" : {
                 "balance" : "0x0de0b6b3a7640000",
-                "code" : "0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff600052601061ffff61010037600051600055",
+                "code" : "0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff600052601061ffff600039600051600055",
```
pirapira, line 92: I feel like a detective here.

Is this change intended?  Even if unintended, it's fine if it's harmless.

GeneralStateTests/stMemoryTest/codeCopyOffset.json
```diff
@@ -2,10 +2,10 @@
     "codeCopyOffset" : {
         "_info" : {
             "comment" : "",
-            "filledwith" : "cpp-1.3.0+commit.dda985df.Linux.g++",
+            "filledwith" : "cpp-1.3.0+commit.97bd859a.Linux.g++",
             "lllcversion" : "Version: 0.4.14+commit.c2215d46.Linux.g++",
-            "source" : "src/GeneralStateTestsFiller/stCodeCopyTest/codeCopyOffsetFiller.json",
```
winsvega, line 7: you have to remove filler and filled test from that folder as well.
try running 
`./testeth -t GeneralStateTests`  
'./testeth -t BCGeneralStateTests'
and see that there are no errors

GeneralStateTests/stMemoryTest/codeCopyOffset.json
jwasinger: No.  it was unintended.  But you're right, it shouldn't matter for this test case (as long as source offset > source size, and the amount of data being written > 0)

PR #389, opened 2018-01-08
evmJit tests for opcodes

winsvega:
opcode arguments as variables

cpp PR https://github.com/ethereum/cpp-ethereum/pull/4744

### DIFF
--- a/src/GeneralStateTestsFiller/stArgsZeroOneBalance/README.md
+++ b/src/GeneralStateTestsFiller/stArgsZeroOneBalance/README.md
@@ -10,9 +10,9 @@ Pseudo code description:
 foreach opcodes as opcode
 {
  a = 0
-opcode (a,a, ...)
+ opcode (a,a, ...)
 
  a = 1
-opcode (a,a, ...)
+ opcode (a,a, ...)
 }
 ```


chfast:
Why this is so huge?

winsvega:
there are tests for every opcode that has input args.
I pass 2 options. 
1) input arg is a variable which value is zero
2) input arg is a variable which value is not zero

for every test it compiles a GeneralStateTest (fast execution)
and a hive blockchain test (all clients execution)

you are intersted in Filler.yml files

chfast:
1. Please do not name it "evmJit" tests. This name does not make sense.
2. We need it only for arithmetic opcodes, these test for BALANCE, CALLs, etc make little sense.
3. Why there are d0v0 and d0v1 variants?
4. Can't they be in a single file?

winsvega:
1. stOpcode
2. the more tests - the better
3. 4. because one test per file paradigm. and because two transactions in general test filler. one where variable is zero. and one where variable is non zero value.

chfast:
Better name: non const arguments.

> the more tests - the better

That's not true. This means we have to spend much more time on testing with very little gain.

> one where variable is zero. and one where variable is non zero value.

What does it mean? If an opcode takes 3 arguments, which one is zero?

winsvega:
first case:   variable a = 0 
opcode (a,a,a)

second case: variable a = 1
opcode (a,a,a)

I plan to rename test folder to `stOpcodes`

pirapira:
@winsvega the name is still not informative.  All state tests contain opcodes, no?  I'd go for `stNonConstArguments` or `stNonConstArgs`, again as @chfast suggested.

winsvega:
then it will be a complete test suite. stOpcode assume that some opCode specific tests might be added in future. Actually tests classification is rather a tricky problem.

pirapira:
@winsvega will you describe in English what the new tests do?  Out of that sentence, we can create a name.

I'm not requesting an exact classification.  I'm requesting an informative name.  Informative names carry ideas.

winsvega:
```
foreach opcodes as opcode
{
 a = 0
opcode (a,a, ...)

 a = 1
opcode (a,a, ...)
}
```

is it better than english?

pirapira:
@winsvega now turn that into English, and then we can figure out an informative name.

winsvega:
so this tests contain opcodes that has input arguments. but that arguments passed as output of another opcode. in this case its Balance. each test has 2 transactions. one makes balance to return 0 another one makes balance  to return 1. 

this is done so evmjit or other compiler would not define opcode arguments as constant but as a variable rather. because Pawel encountered some bugs in evmjit for this kind of scenario, I decided to make a test suite for such cases.

pirapira:
From the first paragraph, it can go like `stArgsZeroOneBalance`.  From the second paragraph, it can go like `stJitVariableArgs`.  Maybe you can come up with a better name.

It'd be nice to put a README in the new directory, containing these descriptions.

pirapira:
@winsvega wait for Travis builds before merging.

src/GeneralStateTestsFiller/stArgsZeroOneBalance/README.md
```diff
@@ -0,0 +1,18 @@
...
+This is done so evmjit or other compiler would not define opcode arguments as constant, but as a variable rather. Because evmjit encountered some bugs for this kind of scenario.
+
+
+Pseudo code description:
+
+```
+foreach opcodes as opcode
+{
+ a = 0
+opcode (a,a, ...)
```
chfast: Can you fix indentation here?

chfast:
Why haven't you amend the last commit?

winsvega:
I used github web edit tool

chfast:
Bad excuse :)

PR #393, opened 2018-01-14
Bugfix in filename conversion in convert_wast

lrettig:


lrettig:
This PR got polluted with later changes, closing in favor of #396 which is in its own branch

PR #400, opened 2018-01-17
Fix convert_wast to allow empty code field for EOAs

lrettig:


src/GeneralStateTestsFiller/stEWASMTests/utils/convert_wast.py
```diff
@@ -28,7 +28,7 @@ def convert_state_test(file_name):
   for test_case in state_test.keys():
     # conver pre account code
     for addr in state_test[test_case]['pre']:
-      if state_test[test_case]['pre'][addr]['code'] != ['']:
+      if state_test[test_case]['pre'][addr]['code'] != '':
         state_test[test_case]['pre'][addr]['code'] = '0x'+wat2wasm(state_test[test_case]['pre'][addr]['code'], file_name)
 
     if state_test[test_case]['transaction']['data'] != [''] and state_test[test_case]['transaction']['to'] == '':
```
axic, line 34: Should this also be `''`?

lrettig: Yea, not sure. @jwasinger san? What was the purpose of the array of empty string to begin with?

jwasinger: The answer is yes.  Pyyaml was returning [''] as the value for empty fields at some point in time.  Not sure what I did to change that...

hugo-dc: I think for data is ok to have [''] because data is a list:
https://github.com/ethereum/tests/blob/wasm-tests/src/GeneralStateTestsFiller/stEWASMTests/utils/getAddressFiller.yml#L45

but code is just one element: 
https://github.com/ethereum/tests/blob/wasm-tests/src/GeneralStateTestsFiller/stEWASMTests/utils/getAddressFiller.yml#L14

lrettig:
Good catch Hugo!

PR #391, opened 2018-01-13
Add GeneralStateTests/stEWASMTests placeholder

axic:
Required by https://github.com/ethereum/cpp-ethereum/pull/4764

PR #401, opened 2018-01-22
add stEWASM test folder

winsvega:


axic:
Duplicate of #391.

winsvega:
yes but this one comes from local branch rather than external. so the CPP PR would become green with no issues

pirapira:
This is the one tested in `cpp-ethereum` Travis run.

PR #402, opened 2018-01-23
Try not only RETURNDATASIZE but also RETURNDATACOPY

pirapira:
This implements
https://docs.google.com/spreadsheets/d/1xat7UI8GtB4ZGVdlK5_XQSHJZaMThi4SrlcL8XMZb5Q/edit#gid=1605675414&range=C59


winsvega:
I guess this was implemented in https://github.com/ethereum/tests/pull/375

(init code) is for transaction init code. testeth now supports lll compilation in trnsaction data field.

pirapira:
Oh I see.

But how do I call RETURNDATACOPY after a create transaction?  In an another transaction?

winsvega:
the idea was to run CREATE opcode from transaction init code. then it is the same.
just put the code from contract to transaction data. it will run it during transaction execution from a new account that is being created. 
you could check storage of this new account.

pirapira:
I see.  That's an interesting test case.

PR #409, opened 2018-01-30
Travis CI: Build PRs once

chfast:


.travis.yml
```diff
@@ -3,6 +3,10 @@ python:
   - "2.7"
 node_js: 
   - "8"
+branches:
+  only:
+    - master
+    - develop
```
axic, line 9: Though it would make sense renaming  `develop` to `master` before merging this?

chfast: It handles both. Later you can remove "develop" but it's not a big deal to keep it there forever.

winsvega: will this build two branches?

pirapira: no.

PR #412, opened 2018-01-31
Optimize and Fix the --filltests

winsvega:
*stCallCodes


winsvega:
its going to be a huge PR. for the https://github.com/ethereum/cpp-ethereum/pull/4823

PR #419, opened 2018-02-27
revert in CREATE in init code, followed by returndatasize/copy

pirapira:
Implements https://docs.google.com/spreadsheets/d/1xat7UI8GtB4ZGVdlK5_XQSHJZaMThi4SrlcL8XMZb5Q/edit#gid=1605675414&range=C59

### DIFF
--- a/JSONSchema/st-filler-schema.json
+++ b/JSONSchema/st-filler-schema.json
@@ -20,11 +20,16 @@
                 { "$ref": "#/definitions/HexQuantity" }
             ]
         },
-        "HexDataOrEmpty": {
+        "TxData": {
           "oneOf": [
               { "$ref": "#/definitions/HexData" },
-              { "$ref": "#/definitions/EmptyString" }
+              { "$ref": "#/definitions/EmptyString" },
+              { "$ref": "#/definitions/LLLCode" }
           ]
+        },
+        "LLLCode" : {
+            "type" : "string",
+            "pattern" : "^{.*}$"
         },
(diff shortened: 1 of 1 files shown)


veox:
For ref: `py-evm` had a hiccup on this one.

Here's what I think is going on it the test: https://github.com/ethereum/py-evm/pull/1224#issuecomment-418775512.

From [gitter chat](https://gitter.im/ethereum/tests?at=5b900a24f5402f32aae16497), the test rationale _might_ be focusing on pre-Byzantium execution, not Byzantium itself:

> maybe [Yoichi] was afraid that returndatasize might be 0 and so this way he checks that transaction fails on <Byzantium . 

The "weird" part about the test is having a {zero-`balance`, no-`code`, zero-`nonce`, some-`storage`} account in the `pre` section:

https://github.com/ethereum/tests/blob/298749a4b2d0d9e932057dabb1ebb38ddb84f116/BlockchainTests/GeneralStateTests/stRevertTest/RevertInCreateInInit_d0g0v0.json#L94-L102

The call to `REVERT` fails on pre-Byzantium, but is expected to have the storage filled on Byzantium:

https://github.com/ethereum/tests/blob/298749a4b2d0d9e932057dabb1ebb38ddb84f116/BlockchainTests/GeneralStateTests/stRevertTest/RevertInCreateInInit_d0g0v0.json#L77-L85

The difference between `py-evm` and the test-as-specified is in gas use, and is discussed in more detail in the `py-evm` PR (first link of this comment).

PR #422, opened 2018-03-01
Return predefined data

jwasinger:
Returns data predefined in a "data" segment.  This is an untested edge case.


jwasinger:
ah wrong repo.. :)

PR #428, opened 2018-03-09
Test whether a failed create transaction reverts account deletion

pirapira:
A modification of stRevertTests/RevertInCreateInInit.  That test, for EIP158 only, accidentially tested the case where a contract creation transaction touches an empty account and then fails.  This one tests the
same thing not just for EIP158 but any network thereafter.


winsvega:
ready to merge?

PR #434, opened 2018-03-21
[wip]Better dup key validator

pirapira:
I noticed I was using a wrong package for detecting duplicate keys (it was missing some duplicate keys).  I found a better library.


pirapira:
Whoa, I see 80 failures.

PR #435, opened 2018-03-22
Better dup detector

pirapira:
I noticed I was using a wrong package for detecting duplicate keys (it was missing some duplicate keys). I found a better library.


pirapira:
See success in https://github.com/ethereum/testeth/pull/22 before merging.

PR #442, opened 2018-03-29
call ID precompiled contract with smaller output size than input size

pirapira:
and check RETURNDATASIZE is the size of the input.


pirapira:
The idea comes from https://docs.google.com/spreadsheets/d/1xat7UI8GtB4ZGVdlK5_XQSHJZaMThi4SrlcL8XMZb5Q/edit#gid=1605675414&range=C42

Issue #415, opened 2018-02-13
Remove VM tests with CALL, CREATE, SELFDESTRUCT and the like, translating them into GeneralStateTests


pirapira:

ehildenb:
@pirapira any recent thoughts on this? I don't think we should do away with VMTests, but perhaps could limit their scope. It doesn't feel good to have KEVM have two versions of several opcodes, one just for VMTests.

Perhaps the VMTests that only deal with VM-related stuff could remain (eg. arithmetic, local memory ops, program counter ops), and anything that deals with Ethereum-specific state could be translated into GeneralStateTests instead?

I've become interested in this because I'm in the middle of refactoring the KEVM control-flow to make it more uniform to make doing proofs with KEVM easier (and also to support more uniform typed exceptions), but there are several spots where special cases are made for VMTests mode which are causing me headache.

ehildenb:
Note that I do like having the VMTests around as an early target for implementers who have only made it through implementing the basic VM functionality. Similarly, someone might implement a wasm VM without making it extend to the ewasm, and would want to be able to run the wasm test-set without worrying about the extra host-functions added for ethereum.

pirapira:
@ehildenb I have exactly the same reasons for keeping VMtests.  I changed the title of this issue, to remove only those problematic opcodes from VMtests.

ehildenb:
Ok, then I guess we should discuss which tests should specifically be removed from the test-suite. At least the ones dealing with `CALL*`, `CREATE`, and `SELFDESTRUCT` I think. I will try to run the VMTests in non-VMTEST mode on KEVM and get a list of tests that break then post them here.

ehildenb:
Here is a list of tests that fail when I disable all VMTESTS specific stuff: [disable-vmtests-mode-initial.txt](https://github.com/ethereum/tests/files/1872541/disable-vmtests-mode-initial.txt)

There are 40 total (including BLOCKHASH tests as well).

ehildenb:
Full list of "problematic" opcodes: `BALANCE`, `EXTCODESIZE`, `EXTCODECOPY`, `CALL*`, `CREATE`. Actually doesn't seem to be any problem with `SELFDESTRUCT`.

ehildenb:
[...]

PR #444, opened 2018-04-04
Remove VM tests using the mock BLOCKHASH/BALANCE instruction just for VM tests

pirapira:
See the discussion https://github.com/ethereum/tests/issues/415#issuecomment-378593659

### DIFF
--- a/src/VMTestsFiller/README.md
+++ b/src/VMTestsFiller/README.md
@@ -1,0 +1,21 @@
+VM tests
+========
+
+VM tests test one instance of EVM (which contains one stack of 256-bit words, and one memory space).
+
+Operations accessing the world state should be tested in GeneralStateTests or BlockchainTests instead of in VM tests.
+
+So VM tests should not contain
+
+* `BALANCE`
+* `CREATE`
+* `CREATE2`
+* `CALL`
+* `CALLCODE`
+* `STATICCALL`
+* `DELEGATECALL`
+* `EXTCODESIZE`
+* `EXTCODECOPY`
+* `BLOCKHASH`
+
+Previously, some VM tests contained these, and clients were supposed to implement some mock of the world state (especially, the block hash of a block is supposed to be the hash of the block number).


ehildenb:
@pirapira I believe the following tests also use `BLOCKHASH`
https://github.com/kframework/evm-semantics/commit/dea8f07b92e3c9326da4c55912eed620861d2f54

pirapira:
I'll have a look into these four, and later modify the vmtest runner in `eth-isabelle` so it fails on any `BLOCKHASH`.

pirapira:
Yes, these should be removed too.
* 201503102037PYTHON.json uses `BALANCE`.
* 201503102148PYTHON.json uses `BALANCE`.
* 201503102300PYTHON.json uses `BALANCE`.
* 201503110050PYTHON.json uses `BALANCE`

pirapira:
And then, 201503120317PYTHON.json uses `BALANCE`.

pirapira:
I can detect those VM tests with disallowed instructions using this modified `testeth`. https://github.com/pirapira/testeth/tree/vmtest-filter

ehildenb:
How about also removing the `callcreates` field of the tests?

pirapira:
@ehildenb that's a good idea.  I created a github issue https://github.com/ethereum/tests/issues/449

Matthalp-zz:
@pirapira If you are removing support for `BALANCE` here (and `SELFDESTRUCT` in another PR) do you also want to remove `EXTCODESIZE` and `EXTCODECOPY` too?

They are in this [PR](https://github.com/ethereum/cpp-ethereum/pull/4731/files) that I was (finally) going to come back and try to land this week. Let me know how I should proceed.

pirapira:
@Matthalp yes, `EXTCODE*` is also being removed from VM tests.  If there is a way to fire an error for these instructions in VM tests, that would be nice.

Matthalp-zz:
@pirapira I think https://github.com/ethereum/cpp-ethereum/pull/4731 will fix the ambiguities around `BALANCE`, `EXTCODECOPY`, `EXTDATASIZE`, and `SUICIDE`, but not the `CALL`/`CREATE`s.

Could you help me understand what you mean by "fire an error"? Does this just mean that the test case is able to fail if the incorrect behavior occurs?

pirapira:
Because VM tests are supposed to test one instance of EVM only, and not to access the world state.

pirapira:
Right.  Still VM tests are useful when you are implementing a new VM.

PR #445, opened 2018-04-05
Remove VM random tests that contain `BALANCE`

pirapira:
because these require special execution mode for VM tests.

See https://github.com/ethereum/tests/issues/415


VMTests/vmRandomTest/201503110226PYTHON_DUP6.json
```diff
@@ -1,51 +0,0 @@
-{
-    "201503110226PYTHON_DUP6" : {
```
winsvega, line 2: this test was failing on python someday

pirapira: Is that a reason to keep the test or to remove the test?

winsvega: was a long time ago. but if it contains CALL CREATE SELFDESTRUCT then it probably does not make sense to keep it in VMTests

pirapira: It contains `BALANCE` so it's being removed.

winsvega: the balance is easy to replace with any other opcode or remove it. maybe the issue was in other part of the code.

pirapira: It doesn't matter if `BALANCE` is easy or not.  When a client is running a VM test, the client should contain no world state, so the client should not be able to run BALANCE.

pirapira: So, I'll move this test to GeneralStateTest.

winsvega: ok

pirapira: I added this one as a generalstate test in this PR.

pirapira:
I translated the test into a GeneralStateTest.

src/GeneralStateTestsFiller/stBugs/201503110226PYTHON_DUP6Filler.json
```diff
@@ -0,0 +1,58 @@
...
+        },
+        "expect" : [
+            {
+	            "indexes" : {
+	                "data" : -1,
+	                "gas" : -1,
+	                "value" : -1
+	            },
+	            "network" : [">=Frontier"],
+                "result" : {
```
winsvega: nothing really have changed here. perhaps better put gas consumed or nonce change into the expect section ?

pirapira: I wanted to change nothing, but `BALANCE` cannot live in VMtests, so the test moved.  It just moved without any additional guarantees.

winsvega: could be poot into random tests then

pirapira: I'll do that.

pirapira:
@winsvega I moved the test into `stRandom2`.

PR #454, opened 2018-04-18
add vm test for arithmetic (add) after byte

jwasinger:
Designed to catch issues like https://github.com/ethereumjs/ethereumjs-vm/pull/293

### DIFF
--- a/VMTests/vmBitwiseLogicOperation/byteBN.json
+++ b/VMTests/vmBitwiseLogicOperation/byteBN.json
@@ -1,0 +1,52 @@
+{
+    "byteBN" : {
+        "_info" : {
+            "comment" : "",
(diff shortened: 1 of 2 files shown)


axic:
Would it make sense to have a script generate test cases bases on the arithmetic/logic/bitwise instruction list and their in/out arguments? The generator would need to combine at least two of those instructions in a single test case. We could also nominate `add` to be always the second op.

pirapira:
@axic those tests are welcome.  It sounds like those tests take some space in the tests repo but they should run quite fast.

winsvega:
not via RPC on cpp they not

winsvega:
I guess we could come to a general test source format from which you just generate blockchain tests and vm tests.

winsvega:
@qbzzt need to review the OP issue and possibly cover with the test cases

PR #460, opened 2018-05-28
VMTests/.../exp8Filler: add test of EXP 0 0

ehildenb:
@winsvega I've added a VMTest. I think this is where it belongs because it only tests VM functionality, nothing in the client. I have not filled it, let me know if I should, or if you can and then just add another commit onto this.


axic:
Would be nice to have state tests if possible because some clients only deal with state tests and not vm tests.

pirapira:
@axic I started adding a GeneralStateTest.

axic:
@pirapira we (well, @ehildenb mostly) have started work to have a script automatically convert vm tests to state tests fillers, so both can be kept.

pirapira:
ahm, then, I drop my manual translation, and work with @ehildenb on the automatic translation (anyway I'm flying to his town this Friday).

winsvega:
convert vmtestFillers to stateTestFillers?

ehildenb:
@winsvega yes, convert VMTestsFiller to GeneralStateTestsFiller. Then they can just be filled using standard techniques. See #461 as the initial PR in that direction.

axic:
@jwasinger do you want to fill this?

axic:
@winsvega but conversion here means it is done by a script and the vmtests can be kept.

jwasinger:
> @jwasinger do you want to fill this?

Sure.

pirapira:
@winsvega do you want a separate PR for a filled test?  Or, shall we wait here for the filling?

ehildenb:
@pirapira I have filling working now, so I can fill the test myself. But I was hoping to get the other sanitation PRs in first (like #462 and #466 ), so that I can exercise the new sanity checks on this PR.

winsvega:
test sources (fillers) and tests must be on the same PR.
testeth will fail if there are tests without filler or fillers without tests or fillers and corresponding tests has different source hash.

ehildenb:
This run *should* fail, because of poorly formatted JSON in the repo. Next I'll fix the formatting and make sure it fails because of invalid JSON Schema, then because of unfilled tests.

pirapira:
It's nice you are trying out the new checkers.

ehildenb:
Do not merge yet, I am investigating a few things still:

1.  How to make sure that filled version of test actually exists (currently I have not committed the filled version of exp8 test, but tests pass).
2.  Failing of `sdiv_dejavu` on cpp-ethereum.

ehildenb:
@pirapira I think I have figured out what happenned with `sdiv_dejavu`.

Commit ddba26a8c refilled the test, but did not give it a post section. This is because it was generating the nonsensical bytecode like `606005` for `PUSH1 0x05` (the same LLL issues we saw). Because there was no post, the fact that the `expect` section has not been met was not caught by either cpp-ethereum or KEVM.

After I fixed the LLL code, it gave a post section which contains `0xffffffff...` as the result of the division (the `expect` section contains `0x00` though). Both cpp-ethereum and KEVM agree that this is correct, though I should work out by hand as well (expression is `(sdiv (sub 0 9) 5)`). I expected that when filling the tests, the `expect` section would make it through to the filled test, but it's actually the result of cpp-ethereum which makes it through. So cpp-ethereum passes the filled test, as does KEVM, but cpp-ethereum reports error during the filling process.

So, I think, if the calculation by hand shows that it's correct, then I think I can update the `expect` section of the test, and refill, and it should be OK.

ehildenb:
Here is an interaction with bc that seems to confirm it is correct

```
obase=16
us0 = (2^256) - 9
us1 = 5
(us0 - (2^256)) / us1 + (2^256)
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
```

Indeed, what is happening is `-9 / 5`, which seems like it should be `-1`.

ehildenb:
Alright. This branch enables sanitation checks on the VMTests (checking that they are filled correctly, etc...). 

I have hcanged the expect section of `sdiv_dejavu`, added the test `exp8` which tests `exp 0 0`. Please review those files with scrutiny. The remaining changes are to enable sanitation of the VMTests.

Because of how much is changing in the way of formatting, I'm waiting for future PRs to make the same changes to the GS and BChain tests (if people want them). I propose that for filling the tests, whatever command is used as the "canonical" command is stored in the `Makefile` for ease of reproducibility.

Currently, to fill the VMTests, I run `make VMTests.fill`. If you want to fill another testset, it`s `make GeneralStateTests.fill` for example. I will push a file documenting this.

PR #461, opened 2018-05-28
Vmtests validator

ehildenb:
Address https://github.com/ethereum/tests/issues/411

The script `test.py` will eventually also do more (eg. convert VMTests to GeneralStateTests). Initial PR in that direction.

@axic

PR #462, opened 2018-05-29
Hex encoding and invalid json

ehildenb:
Fixes that arose from #461. @pirapira and @axic.


ehildenb:
The tests may need to be re-filled given these changes.

ehildenb:
This is a refilled version of the VMTests, along with some things fixed in the VMTests and BlockchainTests. One problematic thing is the test `sdiv_dejavu` which seems to not have the same expected behavior anymore with cpp-ethereum!!! (have not dug into the cpp-ethereum code)

Other tests I had to convert to raw bytecode because the compiler from LLL -> bytecode was throwing a stack underflow error (perhaps doing too much an overapproximation of some static analysis over the basic blocks?). Have discussed with @axic.

@pirapira what should I do?

ehildenb:
Note that while tests are passing here, the test `sdiv_dejavu` does *not* pass on cpp-ethereum. It passes on KEVM.

ehildenb:
I've split the changes which should be inspected into 8fc1201.

ehildenb:
Summary of changes which I think are significant in the VMTests changes. Once we check all these, and have a similar discussion about the changed blockchain fillers, I think we can safely merge. Below when I say "KEVM agrees", I mean "agrees with cpp-ethereum".

- [ ] `sdiv_dejavuFiller` was converted from bad asm to good asm, which means that things like `PUSH1 0x05` became more like `0x05`. The old generated bytecode had ops like `606005`, which is nonsensical and probably an artifact of `PUSH1 0x05` being translated into something like `push push(05)`, hence the `606005`. With the corrected assembly, the test now passes instead of throwing an error, meaning that `sdiv_dejavu` now has a post-section. KEVM agrees that the new test passes.

- [ ] `gasOverFlowFiller` was similar to `sdiv_dejavuFiller`, except the result both before and after the fix was still an exception being thrown. No changes to post-state, just to code chunks.

- [ ] `DynamicJump_value{1,2,3}Filler` had multiline ASM code in them, which is invalid JSON. The old filled tests had empty code blocks (not sure why, perhaps the invalid json meant LLL was not invoked on them, @axic?). KEVM agrees that the newly filled tests pass.

- [ ] `DynamicJump_valueUnderflowFiller` also had multiline ASM code in it. However, because the invalid JSON was resulting in empty code blocks in the filled tests, no code was being run, which meant that the value underflow was not being triggered (which meant that a post-state was present in the filled test). Now the post state is gone, and KEVM agrees that exception is thrown.

- [ ] `jumpHighFiller` had some old unsupported syntax for inline assembly in LLL code. It's been converted, no change to filled test. KEVM still agrees on this test.

- [ ] `stack_loopFiller` and `swapjump1Filler` both have obvious code paths that can result in stack underflow. I guess LLL has since been taught to detect this, and is reporting stackunderflow errors at compilation, resulting in no test being generated. I've saved the old code as `_code` and manually compiled it to bytecode as `code`. In both cases, before there was no code generated, and now there is. The poststate for `swapjump1` remains, but `stack_loop` now throws and so there is no poststate. KEVM agrees with the new behavior of both.

The blockchain tests just have changes in the `stateRoot`, where there was not a `0x` at the front of many of them which needs to be there for parsing. They should be refilled as well to update the source hashes.

@pirapira ?

pirapira:
Let's separate the `sdiv_dejavu` and merge the rest.

ehildenb:
@pirapira I've pulled out the changes to blockchain tests into a different branch, though am not sure why you want to separate the changes to `sdiv_dejavu` from the rest. I can pull those changes out as well if needed.

ehildenb:
Ah, I see, it's because `sdiv_dejavu` actually does not pass on cpp-ethereum. forgot.

ehildenb:
@pirapira I've removed the sdiv_dejavu changes. I've also snuck in a `Makefile` which allows filling the tests by calling `make fill-TESTSETNAME`. This is advantageous because then we can call `make fill-TESTSETNAME -j4` and fill in parallel, which I guess Boost testing framework cannot do.

winsvega:
if it touches only VMTests
merge at @pirapira  aproval

axic:
Those LLL fixes happened September of 2016 and 2017, but seemingly the tests weren't refilled since. `PUSHn` keywords in assembly were disabled in 2017 as they were just pushing the opcode without an immediate and number literals were pushing both the opcode and the immediate, so `PUSHn` was useless.

PR #465, opened 2018-05-30
Check test filled

ehildenb:
To wait for #461. This checks that the filled tests are up to date with the filler, and fails the CI if not.

PR #466, opened 2018-05-31
Test validator script

ehildenb:
Subsuming #465 and #461 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,11 @@ install: "pip install jsonschema pysha3"
 script:
 
 # check valid JSON/well formatted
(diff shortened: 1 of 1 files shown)


ehildenb:
I've also disabled the json validity checks that .travis.yml was doing because ./test.py does them anyway.

winsvega:
same here. if it touches only VMTests merge at @pirapira  approval

pirapira:
I skimmed commit by commit.  Not quite only VMTests, but looks good.

pirapira:
Let me break some test files locally and see what the new validator does.

PR #467, opened 2018-05-31
Test converter

ehildenb:
I'm looking at making a test converter between the formats, but I don't want to build any advance knowledge of crypto or how to  make blocks properly into this little script. I think perhaps a better solution would be to just insert a new `_info` field into the `src/BlockhcainTestsFiller/GeneralStateTests`, and similarly in `src/BlockchainTestsFiller/VMTests` so that the script can check on PRs into this repo that the external tool properly filled the fillers. This would be similar to what #466 is doing to check that the filled versions are up-to-date.

Thoughts @pirapira and @winsvega?

### DIFF
--- a/test.py
+++ b/test.py
@@ -134,18 +134,38 @@ def checkFilled(jsonFile):
             if fillerHash != hashFile(fillerSource):
                 _logerror("Test must be filled:", jsonFile)
 
+# Test format converters
+
+def convertvm2bchain(jsonFile):
+    _logerror("Unimplemented:", jsonFile)
+
+def convertgs2bchain(jsonFile):
+    _logerror("Unimplemented:", jsonFile)
+
+def convert2bchain(jsonFile):
+    if jsonFile.startswith("./src/VMTestsFiller/"):
+        convertvm2bchain(jsonFile)
+    elif jsonFile.startswith("./src/GeneralStateTestsFiller/"):
+        convertgs2bchain(jsonFile)
+    else:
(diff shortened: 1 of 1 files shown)


winsvega:
what kind of a new field? 
actually I am thinking of making a general test source format. from which retesteth could generate the test.
1) a blockchain test 
2) a vm test
3) state test format would go obsolite
4) transaction test format would go obsolite

ehildenb:
Just a new field `"_info": { "source" : PATH_TO_SOURCE , "sourceHash": HASH_OF_SOURCE }` similar to how the filled tests have. I would propose adding them to the tests in `src/BlockchainTestsFillers/VMTests` and `src/BlockhcainTestsFillers/GeneralStateTests`.

Sure, maybe we'll obsolete as we move forward, but we will still need CI infra for the test repo either way. I think it's an orthogonal discussion.

winsvega:
wait, are you talking about source field for the test fillers? 
then it looks strange. a test has a source that points to a filler. a filler has source that points to another filler? 

I really need help with rpc methods for making the current tests over RPC work. 
the rpc methods we use at the moment are mess. the cpp needs lots of optimization and chages. and even a small PR to cpp takes ages to be merged.

pirapira:
@winsvega I don't think it's strange.  All files except manually written should explain how they were produced, I think.

winsvega:
*Filler.json is supposed to be manually written

pirapira:
@winsvega Regardless of the original design, some filler files are automatically generated.  I would not accept manually editing those translated test.  Your "supposed to be" is not true for a long time and should be forgotten.

PR #470, opened 2018-06-02
[BLOCKED] Vmtests 2 generalstatetests

ehildenb:
Do not merge until #469 is merged.

This converts the VMTests to GeneralStateTests in the simplest way possible. I do not know that it works correctly yet, but it converts existing VMTests all over without throwing errors.

The idea here is that operation of many low-level VM stuff is important, and should be tested in isolation. Hopefully we can evolve the VMTests to only test core functionality, I would like to even see us avoid using `SSTORE` to inspect poststate eventually (instead preferring to inspect the stack/local memory directly).

### DIFF
--- a/JSONSchema/st-filler-schema.json
+++ b/JSONSchema/st-filler-schema.json
@@ -81,7 +81,7 @@
                 "storage": {
                     "additionalProperties": false,
                     "patternProperties": {
-                        "^0x[0-9a-f]+": {
+                        "^0x[0-9a-f]*": {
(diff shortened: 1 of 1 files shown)


winsvega:
up

winsvega:
vmtests are depricated, however having all test cases in state tests is good. 
state tests then also get depricated but the state test fillers not. fillers could be converted into blockchain tests.

winsvega:
https://github.com/ethereum/tests/issues/593

winsvega:
https://github.com/ethereum/tests/issues/593

PR #471, opened 2018-06-05
VMTests/vmArithmeticTest/arith1.json remain gas is incorrect

jjyr:
**Gas remain incorrect**
[arith1](https://github.com/ethereum/tests/blob/develop/VMTests/vmArithmeticTest/arith1.json) test case expect 17777087 gas remained after exec, but the gas should be 17777047. 
I checked the result on go-ethereum evm.

**Reason**
Cause [EIP160](https://github.com/ethereum/EIPs/blob/984cf5de90bbf5fbe7e49be227b0c2f9567e661e/EIPS/eip-160.md) updated EXP gas cost, the cost of EXP from 20 gas to 60 in this case, so the test case gas should minus 40,  it's makes sense.

### DIFF
--- a/VMTests/vmArithmeticTest/arith1.json
+++ b/VMTests/vmArithmeticTest/arith1.json
@@ -25,7 +25,7 @@
             "origin": "0xcd1722f2947def4cf144679da39c4c32bdc35681",
             "value": "0x0de0b6b3a7640000"
         },
-        "gas": "0x0f41bf",
+        "gas": "0x0f4197",
         "logs": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
         "out": "0x0000000000000000",
         "post": {


ehildenb:
The VMTests are against the FRONTIER gas schedule, not the most recent gas schedule.

I've also argued that nothing about gas costs should be asserted in the VMTests.

Either way, I think we should stick to the FRONTIER schedule until we either decide to (i) support multiple schedules in the VMTests, or (ii) remove gas checks from the VMTests.

ehildenb:
But I don't think we should update it for only a single VMTest. If you want to update it, it should be for all of them at the same time.

jjyr:
Gas check is still helpful, since there no specific tests to guarantee gas calculation.
I think the [documentation](http://ethereum-tests.readthedocs.io/en/latest/test_types/vm_tests.html) should mention about the FRONTIER gas schedule is used in tests.
And maybe we should add "gas_schedule" field in test json file to specific which gas schedule to use. 
like:
``` json
"gas": ...,
"gas_schedule": "FRONTIER"
```
or support multiple schedule

``` json
"gas_schedule": {
  "FRONTIER": "0x0f41bf",
  "EIP160": "0x0f4197"
}
```

ehildenb:
Well, the argument would be that the VMTests are just for core VM functionality (in the case of EVM, just the wordstack and the local memory operations), and the other test sets (like GeneralStateTests and BlockchainTests) are for other faucets of EVM execution (including things like Gas costs).

I go back and forth on it though. Either way, if we have gas costs, it should be schedule dependent. But probably model it off how the GeneralStateTests do it if so.

winsvega:
we have state tests that support different gas scheme

jjyr:
Good

PR #472, opened 2018-06-06
Vmtests valid

ehildenb:
This enables all the checkers that are able to pass at the moment, documents the ones that cannot (in the `Makefile`), and documents how to use the `Makefile` in the `README`.

@pirapira and @winsvega


ehildenb:
If the tests pass and you guys approve, I think this is ready for merge.

PR #475, opened 2018-07-03
Create test

winsvega:


src/GeneralStateTestsFiller/stCreateTest/CreateOOGafterInitCodeFiller.json
```diff
@@ -0,0 +1,83 @@
...
+                "code" : "",
+                "nonce" : "0",
+                "storage" : {
+                }
+            },
+	    "b94f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
+                "balance" : "0",
+		"nonce" : "0",
+		"//code" : "(SSOTRE 1 1) RETURN",
+		"code" : "{ (MSTORE 0 0x6460016001556000526005601bf3) (CREATE 0 18 14) }",
```
chfast: What `0x6460016001556000526005601bf3` does?

winsvega: (SSOTRE 1 1) RETURN 

returns sstore 1 1 as a contract deployment code

src/GeneralStateTestsFiller/stCreateTest/CreateOOGafterInitCodeFiller.json
```diff
@@ -0,0 +1,83 @@
...
+                "balance" : "1000000000000",
+                "code" : "",
+                "nonce" : "0",
+                "storage" : {
+                }
+            },
+	    "b94f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
+                "balance" : "0",
+		"nonce" : "0",
+		"//code" : "(SSOTRE 1 1) RETURN",
```
chfast: You have a typo here.

chfast:
Only 4 cases partly reproduce the issue.
The problem is the RETURNDATA buffer is not inspected after the CREATE so the state test are still passing.

winsvega:
Return data after failing deployment?

chfast:
Yes

chfast:
Yes. In the buggy implementation the runtime code is put in returndata buffer. But this is not visible in the final state because the caller does not do anything with this data. At least add `SSTORE(0, RETURNDATASIZE)`.

winsvega:
added a test case:
1) callcode -> ( create | OOG on deployment ) | check returndata 

is this triggering your bug?

src/GeneralStateTestsFiller/stCreateTest/CreateOOGafterInitCodeReturndata3Filler.json
```diff
@@ -0,0 +1,75 @@
...
+		"nonce" : "0",
+		"code" : "{ (CALLCODE (GAS) 0xb94f5374fce5edbc8e2a8697c15331677e6ebf0b 0 0 0 0 32) (RETURNDATACOPY 0 0 32) [[ 1 ]] (MLOAD 0) }",
+		"storage": {
+		}
+            },
+	    "b94f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
+                "balance" : "0",
+		"nonce" : "0",
+		"//code" : "(SSTORE 1 1) RETURN",
+		"code" : "{ (MSTORE 0 0x6460016001556000526005601bf3) (CREATE 0 18 14) }",
```
chfast, line 54: You were supposed to check returndata after the `CREATE`, not to the higher level of `CALLCODE`.

src/GeneralStateTestsFiller/stCreateTest/CreateOOGafterInitCodeReturndataFiller.json
```diff
@@ -0,0 +1,68 @@
...
+                "code" : "",
+                "nonce" : "0",
+                "storage" : {
+                }
+            },
+	    "b94f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
+                "balance" : "0",
+		"nonce" : "0",
+		"//code" : "(SSTORE 1 1) RETURN",
+		"code" : "{ (MSTORE 0 0x6460016001556000526005601bf3) (CREATE 0 18 14) [[ 1 ]] (RETURNDATASIZE) (RETURNDATACOPY 0 0 32) [[ 2 ]] (MLOAD 0) }",
```
chfast, line 47: This is wrong, because the `RETURNDATACOPY` will cause the failure. You are not allowed to access more that at was returned and `CREATE` returns nothing.

chfast:
Still wrong. Non of these reproduces the case. I marked single test that is almost good and can be easily fixed.

src/GeneralStateTestsFiller/stCreateTest/CreateOOGafterInitCodeReturndataFiller.json
winsvega: I know.

src/GeneralStateTestsFiller/stCreateTest/CreateOOGafterInitCodeReturndata3Filler.json
winsvega: Right. It is just that if I put any more instructions father it wont OOG anymore.

PR #476, opened 2018-07-25
Fix typo name of a test in VMTests

hqueue:
- Rename a file `iszeo2Filler.json` to `iszero2Filler.json`
- Also rename a test from `iszeo2` to `iszero2`

Signed-off-by: Hyung-Kyu Choi <hqueue@users.noreply.github.com>


hqueue:
Oops. Didn't know that `src json` filie and `generated json` file should be included in the same PR.
If you don't mind, I will update this PR with a `generated json` file from `src json`. However it may take some time, since it's my first time to generate tests from `src` file.

winsvega:
up

PR #485, opened 2018-08-10
restore the rlp test

winsvega:


axic:
This rolls back #478. What was the point merging #478 if it is rolled back? Which is the valid value? IIRC based on the Yellow Paper #478 seemed correct.

axic:
cc @holgerd77

gumb0:
RLP encoding of 0 is encoding of empty byte array, so 0x80 is correct.

0x00 is encoding of byte array of length 1 containing one byte 0, not encoding of integer 0.

hamdiallam:
Why is `0` considered an empty byte array? 

`rlp.encode(1).toString('hex') == '01'`. `1` is interpreted as `0x01`. I do not see why we should treat `0` any differently. 

I think `[]` or `''` should be interpreted as an empty byte array

winsvega:
I think YP states exactly that. Rlp(0) = 0x00

However https://github.com/ethereum/wiki/wiki/RLP

hamdiallam:
I used that wiki when learning on RLP and those examples seem to have been added after.

Inconsistent implementation imo

hamdiallam:
I used that wiki when learning on RLP and those examples seem to have been added after.

Inconsistent implementation imo

PR #500, opened 2018-09-07
Add new touched cleanup case

mattdean-digicatapult:
Regression test associated with https://github.com/ethereumjs/ethereumjs-vm/pull/335

First call marks an empty account as touched. Second call errors OOG. Transaction should still cleanup empty account.

This fails in ethereumjs-vm due to the incorrect handling of the touched set when an error occurs.

Apologies if I've put this in the wrong place, this is my first time writing an evm test case and I was just feeling around a bit for conventions.


winsvega:
use  --filltests --fillchain to create a copy of this test for hive. 
please make sure that you update testeth with latest origin/master branch

Issue #504, opened 2018-09-10
Is TransactionNonceCheck state test incorrect?


germsvel:
I was looking through the [transaction nonce check state test](https://github.com/ethereum/tests/blob/develop/GeneralStateTests/stTransactionTest/TransactionNonceCheck.json) and its [filler](https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stTransactionTest/TransactionNonceCheckFiller.json). And I can't quite understand why it creates the beneficiary's account `2adc25665018aa1fe0e6bc666dac8fc2697ff9ba ` (before EIP-158).

The `pre` section does not have the account, but the `post` section does. It seems to me that since the sender account's nonce is `10` and the transaction nonce is `0`, the transaction should never get executed because it fails step (3) of the initial tests of intrinsic validity outlined at the beginning of Section 6 of the Yellow Paper. 

![screen shot 2018-09-10 at 11 29 36 am](https://user-images.githubusercontent.com/3245976/45308405-63d6b900-b4ef-11e8-94af-46810ae15567.png)

Someone in gitter mentioned that perhaps state tests should not have tests with invalid transactions, so I'd be happy to open a PR removing the test if that's the case.

germsvel:
After looking at some of [mana's](https://github.com/poanetwork/mana) Frontier state test failures, it seems that all of the following are tests with invalid transactions, 

```
stCallCreateCallCodeTest/createNameRegistratorPerTxsNotEnoughGas
stInitCodeTest/NotEnoughCashContractCreation
stRandom2/201503110226PYTHON_DUP6
stTransactionTest/CreateTransactionReverted
stTransactionTest/EmptyTransaction
stTransactionTest/OverflowGasRequire
stTransactionTest/RefundOverflow
stTransactionTest/RefundOverflow2
stTransactionTest/TransactionNonceCheck
stTransactionTest/TransactionNonceCheck2
stTransactionTest/TransactionToItselfNotEnoughFounds
stTransactionTest/UserTransactionGasLimitIsTooLowWhenZeroCost
```

If the state tests should not have invalid transactions, should all of those be deleted as well?

winsvega:
No. Thanks for the list. 
And yes you could open a PR that removes it. 
I will reimplement those cases in a form of correct blockchain test.

germsvel:
[...]

Issue #203, opened 2017-06-30
General State Tests contain transactions with invalid encoding


fjl:
The transaction value field should use RPC quantity encoding.

Specifically, 

* `0x` should not be used to encode numeric values. Please use `0x0` to encode tx value zero.
* For non-zero value, the field should not contain leading zero digits after `0x`.

Other fields should use RPC data encoding. Hex data should have `0x` prefix, it's missing for transaction input data in `stRevertTest/RevertDepthCreateAddressCollision.json` and a few other files.

winsvega:
We must come to a standart describing  hashes, hex encodings of bytes and integers in the .json tests. 
as well as the length and leading zeroes. (0x0 vs 0x00 and   0x00000001   values)

winsvega:
@pirapira  what is the current state of .json encoding.
I was trying to keep an EIP but we end up in a state we are currently are. 

So it looks like this 
```
every field value is starting with `0x` prefix except information comment fields.
if address is a key in json it is also prefixed with `0x`
if address is in transaction `to`  it is also prefixed with `0x`
"extraData" : ""   code and data fields might be empty
zero value is  "0x00"
```

pirapira:
I think we can make @jwasinger 's syntax checker stricter and stricter.  Then we'll hear opinions.

winsvega:
Current state: 

All fields in tests are 0x prefixed hex. (with exception of blocknumber field in bc tests and srcHash in _info section)
code/data/input/extraData fields are "0x" if empty, 0xbytes if not. 
0 values are 0x00
no leading 00 seen.

PR #507, opened 2018-09-13
Remove TransactionNonceCheck from GeneralStateTests

germsvel:
Resolves https://github.com/ethereum/tests/issues/504

What changed?
=============

Removes `TransactionNonceCheck` from `GeneralStateTests/stTransactionTest`

Why?
====

The `TransactionNonceCheck` test never passes the intrinsic validity required of a transaction outlined at the beginning of Section 6 of the YP, since the sender's nonce and the transaction's nonce are different. Because of that, this transaction should not execute at all, and for that reason, it does not make for a good state test. State tests should not have tests with invalid transactions


holiman:
>  State tests should not have tests with invalid transactions

I whole-heartedly disagree. Where does that come from?

As opposed to blocktests, which are _incapable_ of proving that a transaction is rejected, a statetest is a perfect format to ensure that a certain transaction is properly rejected.

germsvel:
@holiman that's a great point. I was going off a conversation that I had on gitter, so I could be totally wrong. I'm not sure if I can link to the gitter conversation in https://gitter.im/ethereum/tests, but I'll just copy that part of the messages,

> German Velasco @germsvel Sep 10 11:27
I have another question for a different test. Does anybody know why https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stTransactionTest/TransactionNonceCheckFiller.json creates the beneficiary's account before EIP 158? The pre section does not have it, but the post section does. It seems to me that since the account nonce is 10 and the transaction nonce is 0, the transaction never gets executed because it fails step (3) of the initial tests of intrinsic validity.

> Eth-Gitter-Bridge @Eth-Gitter-Bridge Sep 10 11:33
[W Dimitry] Actually wrong nonce transaction is smth that we could check via rpc or forged block tests. **The state tests should not have invalid transactions**
[W Dimitry] How to get a list of transactions with gasprice = 0 in the main chain that has been executed?

> German Velasco @germsvel Sep 10 11:38
I don't know how to get that list of transactions
Are you saying https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stTransactionTest/TransactionNonceCheckFiller.json should not be something we test in state tests? Should I open an issue regarding that?

> Eth-Gitter-Bridge @Eth-Gitter-Bridge Sep 10 11:50
[W Dimitry] Already have an issue about that somewhere

> German Velasco @germsvel Sep 10 11:53
Oh no! I just opened an issue ethereum/tests#504. I'll close it , saying its a duplicate. Do you know which issue it was so I can link to it?

> Eth-Gitter-Bridge @Eth-Gitter-Bridge Sep 10 11:55
[W Dimitry] ethereum/tests#203
[W Dimitry] Yes. Perhaps not exactly like this. You could open a new one
[W Dimitry] But I think such tests should be deleted
[W Dimitry] The transaction is not on real env (parsed by test tool) the blocks frim such tests are just empty blocks

winsvega:
when testing a wrong transaction in state tests it is a test parser tool that check a transaction.
when testing a wrong transaction in bc tests it is most certainly a block validation function that should detect a wrong nonce transaction

winsvega:
The schema check is red. Looks like you forgot to remove some file. 
But we have to delay this PR. We discussed with holiman on skype;


> Because we dont have spare time to move those tests into blockchain and you say its ok to leave incorrect transaction check to state tests parsers I will have to leave those tests. 
> 
> But I say I would prefere to remove all incorrect transaction tests from state tests and make blockchain tests out of them. 
> 
> From the other side if you say all clients has to implement state tests parsers (which I think should not be the case. clientss should implement RPC test methods) if clients implement test RPC methods retesteth will be able to run StateTests/BlockchainnTests + future tests without need for a client to change the parser

germsvel:
I'm closing this for now. I think after understanding https://github.com/ethereum/tests/issues/513, this PR seems like the wrong approach.

winsvega:
@holiman  suggested to keep such option (invalid transaction) in the StateTests. 
we decided to keep it.

germsvel:
Yeah, I think keeping them makes a lot of sense. 

The original impetus for me opening up this PR (and the associated issue #504) was that I couldn't understand how some state tests were expected to pass when they failed intrinsic validity before the transaction was even ran, while still creating the coinbase account (and thus having a changed state root). 

After some research, I discovered that the miner reward and possible clean up of an empty coinbase account need to be simulated in state tests. That is not documented anywhere in the docs, so I opened up https://github.com/ethereum/tests/issues/513.

But I agree that we should keep invalid transaction tests.

PR #522, opened 2018-10-11
remove artifact tests

winsvega:


holgerd77:
These tests still exist as state tests: https://github.com/ethereum/tests/tree/develop/GeneralStateTests/stArgsZeroOneBalance

Are these legacy there as well?

winsvega:
by legacy I mean old tests that were filled but not dont have filler file.

PR #527, opened 2018-10-15
Setup Improvements

ansermino:
Added a `requirements.txt` and instructions for setup to `README.md`. Instructions are based on EVMLab's (https://github.com/ethereum/evmlab/blob/master/README.md#installation). This will hopefully make it easier for newcomers to work with and contribute to this repo.

Currently the known requirements are:
```
pyaml
jsonschema
pysha3
```

Setup instructions:

(Requires `virtualenv`)
```
#> virtualenv -p python3 .env3
#> . .env3/bin/activate
#> python3 -m pip install -r requirements.txt
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,4 @@ docs/_build/
 .vagrant/
 *.pem
 src/GenStateTestAsBcTemp/
-.env/
+.env3/


README.md
```diff
@@ -40,6 +40,15 @@ Arrays are allowed, but don't use them for sets of properties - only use them fo
 Test Set Sanitation
 -------------------
 
+### Setup
+
+(Requires `virtualenv`)
+```
+#> virtualenv -p python3 .env3
+#> . .env3/bin/activate
```
holgerd77, line 48: Can you add the ``.env`` folder you used here to the ``.gitignore`` file, so that this is not checked in accidentally if someone creates the Python environment?

requirements.txt
```diff
@@ -0,0 +1,3 @@
+pyaml
+pysha3
+jsonschema
```
holgerd77: Can you limit/narrow this to at least the major versions of the libraries, so ``pyaml>=17.0,<18.0``, ``pysha3>=1.0,<2.0`` and ``jsonschema>=2.6,<3.0``, otherwise chances are high that things will break on the next major update of a library.

Wonder if one should even stick to the minor versions, but maybe for these generic types of libraries that's ok (though my over-time experience is, that stuff like this breaks rather more often than expected, but choose as you think is appropriate).

README.md
winsvega: Setup of what ?

holgerd77: Setup for the test set sanitation, but that's already stated in the headline above, so there shouldn't be an additional need to expand the headline, should it?

winsvega: oh, I just noticed. 
So we have tests. we have a test tool. and we have a tool for that test tool to run and check the tests. ) 

would be nice to have a script then that will compile / get the latest version of testeth
and check all of it. but I think we should better discuss how to make such a tool which could be independent of aleth (which testeth depend). On devcon there will be a discussion about this.

ansermino: Would love to see this truely decoupled from aleth!

ansermino:
@holgerd77 Let me know if there is anything else :)

PR #535, opened 2018-10-24
sstore collision overwrite sstore

winsvega:
external call overwrite sstore


Retesteth/eth/eth.sh
```diff
@@ -0,0 +1,2 @@
+echo "Starting $1 $2"
+eth --test --db-path $1 --ipcpath $2 --log-verbosity 4
```
holgerd77, line 2: Just writing up a changelog for a new release (will make a draft first), is this for internal usage or is it meant to be documented?

What is this ``eth`` tool generally doing?

winsvega: Its a cpp ethereum client

PR #538, opened 2018-11-03
Tests where data is copied to a memory range longer than the data itself

ajsutton:
Adds tests to cover cases where the result of an EVM opcode is written to a specified memory range and the result is shorter than the specified range. This includes:

 * CALL to ecrecover precompile with a completely invalid key
 * CALL to ecrecover precompile with a sensible key but a signature can't be recovered
 * CALL to identity precompile
 * EXTCODECOPY where code being copied is shorter than the target memory range

These cover cases where Pantheon behaviour differed from other clients - initially because EXTCODECOPY did not zero out memory beyond the end of the data and then we fixed that which broke CALL which expects the output length to be treated as a maximum length so remaining memory is not zeroed out. The reference tests don't appear to have any coverage of this category of behaviour.


ajsutton:
Note that there are some other opcodes that may need tests like these too (at minimum CODECOPY).  Wanted to check this is a sensible approach to this kind of test before going too far.

holgerd77:
Hi @ajsutton, to follow up the mail exchange: great that you also want to help along the PR to improve documentation!

I have just submitted a PR with a transition of the creation docs to this repository and some initial updates https://github.com/ethereum/tests/pull/539, this is nevertheless just a starting point.

Feel free to write initial questions you have directly in the code of the PR, I will try to update the PR accordingly. We can nevertheless also do subsequent PRs once this initial PR is merged.

holgerd77:
Oh, tests are passing, great! 👍

ajsutton:
Yeah I misinterpreted the error message - it was actually telling me that I’d missed refilling one test so just had to do that and I also generated the blockchain variants based on the intructions in the PR you mentioned.

src/GeneralStateTestsFiller/stPreCompiledContracts2/CallEcrecoverInvalidSignatureFiller.json
```diff
@@ -0,0 +1,60 @@
+{
+    "CallEcrecoverInvalidSignature" : {
+        "env" : {
```
winsvega, line 6: could you please add 
```
"_info" : {
     "comment" : " Test Description "
}
```

section describing the logic behind LLL code.  please?
in other test fillers too.

ajsutton: I didn't know that was possible but definitely good to have.  I've added some, let me know if they explain things well enough.

winsvega:
ready to merge?

ajsutton:
Yep I’m happy if you are.

holgerd77:
We just did a first test run with your tests added and directly [stumbled over](https://github.com/ethereumjs/ethereumjs-vm/pull/387), thanks for catching this! 👍

ajsutton:
Well I'm glad it's not just me that got this wrong then. :)

Issue #541, opened 2018-11-05
Test case: EXTCODEHASH of nonexistent account is 0


jwasinger:
From https://eips.ethereum.org/EIPS/eip-1052

winsvega:
please mark it to this page 
https://docs.google.com/spreadsheets/d/1xat7UI8GtB4ZGVdlK5_XQSHJZaMThi4SrlcL8XMZb5Q/edit?pli=1#gid=1811198384

jwasinger:
Had originally opened this so that the test case could be bountied but @hugo-dc picked it up.

Issue #540, opened 2018-11-05
Test case: EXTCODEHASH of account without code


jwasinger:
From https://eips.ethereum.org/EIPS/eip-1052 :
`The EXTCODEHASH of the account without code is c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470 what is the keccack256 hash of empty data.`

winsvega:
please post this issues to: 
https://docs.google.com/spreadsheets/d/1xat7UI8GtB4ZGVdlK5_XQSHJZaMThi4SrlcL8XMZb5Q/edit?userstoinvite=hugo@ethereum.org&ts=5be5af56&actionButton=1#gid=1811198384

PR #542, opened 2018-11-08
new EXTCODEHASH tests

hugo-dc:
- [x] EXTCODEHASH to an account without code - https://github.com/ethereum/tests/issues/540
- [x] EXTCODEHASH to a non existing account - https://github.com/ethereum/tests/issues/541
- [x] EXTCODEHASH of contract with codesize equal to maxcodesize - 1 (24576 bytes)
- [x] EXTCODEHASH of precompiles accounts (1-8)


holgerd77:
👍 👍 👍

winsvega:
Dont cover all tests in one go. Make several small PRs

winsvega:
Which testeth commit you use? 
Filler hash is either outdated testeth issue or the way you construct yaml sources.

Make sure the fields are string and int accordingly: https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stArgsZeroOneBalance/eqNonConstFiller.yml

holgerd77:
Question for the docs: what is the current state of .yml vs .json? Are both formats supported equally? or is .yml to be preferred (and .json even to be removed in the future?)?

Then I would eventually update all the filler .json examples in the docs with the corresponding .yml variant.

winsvega:
both are supported. no preference. 
the issue is that testeth does not check yml semantics. so you could put string instead of int and vice verca and it will still be able to run it. while this kind of change is crucial to hash calculation by the python script.

holgerd77:
Ah, thanks!

hugo-dc:
> Which testeth commit you use?

I tried with testeth 1.4.0, then tried with testeth 1.5.0.dev2-76+commit.2fd57400, both give different hashes and both fail in travis ci

> Filler hash is either outdated testeth issue or the way you construct yaml sources.

I'll check the yaml files to see if there is something to fix

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashPrecompilesFiller.yml
```diff
@@ -0,0 +1,160 @@
+# EXTCODEHASH of precompiles addresses
+---
+extCodeHashPrecompiles:
+  env:
+    currentCoinbase: 2adc25665018aa1fe0e6bc666dac8fc2697ff9ba
+    currentDifficulty: '0x20000'
```
hugo-dc, line 6: The problem was I wasn't using quotes for this value

hugo-dc:
@winsvega @holgerd77 this is ready for review, I'll continue working on new tests in a different PR

holgerd77:
@hugo-dc Great, thanks so much for your work Hugo! 👍

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashPrecompilesFiller.yml
```diff
@@ -0,0 +1,160 @@
...
+    aaaaaaaa00000000000000000000000000000002:
+      balance: '1000000000000000000'
+      code: |
+        {
+          (MSTORE 0 (EXTCODESIZE 0x0000000000000000000000000000000000000002))
+          (RETURN 0 32)
+        }
+      nonce: '0'
+      storage: {}
+    # account with random bytecode with code size = MAXCODESIZE - 1 = 24575
```
winsvega: comment from the previous test file. no longer actual

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashPrecompilesFiller.yml
```diff
@@ -0,0 +1,160 @@
...
+        4: "0x01",
+        5: "0x01",
+        6: "0x01",
+        7: "0x01"
+        }
+    aaaaaaaa00000000000000000000000000000001:
+      balance: '1000000000000000000'
+      code: |
+        {
+          (MSTORE 0 (EXTCODEHASH 0x0000000000000000000000000000000000000001))
```
winsvega: extcodehash / extcodsize  are mixed here (in this test) inconsistent. 
you could also pass the contract that you want to call as DATA for the call.

```
(MSTORE 0 0x1122...32) (CALL 50 account 0 0 32 0 0)
```

and load that data in account with `(CALLDATALOAD 0)`
so you would have one contract as function that execute and store return values of extcodesize and extcodehash  of an input argument. then you just use different data in general transaction to call that contract/function with different data.  in this case the data (argument for the function) is precompiled addresses.

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashPrecompilesFiller.yml
winsvega: please use the latest master branch testeth as it must work anyway.

winsvega:
related: https://github.com/ethereum/tests/issues/541
related: https://github.com/ethereum/tests/issues/540
require testeth update if that wasnt the case. 

(I mean if those conditions are not true with the latest testeth then cpp aleth must be fixed and testeth must be rebuilt)

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashPrecompilesFiller.yml
hugo-dc: Thanks for reviewing, I have updated the test case according to your comments. I'm sending the precompiles addresses as data. I updated the test cases using latest aleth/testeth master (i had to change it to accept the stExtCodeHash directory), I don't know if that is what you meant or if I should use https://github.com/ethereum/testeth instead.

GeneralStateTests/stExtCodeHash/extCodeHashAccountWithoutCode.json
```diff
@@ -0,0 +1,84 @@
+{
+    "extCodeHashAccountWithoutCode" : {
+        "_info" : {
+            "comment" : "",
+            "filledwith" : "testeth 1.5.0-alpha.6-1+commit.19ad7d95.dirty",
```
winsvega: .dirty means you have uncommited changes to testeth. thats why. 
ideally a PR to aleth repo is open. testeth modified with new test suites and commited. (so no dirty flag) 
then aleth run it on a specific tests branch with new tests. (also set in a PR to aleth) 

*note you dont need to do that logic. as it is already done on aleth create2 branch. what I meant was 
if testeth from create2 branch or the latest version of testeth that you use produce incorrect test results (from one expected by EIPs) then testeth must be updated (and aleth as well) 

https://github.com/ethereum/tests/issues/540
https://github.com/ethereum/tests/issues/541

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashPrecompilesFiller.yml
winsvega: oh. this comment was about failing test hash issues. but seems like hash reports from python script are due to the .yml files format. (quotes)

winsvega:
looks good. 
the .yml format though keep failing by python script check. (which is done separate from testeth) thats why you could not check it while filling.
and ideally would be to get rid of `.dirty` flag in testeth commit version.

hugo-dc:
@winsvega i refilled tests using the `create2` branch of aleth, now all validation tests are passing.

winsvega:
looks good. please run --filltests --fillchain and commit the changes

hugo-dc:
filled blockchain files

holgerd77:
Hi @hugo-dc, congrats, super cool, I actually wasn't aware that you have got such deep knowledge on test creation! 😄

One thing: these additional "info" fields in the fillers are actually super-important for using the tests on the client side, people are otherwise spending up to hours figuring out what a test case really is about (I know there is some hint through the test case naming, but this is normally not detailed enough).

I know this is a bit annoying but could you do this extra round and add info-comments to the ``yml`` files? That would be great! There also (probably?) shouldn't be the need to re-fill or something, this will organically happen somewhat later, so just a PR against the initial yml fillers.

[Here](https://github.com/ethereum/tests/blob/428842e9731b7065366dcd7829bbfc0f9e1135fa/src/GeneralStateTestsFiller/stSystemOperationsTest/doubleSelfdestructTestFiller.yml) is an example of a comment integrated in a yml filler (just for the syntax, write in whatever you like).

holgerd77:
Additional note: have just seen that you have added explanations as comments in the beginning of the yml files. This would nevertheless be better to be placed in the info comment fields, since this is then also transported to the filled tests and can there further be used (e.g. also programatically by printing out when running the tests).

winsvega:
To understand what test does you better study filler.json or filler.yml which has comments and source code of the bytecode

hugo-dc:
@holgerd77 thanks for your comments!, definitely it would be better to include detailed comments to the `_info` field, I will consider it for the next test cases, I can add it to these tests as well, thanks for showing an example.

PR #543, opened 2018-11-10
EXTCODEHASH: test for deleted account

hugo-dc:
- [ ] Account A is already in the state, get EXTCODEHASH of account A, transaction to a contract deletes Account A, then execute EXTCODEHASH again.

This test case first gets the EXTCODEHASH, then deletes the account, then try to get the EXTCODEHASH again and still get the same code hash, am i missing something?

Error:
```
error: in "GeneralStateTests/stExtCodeHash": extCodeHashDeletedAccount Check State: 095e7baea6a6c7c4c2dfeb977efac326af552d87: incorrect storage [0x01] = 0x9dff858fbcae72e5644a465fa893f46deb3e03b9cd5a8b531109cd577f7fb293, expected [0x01] = 0x
```

### DIFF
--- a/GeneralStateTests/stExtCodeHash/extCodeHashDeletedAccount1.json
+++ b/GeneralStateTests/stExtCodeHash/extCodeHashDeletedAccount1.json
@@ -5,7 +5,7 @@
             "filledwith" : "testeth 1.5.0-alpha.6-49+commit.4897b8bb",
             "lllcversion" : "Version: 0.5.0-develop.2018.11.8+commit.20481055.Linux.g++",
             "source" : "src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashDeletedAccount1Filler.yml",
-            "sourceHash" : "e78f6efbbaa04c5bd3a87d63d771252beee6dc83d506b7db562837b594ee9346"
+            "sourceHash" : "ce870dddcddc514d6bf9d0232cd1926e4ec1e875a330b80d8f7cc2aabbd6f0d0"
         },
         "env" : {
             "currentCoinbase" : "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
@@ -18,7 +18,7 @@
         "post" : {
             "Constantinople" : [
                 {
-                    "hash" : "0x26a950b1fd969e55a6e213283d3e216c8b9198af299682a089858849d3d55715",
+                    "hash" : "0xa3351a168ff2dcecf68704dcb541a7ee5597d205453834d265dfc7300e38e563",
                     "indexes" : {
                         "data" : 0,
                         "gas" : 0,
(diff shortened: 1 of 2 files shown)


winsvega:
if you do it in one go looks like it should still return the code hash as suicide queries are perfomed at the end of transaction during state finalization stage. leave this case to expect the same hash after suicide.

I think possible test cases here could be: 
1) account A already in the state with the code
call ->  sstore accountA hash to 1 -> run selfdesturct of A -> sstore accountA hash to 2  (expext section check 1 and 2 storage values) 

2) account A already in the state with the code
call ->  sstore accountA hash to 1. exit the call.  another call (same transaction) -> run selfdesturct of A .  exit call.   another call (same transaction) -> sstore accountA hash to 2  (expext section check 1 and 2 storage values) 

*3) *4)  are same but account A is created on the fly and not set before the first call. 

5) a blockchain test with block having 2/3 transactions. account a is in the state with the code before the first transaction.
1st transaction (sstore 1 codehash(a)) 
2nd transaction suicide a  (could be merged with 1st transaction) 
3d transaction (sstore 2 codehash(a)) 
check by expect section sstore value of 1 and 2.

6) same as above but account is being created by the first transaction.

holgerd77:
Same here with the info comments (probably also the yml comment on the pre field).

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashDeletedAccountFiller.yml
```diff
@@ -0,0 +1,77 @@
...
+        gas:  !!int -1
+        value: !!int -1
+      network:
+        - '>Byzantium'
+      result:
+        095e7baea6a6c7c4c2dfeb977efac326af552d87:
+          balance: '1000000000000000001'
+          storage: {
+            0: '0x9dff858fbcae72e5644a465fa893f46deb3e03b9cd5a8b531109cd577f7fb293',
+            1: '0'
```
winsvega: so we concluded that here should be expected the same hash. must be no error.

hugo-dc:
thanks @winsvega !!, I have updated the test case and added new ones.

winsvega:
5) and 6) I covered with blockchain tests.

I would say the way Yaml allows you to write contract source in LLL is amazing. 
could you please also add in the same scenarious a check for
` extcodecopy and extcodesize return values. `
same as for extcodehash.

hugo-dc:
@winsvega yes, I really liked Yaml to write tests it makes lll/ewasm code more readable.
I have updated the tests adding checks for extcodecopy and extcodesize, however I have a problem with one test case:

When I try to store the external code into key [[6]] the whole test fails, seems like the transaction gets reverted but still don't know why: https://github.com/ethereum/tests/pull/543/files#diff-add14fc5fc096a1cebbcf46e40381f77R59 , I commented out that line and everything worked. Do you have any idea what the problem may be?

ajsutton:
Any chance the transaction is running out of gas? Storage tends to be pretty expensive.

winsvega:
Try filling that test with --vmtrace to see what s going on. You will see OOG exception notes if it fails on last instruction with oog.

hugo-dc:
@ajsutton @winsvega you're right, it was running out of gas. I fixed the test case.

PR #544, opened 2018-11-12
Add EXTCODEHASH tests

ajsutton:
- [x] a contract execute EXTCODEHASH of itself 
- [x] a contract execute EXTCODEHASH with address being a dynamic argument 
- [x] a contract execute EXTCODEHASH of itself during the init code 

I figured small batches was a good balance between review size and amount of time spent wrangle git branches but I'm happy to create separate PRs for each of these if that's easier.


src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashDynamicArgumentFiller.json
```diff
@@ -0,0 +1,93 @@
...
+                "balance": "1000000000000000000",
+                "code": "0x",
+                "nonce": "0",
+                "storage": {
+                }
+            }
+        },
+        "transaction": {
+            "data": [
+                "0x000000000000000000000000deadbeef00000000000000000000000000000001"
```
winsvega: this could be extended (just by adding more data elements) to dynamic argument to be 
1) non existent account
2) precompiled account
3) empty account / no code

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashDynamicArgumentFiller.json
```diff
@@ -0,0 +1,93 @@
...
+                }
+            },
+            {
+                "indexes": {
+                    "data": -1,
+                    "gas": -1,
+                    "value": -1
+                },
+                "network": [
+                    "<=Byzantium"
```
winsvega: if this checks that extcodehash is failing it might be just enough to have one test checking that extcodehash failing on all forks <= Byzantium. and in this and other tests it is not required to check its failing on <=Byzantium. 
cause otherwise it will produce many tests that all do the same thing - checking that extcodehash is failing on <= Byzantium.

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashSelfFiller.json
```diff
@@ -0,0 +1,86 @@
...
+        }
+      },
+      {
+        "indexes": {
+          "data": -1,
+          "gas": -1,
+          "value": -1
+        },
+        "network": [
+          "<=Byzantium"
```
winsvega: same tests overload (see above comments)

winsvega:
dont forget to run --filltests --fillchain once done. to enable this tests for hive.

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashDynamicArgumentFiller.json
ajsutton: True.  I've added these cases - actually 2 cases for precompiles, one where it exists as an account and one where it doesn't.  I believe we have PRs to add tests specifically for these cases as well, do we want both dynamic and static versions or is the dynamic case enough?

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashDynamicArgumentFiller.json
ajsutton: Agreed, I copied this from another test.  I suspect the best approach is to have a single test that is specifically checking EXTCODEHASH is unsupported <= Byzantium. The purpose is fairly ambiguous with it mixed in with other tests.

For now I've removed the <=Byzantium cases from the tests I'm adding and also change >Byzantium to >=Constantinople to be clearer that these are added in Constantinople (my little brain struggles to remember the ordering of hard forks).

ajsutton:
Added the blockchain variants.

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashDynamicArgumentFiller.json
winsvega: I think dynamic case is enough. also try not to use <=Byzantium expect section as it multiplies the amount of tests. and extcodehash not even supported on <=Byzantium.

PR #545, opened 2018-11-12
Moresstore

winsvega:
Sstore combination calls tests
(See the first commit for review)

initial storage {0, 1, 2}  -> changed by {calltype} to {0,1,2} then {calltype} and do {nothing, oog, revert} then change the storage again by {calltype} to {0,1,2} then by {calltype} do {nothing, oog, revert}


holgerd77:
Actually I am still not getting when there is the need to recreate all the files like in here and in what cases it is enough to just re-create the specific files like in the PR from @hugo-dc https://github.com/ethereum/tests/pull/542. If you could give a short explanation I would be grateful!

winsvega:
I dont understand your question.
Test is only updated if it is filled from the latest filler. to check that we use the filler hash check. 

test filler  ->  actual test (filled from test filler with hash: 0x2223344) 

so if you change test filler. you will have to refill the test, cause it will change the hash. 
then blockchain tests: 

state test filler ->  blockchain test filler -> actual blockchain test. 

this is a workaround. we convert all existing state tests into blockchain tests.
so if you change state test filler. you might miss the update of its version as a blockchain test
cause it is actually two separate tests.

holgerd77:
@winsvega Thanks!

PR #549, opened 2018-11-18
Add test for EXTCODEHASH of account created and deleted in same transaction

ajsutton:
Adds tests for EXTCODEHASH and EXTCODESIZE on an account that is created and then deleted in the same transaction.  Note that the EXTCODEHASH is the same before and after deletion because SELFDESTRUCT doesn't take effect until after the end of the transaction.

This may make #548 redundant as it is grabbing the EXTCODEHASH between creation and deletion so is covering the newly created account case as well.  Not sure if we to just have this version or keep both (and potentially simplify this version a little).

There's also some overlap with #543 which is testing deletion of an existing account. There are a number of other tests identified there but I don't think they apply in this case because we're specifically interested in an account that was created in the same transaction it was destroyed - once the transaction exits it's just an existing account. It may be worth a blockchain test where transaction 1 creates and deletes the account, then transaction 2 gets the EXTCODEHASH of it but it seems a stretch.

### DIFF
--- a/src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashCreatedAndDeletedAccountCallFiller.json
+++ b/src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashCreatedAndDeletedAccountCallFiller.json
@@ -29,6 +29,9 @@
               "0x02": "0x396daaf3a7871f857cc9b5878c6bb66c394a6d109706095617db9f4e4c22edcd",
               "0x03": "0x03"
             }
+          },
+          "7be7677a20a298ff0676ab31f91a0a923f8d9471": {
+            "shouldnotexist": "1"
           }
         }
       }
--- a/BlockchainTests/GeneralStateTests/stExtCodeHash/extCodeHashCreatedAndDeletedAccountCall_d0g0v0.json
+++ b/BlockchainTests/GeneralStateTests/stExtCodeHash/extCodeHashCreatedAndDeletedAccountCall_d0g0v0.json
@@ -5,7 +5,7 @@
             "filledwith" : "testeth 1.5.0-alpha.6-1+commit.19ad7d95.dirty",
             "lllcversion" : "Version: 0.5.0-develop.2018.9.29+commit.c4b5101b.Darwin.appleclang",
             "source" : "src/BlockchainTestsFiller/GeneralStateTests/stExtCodeHash/extCodeHashCreatedAndDeletedAccountCall_d0g0v0Filler.json",
-            "sourceHash" : "471231989d740e9224dc7fdcb3a416938c9c918a711ed2676dbd8b028c9fe3ff"
+            "sourceHash" : "43dc93d23b9377a43f4684e7c3abfb612edf3b876cde03e20b1706e18a30f023"
         },
         "blocks" : [
             {
--- a/GeneralStateTests/stExtCodeHash/extCodeHashCreatedAndDeletedAccountCall.json
+++ b/GeneralStateTests/stExtCodeHash/extCodeHashCreatedAndDeletedAccountCall.json
@@ -5,7 +5,7 @@
             "filledwith" : "testeth 1.5.0-alpha.6-1+commit.19ad7d95.dirty",
             "lllcversion" : "Version: 0.5.0-develop.2018.9.29+commit.c4b5101b.Darwin.appleclang",
             "source" : "src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashCreatedAndDeletedAccountCallFiller.json",
-            "sourceHash" : "929618720bfddefd40804dbb348ac7180e9fbe484a9e0fbaf7cc030f66ece6f5"
+            "sourceHash" : "a0cc361e8f2e2d0400df6852c05e605fa8028a1d51a207e78819b78fca658ae8"
         },
         "env" : {
             "currentCoinbase" : "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
--- a/src/BlockchainTestsFiller/GeneralStateTests/stExtCodeHash/extCodeHashCreatedAndDeletedAccountCall_d0g0v0Filler.json
+++ b/src/BlockchainTestsFiller/GeneralStateTests/stExtCodeHash/extCodeHashCreatedAndDeletedAccountCall_d0g0v0Filler.json
@@ -32,6 +32,8 @@
             {
                 "network" : "Constantinople",
                 "result" : {
+                    "0x7be7677a20a298ff0676ab31f91a0a923f8d9471" : {
+                    },
                     "0xdeadbeef00000000000000000000000000000000" : {
                         "storage" : {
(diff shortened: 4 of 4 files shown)


src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashCreatedAndDeletedAccountFiller.json
```diff
@@ -0,0 +1,68 @@
...
+      "a94f5374fce5edbc8e2a8697c15331677e6ebf0b": {
+        "balance": "1000000000000000000",
+        "code": "0x",
+        "nonce": "0",
+        "storage": {
+        }
+      },
+      "deadbeef00000000000000000000000000000000": {
+        "balance": "1000000000000000000",
+        "code": "{ (MSTORE 0 (CREATE2 0 128 (lll { (RETURN 0 (lll { (SELFDESTRUCT 0x0) } 0)) } 128) 0x10)) [[0]] (EXTCODEHASH (MLOAD 0)) [[1]] (EXTCODESIZE (MLOAD 0)) (CALL 0x10000 (MLOAD 0) 0 0 0 0 0) [[2]] (EXTCODEHASH (MLOAD 0)) [[3]] (EXTCODESIZE (MLOAD 0)) (STOP) }",
```
winsvega, line 46: it make sense to check results on different type of calls here
create account -> check extcodehash -> {call, callcode, delegatecall, staticcall} selfdestruct -> put extcodehash to storage.

winsvega:
also one more interesting case: 

transaction ->  call to account ->  create new contract.  check its extcode hash. run selfdestruct. exit call | in a upper call check extcodehash again

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashCreatedAndDeletedAccountFiller.json
ajsutton: I've added a static call variant of the test.

delegatecall or callcode don't change the recipient address so when the new account executes selfdestruct, it destroys the calling contract which is where we wanted to capture the extcodehash results.  We could do a CALL -> DELEGATECALL -> SELFDESTRUCT type of sequence but does that defeat the point?

ajsutton:
I've added what I think you meant for "transaction -> call to account -> create new contract. check its extcode hash. run selfdestruct. exit call | in a upper call check extcodehash again" as extCodeHashCreatedAndDeletedAccountRecheckInOuterCall but not entirely sure I've understood what you were suggesting.

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashCreatedAndDeletedAccountFiller.json
winsvega: you right. the idea was to check what extcodehash(A) returns if prior to this extcodehash instruction the account A was selfdestructed by delegatecall / callcode 

1) transaction to B
2) B call to A
3) A delegatecall/callcode to C  (C has selfdestruct)
4) A selfdestructed. returned to B. now we could check extcodehash of A (in account B code)

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashCreatedAndDeletedAccountRecheckInOuterCallFiller.json
```diff
@@ -0,0 +1,81 @@
...
+        "result": {
+          "deadbeef00000000000000000000000000000000": {
+            "storage": {
+              "0x00": "0x396daaf3a7871f857cc9b5878c6bb66c394a6d109706095617db9f4e4c22edcd",
+              "0x01": "0x03",
+              "0x02": "0x396daaf3a7871f857cc9b5878c6bb66c394a6d109706095617db9f4e4c22edcd",
+              "0x03": "0x03"
+            }
+          },
+          "deadbeef00000000000000000000000000000001": {
```
winsvega, line 33: I think it is better to add account address (MLOAD 0) here 
and check that it does not exit. this way we make sure that call to that account was successfull. 
```
"accountaddress": {
"shouldnotexist" : "1"  
}
```

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashCreatedAndDeletedAccountStaticCallFiller.json
```diff
@@ -0,0 +1,68 @@
...
+        "indexes": {
+          "data": -1,
+          "gas": -1,
+          "value": -1
+        },
+        "network": [
+          ">=Constantinople"
+        ],
+        "result": {
+          "deadbeef00000000000000000000000000000000": {
```
winsvega, line 25: so if in previous test we were checking that address "shouldnotexist": "1"
here in this case we need to check that it does exist. cause staticcall would fail to perform selfdestruct

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashCreatedAndDeletedAccountFiller.json
winsvega: better be done in a separate PR

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashCreatedAndDeletedAccountCallFiller.json
```diff
@@ -0,0 +1,68 @@
...
+        "indexes": {
+          "data": -1,
+          "gas": -1,
+          "value": -1
+        },
+        "network": [
+          ">=Constantinople"
+        ],
+        "result": {
+          "deadbeef00000000000000000000000000000000": {
```
winsvega, line 25: check that newly created account shouldnotexist: 1

winsvega:
extCodeHashCreatedAndDeletedAccountRecheckInOuterCall 
looks legit. 
the idea is to create contract in a sub call. and then remove it in a same subcall. but after this check in the upper call its extcodehash.

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashCreatedAndDeletedAccountRecheckInOuterCallFiller.json
ajsutton: Added.

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashCreatedAndDeletedAccountStaticCallFiller.json
ajsutton: Added.

src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashCreatedAndDeletedAccountCallFiller.json
ajsutton: Added.

PR #550, opened 2018-11-18
Add test for EXTCODEHASH before and after changing account balance, nonce and storage

ajsutton:
contract A exists. delete contract A. create a new version of contract A. check EXTCODEHASH of A

Since none of those changes affect the EXTCODEHASH, the result of each call is the same.

### DIFF
--- a/src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashChangedAccountFiller.json
+++ b/src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashChangedAccountFiller.json
@@ -1,7 +1,7 @@
 {
   "extCodeHashChangedAccount": {
     "_info": {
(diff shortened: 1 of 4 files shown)


src/GeneralStateTestsFiller/stExtCodeHash/extCodeHashChangedAccountFiller.json
```diff
@@ -0,0 +1,89 @@
+{
+  "extCodeHashChangedAccount": {
+    "_info": {
+      "comment": "EXTCODEHASH/EXTCODESIZE of an account created then deleted in same transaction"
```
winsvega: the account changes its nonce and storage.

ajsutton: I knew the copy/pasting would bite me eventually.  Fixed.

PR #552, opened 2018-11-21
delete account and check its extcodehash in blockchain

winsvega:


holgerd77:
Really impressive all the cases you guys come up with!

PR #554, opened 2018-11-27
add new test case not covered by RLP tests

ayrat555:
We discovered that our implementation of RLP encoding decodes some RLP binaries incorrectly after recent PR (https://github.com/mana-ethereum/ex_rlp/pull/17) but it passes all RLP test cases. 

This PR adds a new test case which is not covered by existing RLP tests

PR #557, opened 2018-11-28
prefix hex values with 0x in RLP tests

ayrat555:
We need to prefix hex values with `0x` so hex values can be in the input field. See https://github.com/ethereum/tests/pull/554


holgerd77:
Just some note: this definitely has to be stated in release notes, since this likely requires some changes on libraries using the tests and comparing results.

Issue #556, opened 2018-11-28
Remove PoW tests


holgerd77:
According to a [comment](https://gitter.im/ethereum/tests?at=5bfd8634cfa682348d931f6c) from @chfast on Gitter PoW tests have no great value and should be removed.

To make this super-transparent, I would suggest the following workflow (which can then be also applied to similar cases on a way to clean up the repository):

- We mark these tests as ``DEPRECATED`` in the next release
- Then this gets some attention and - if people still rely on these tests - they can object and drop a note
- If there is no objection we can (relatively) safely remove the tests here

Does this make sense?

chfast:
Yes.

holgerd77:
Cool. 😄

lithp:
I did a review of the major clients. As far as I can tell none of Geth, Aleth, Parity, Trinity, Harmony, or Mana run the PoW Tests. I might have missed something though, and there might be other clients which do use them. 

I also wasn't able to find any usage of the ABITests or GenesisTests. It might be worth deprecating those at the same time?

lithp:
Pulling in [a comment from gitter](https://gitter.im/ethereum/tests?at=5c07d259f4880a60a22e3f4d):

> @ germsvel we run ethash PoW tests in Nethermind
> @ germsvel they were very helpful when we were implementing PoW

holgerd77:
Ah, a bit embarrassing, but we are using these tests actually as well ourselves [here](https://github.com/ethereumjs/ethashjs/blob/master/tests/ethash.js#L6) in the ``ethashjs`` library, just looked it up, lol. 😛

Ok, then let's leave the tests.

I'll take this experience as a sign that is is indeed super necessary to go through this deprecation process and not just delete stuff, various potential users out there who might have embedded things (and if it's only for whatever legacy reasons).

jleni:
Trying to catch up and understand the long term plan.
Does it make sense that I look at adding Clique (PoA) as another engine and generate tests for this?
How are most implementations testing Clique if it is not through the json files provided here?

chfast:
Not relevant, @jleni,
These tests only check PoW mining and verification which PoA obviously does not have.

winsvega:
sub protocol implementations tests could potentially be added to rpc debug methods. so this tests could be run with a universal test tool: https://github.com/ethereum/retesteth/wiki/

PR #559, opened 2018-11-29
remove POW tests

winsvega:
https://github.com/ethereum/tests/issues/556


holgerd77:
Can we first do an in-between release and mark these tests as ``DEPRECATED``, before merging here (as proposed in #556)? Or are you super-sure that no one is using these tests and they can be deleted safely?

lithp:
The comments in #556 have examples of active users of these tests, closing this PR as the PoW tests seem to be valuable.

PR #566, opened 2018-12-10
extcodehash of overwritten account

winsvega:


winsvega:
note:  yaml  vs json storage cell number auto ordering does not work when computing the hash.  
`0x01 0x02 0x12 0x03  vs  0x01 0x02 0x03 0x12`

Issue #531, opened 2018-10-19
[Meta] Test releases for the tests repository


holgerd77:
Hi @winsvega, have you ever thought about making releases on the tests repository, think these could already solve a lot of problems. This can be low-level, e.g. just doing tags on the GitHub [release](https://github.com/ethereum/tests/releases) page and adding some CHANGELOG entry, we are doing this over on our proxy [ethereumjs-testing](https://github.com/ethereumjs/ethereumjs-testing) library and have made good experiences with it.

This would have the effect that people stay a bit on board about what's going on (test format has changed, Constantinople tests for EIP xxx are ready,...), significantly reduces the risk of people using inconsistent states of the repository and would also help a lot to talk about problems ("What release are you using?" instead of "I am referring to the library in state with latest commit xxx").

What do you think? I could also take on this job this I am doing this anyhow for the JS library mentioned above, but I am also happy with you taking on it if you want to have control over the process.

holgerd77:
Hi everyone,
here is a proposal for release notes, I would leave this open for discussion and comments for 24 hours and then publish a tagged release. I think this will generally be a good way ensure to not do unintended releases and give @winsvega and others working on the tests the chance for some feedback ("please wait, develop branch not stable", "please wait, would like to merge PR xx",...), so I will stick to that concept and publish draft notes maybe here or within another dedicated issue.

After some more thoughts I also think that I would like to weaken the initially proposed strict bi-weekly release schedule a bit and instead still do regular releases but within a period of 1-4 weeks. In some more dense periods (like atm) there might be the need for more regular releases, e.g. if the last EIP tests are merged, in other periods with low work activity doing a release after 4 weeks might be enough.

Due to the nature of this library with more constant updates here and there on tests and less of really feature-centric events/large changes I would generally stick to a regular-release-concept.

Here is a draft for some first release notes:

---

**Tag:** v6.0.0-beta.1

**Title:**
v6.0.0-beta.1 Release CW 43

**Description:**

This is the first of a new regular series of test releases, which can be used to reference a static snapshot of the tests within your library. For a starter release notes will give an overall summary of the status of consensus ``Constantinople`` tests, a summary of recent noteworthy changes and an overview on release versioning. Subsequent releases will then provide a ``CHANGELOG`` towards the release before.

### Constantinople Test Summary

[...]

PR #571, opened 2018-12-17
Add a CHANGELOG.MD

lithp:
Backfilled with the last two releases.

Part of #531, opened by @holgerd77

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ No changes yet :)
 
 ### Potentially breaking changes
 
-- #547 requires your integration to understand the network `ByzantiumToConstantinopleAt5`, which is analogous to existing pseudo networks such as `FrontierToHomesteadAt5` and `EIP158ToByzantiumAt5`. It also changed the format of some blockchain tests, [keys of "0x" are now "0x00"](https://github.com/ethereum/tests/pull/547/files#diff-d31f2819c1febd4c8d673a15a8375af5)
+- #547 requires your integration to understand the network `ByzantiumToConstantinopleAt5`, which is analogous to existing pseudo networks such as `FrontierToHomesteadAt5` and `EIP158ToByzantiumAt5`.
 
 - #557 changes the format of the rlp tests, byte strings are now prefixed with "0x"
 


CHANGELOG.md
```diff
@@ -0,0 +1,76 @@
+## Unreleased
+
+No changes yet :)
```
winsvega, line 3: What this line stands for?

CHANGELOG.md
```diff
@@ -0,0 +1,76 @@
...
+
+## v6.0.0-beta.3
+
+### Potentially breaking changes
+
+- #547 requires your integration to understand the network `ByzantiumToConstantinopleAt5`, which is analogous to existing pseudo networks such as `FrontierToHomesteadAt5` and `EIP158ToByzantiumAt5`. It also changed the format of some blockchain tests, [keys of "0x" are now "0x00"](https://github.com/ethereum/tests/pull/547/files#diff-d31f2819c1febd4c8d673a15a8375af5)
```
winsvega: 0x to 0x00 was done to the filler. It doesn't affect anything. Its just that 0x  confuses me. No need to mention this change as it might confuse. Because real 0x requirement done to RLP tests

CHANGELOG.md
lithp: Eventually this section will have the changes which are in develop but not part of any release. For now it's empty, this is just a placeholder so the empty section doesn't look weird, it will be removed once a PR is merged.

CHANGELOG.md
lithp: Good point, I'll remove it.

lithp: Fixed

lithp:
Friendly ping :)

holgerd77:
@winsvega What is actually your policy for dealing with branches merged? I'm always having a first-time-impulse to press the "Delete branch" button above since we are doing this on EthereumJS and I find this a good practice to clean up the old branches.

And then I hesitate on second thought and don't do because I wonder if you handle this otherwise?

winsvega:
I delete old branches from time to time once there are too many. In rare cases I save some tests on a branch for a long time in case the work might be needed. Like tests for transaction from 0 address

Issue #119, opened 2016-10-04
RPC Tests?


danfinlay:
Would it be possible to extend this suite to also test the RPC API responses for the different clients?

Ideally I could submit a blockchain state, and a query, and compare the responses of every client.

chriseth:
We are planning to do this inside http://github.com/ethreum/interfaces or as part of the hive project. The reason is that this repository requires the test runner to be implemented in the clients themselves, while hive provides a generic interface to run the clients.

frozeman:
The rpc test currently require the client to load a BC state and it teste against those. It should also already work with at least cpp and geth.

But if we can do that better in hive, im all for it :P

winsvega:
@cdetrio @holiman @karalabe 
does this implemented on hive? 

I have some work done in this direction for cpp client RPC tests:
https://github.com/ethereum/tests/tree/develop/RPCTests

karalabe:
The old suite is on hive, but that's not maintained. Bas wrote a new suite that I think is a pending PR on hi e.

winsvega:
A topic is to be implemented with retesteth.

PR #572, opened 2018-12-18
RPC tests prototype (for retesteth)

winsvega:
retesteth branch to use: https://github.com/ethereum/retesteth/pull/22
client support: any
exec cmd: ` ./retesteth -t RPCTests`

to attach new client edit the configs: https://github.com/ethereum/tests/tree/develop/Retesteth
a command to select clients with retesteth: `./retesteth -t RPCTests -- --clients 'aleth, parity' `  

you could also try running statetests: ` ./retesteth -t GeneralStateTests`
but your client has to support test RPC interface in order to talk with retesteth: https://github.com/ethereum/retesteth/wiki/RPC-Methods


src/RPCTestsFiller/AvailabilityTests/eth_getBalanceFiller.json
```diff
@@ -0,0 +1,13 @@
+{
+    "eth_getBalance" : {
+        "request_method" : "eth_getBalance",
+	"request_params" : [
+	     "0x1122334455667788991011121314151617181920",
+	     "0x01"
+	],
+	"expect_return" : {
+	     "//" : "exactly checks that returned json object is exactly the same as described here. including all subojects",
+	     "exactly" : "0x0"
```
pipermerriam: I'm thinking that json-schema would work well here rather than defining our own DSL.  Both of the `exactly` and `checkstring` can be expressed in json schema.

```javascript
{
  "type": "string",
  "enum": ["0x0"]
}
```

```javascript
{
   "type": "string"
}
```

If we want to leave this open to be extensible then we could define a `meta` field that defines the validation engine and start with just `jsonschema` as our only validation engine.

winsvega: this is just a notation. but right having an engine field would be even better.
so I could change it to have json shema in the filler itself.

winsvega:
https://github.com/ethereum/tests/issues/119

basically the idea is that retesteth could just ask the client by sending different rpc requests 
and validate the response json.
if clients support test_setChainParams much more tests are possible for rpc requests.

Issue #611, opened 2019-07-12
Test format check failure in just-cloned repo.


acoglio:
I just cloned the repo and `make RLPTests.format` results in a `git diff` error. The `.json` files under the `RLPTests` directory show up as modified.

Have I missed some steps? I followed the instructions under 'Test Set Sanitation' at https://github.com/ethereum/tests.

Background: I'd like to make a PR with some new RLP tests and I understand that I should ensure that the sanitation tests pass first, but the problem is that I'm getting this failure without my new tests.

winsvega:
Oh. I don't support this scripts. And it is probably way outdated. Also would not recommend using this scripts for the test filling. Consensus test execution is a separate tooling. Currently there is `testeth` from `aleth` to run most of this tests. 

To check RLP tests:
```
./testeth -t RlpTests
```

Another project to run consensus tests against different client implementations: 
https://github.com/ethereum/retesteth

winsvega:
https://github.com/ethereum/tests/pull/631

PR #614, opened 2019-07-23
Add RLP invalid tests for insufficient bytes.

acoglio:
These new proposed tests are for invalid encodings that do not have enough bytes
for the lengths indicated in the first few bytes.

The following, repeated from (now merged)  PR #611, apply here as well:
- I tried running the scripts that check the well-formedness of the tests, but ran into problems and created Issue #611. Also, I'm not sure which other requirements should be fulfilled in order for these new tests to be accepted.
- To provide some assurance, the new proposed tests are rejected by the formally verified RLP decoder available at https://github.com/acl2/acl2/blob/master/books/kestrel/ethereum/rlp/decoding-executable.lisp.

### DIFF
--- a/RLPTests/invalidRLPTest.json
+++ b/RLPTests/invalidRLPTest.json
@@ -82,6 +82,51 @@
 	"nonOptimalLongLengthList2": {
 		"in": "INVALID",
 		"out": "f803112233"
+	},
+
+	"emptyEncoding": {
+		"in": "INVALID",
+		"out": ""
+	},
+
+	"lessThanShortLengthArray1": {
+		"in": "INVALID",
+		"out": "81"
+	},
+
+	"lessThanShortLengthArray2": {
+		"in": "INVALID",
+		"out": "a0000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e"
+	},
+
+	"lessThanShortLengthList1": {
+		"in": "INVALID",
+		"out": "c5010203"
+	},
+
+	"lessThanShortLengthList2": {
+		"in": "INVALID",
+		"out": "e201020304050607"
+	},
+
+	"lessThanLongLengthArray1": {
+		"in": "INVALID",
+		"out": "ba010000aabbccddeeff"
+	},
+
+	"lessThanLongLengthArray2": {
+		"in": "INVALID",
+		"out": "b840ffeeddccbbaa99887766554433221100"
+	},
+
+	"lessThanLongLengthList1": {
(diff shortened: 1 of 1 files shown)


RLPTests/invalidRLPTest.json
```diff
@@ -82,6 +82,51 @@
 	"nonOptimalLongLengthList2": {
 		"in": "INVALID",
 		"out": "f803112233"
+	},
+
+	"emptyEncoding": {
```
winsvega, line 87: empty string is 80 
https://github.com/ethereum/tests/blob/develop/RLPTests/rlptest.json#L2

acoglio: This is a test for the empty encoding, not the empty string. An RLP encoding (i.e. the result of an RLP encoder) is never empty, it always has at least one byte. So if an RLP decoder is presented with the empty sequence of bytes, it should reject it because it is not a valid encoding. This test is meant to cover this case. Perhaps we can use a different name if it causes confusion with the test for encoding the empty string.

acoglio:
I added a reply to the comment in the file. This is a test for the empty encoding, not for the encoding of the empty string.

RLPTests/invalidRLPTest.json
winsvega: How to tell the difference in the tests? 
this PR requires aleth changes

acoglio: My understanding is that RLP tests of the form
```{"in" : "INVALID", "out" : <encoding>}```
are meant to test that whether the decoder rejects `<encoding>`, while RLP tests of the form
```{"in" : <string-or-list>, "out" : <encoding>}```
are meant to test whether (1) the encoder produces `<encoding>` given `<string-or-list>` and (2) the decoder produces `<string-or-list>` given `<encoding>`.

If that's the case, then the test for the empty encoding is
```{"in" : "INVALID", "out" : ""}```
as above, while the test for the empty string is
```{"in" : "", "out" : "80"}```
Does this address your question, or have I misunderstood what you were asking?

If aleth accepts the empty encoding, I would argue that it's a bug. The specification of RLP encoding prescribes that both strings and lists start with a byte that describes the type and either the length or the length of the length. So encodings are never empty. If an RLP decoder is given a purported encoding that is empty, it must reject it because it does not encode anything. Do you agree?

winsvega: Yes.
I still will have to change aleth to treat empty string in this test as empty byte input

PR #619, opened 2019-08-05
Blake2f test vectors

shemnon:
Add CALL and CALLCODE tests for the standard unit test vectors for Blake2F, targeting Istanbul.


src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLCODEBlake2fFiller.json
```diff
@@ -0,0 +1,418 @@
...
+      "a94f5374fce5edbc8e2a8697c15331677e6ebf0b": {
+        "balance": "1000000000000000000000",
+        "code": "",
+        "nonce": "0",
+        "storage": {}
+      }
+    },
+    "transaction": {
+      "data": [
+        "0x000000016a09e667f2bd8948bb67ae8584caa73b3c6ef372fe94f82ba54ff53a5f1d36f1510e527fade682d19b05688c2b3e6c1f1f83d9abfb41bd6b5be0cd19137e217907060504030201000f0e0d0c0b0a090817161514131211101f1e1d1c1b1a191827262524232221202f2e2d2c2b2a292837363534333231303f3e3d3c3b3a3938000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000080000000000000000001",
```
winsvega: need a comment of what this bytecode means. you could also use LLL in transaction data

src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLBlake2fFiller.json
```diff
@@ -0,0 +1,418 @@
...
+      "currentDifficulty": "0x20000",
+      "currentGasLimit": "10000000000",
+      "currentNumber": "1",
+      "currentTimestamp": "1000",
+      "previousHash": "5e20a0453cecd065ea59c37ac63e079ee08998b6045136a8ce6635c7912ec0b6"
+    },
+    "expect": [
+      {
+        "indexes": {
+          "data": [0],
```
winsvega, line 14: also a comment about what is the case on data 0

src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLBlake2fFiller.json
```diff
@@ -0,0 +1,418 @@
...
+          "gas": -1,
+          "value": -1
+        },
+        "network": [">=Istanbul"],
+        "result": {
+          "1000000000000000000000000000000000000000": {
+            "storage": {
+              "0x01": "0x01",
```
winsvega: a comment about why expect storage values are expected to be like this

src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLBlake2fFiller.json
```diff
@@ -0,0 +1,418 @@
...
+              "0x02": "0x62305ad4d48dade8269ef60a1bcb8b7bef6e479e643b5ac1f8017e6422ce89fb",
+              "0x03": "0x62d09ecaa81d095e855540dcbc07bd0feb3d4f5e5e50541260ed930f027cfd8d"
+            }
+          },
+          "a94f5374fce5edbc8e2a8697c15331677e6ebf0b": {
+            "balance": "999999999999999915711"
```
winsvega: if possible do not use balance on post condition. in case some gas calculation rule change, this will affect this test.

shemnon: If we don't judge balance how do we test gas calculation?

shemnon:
Is there some sort fo a standard published as to what is expected from a community contributed test case?  These objections seem arbitrary when I can point to other tests cases that don't have these features.  While it is valid to say "past bad behavior does not set a precedent" without a clear document describing what is expected all I have are the past behaviors and I don't know which are bad, which are acceptable, and which are good.

winsvega:
hm. no such standard created yet. good point this notes is well to be published in test creation wiki. 
we have a lot of tests for the past 4 years, it would be good to have more comments in tests, but not enough hands to write both tests, manuals and tools 

and also non of the external contributors mostly ever wrote any .json tests to the repo. so thats why no manuals. now when other clients could write their own tests the manuals are to be created from precedents 

https://github.com/ethereum/tests/pull/620

src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLBlake2fFiller.json
winsvega: if test specificly need to test gas calculation it is ok to use balance in post section.

src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLBlake2fFiller.json
shemnon: described the outputs in the "// data" section, they are all hash values so they are quite opaque.

src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLBlake2fFiller.json
shemnon: covered in "// data" section.

src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLBlake2fFiller.json
shemnon: Gas calculation varies per call (i.e. the blob passed to the precompile has different gas results), so we need the calculations.  Can all remain or should I only leave in a representative few?

shemnon:
Also added a few more test cases.

Also, we should have cross comparison against another before committing.  Some details are not settled yet that are covered by these test cases.

winsvega:
Stay in touch. Once geth have the implementation you could run this test against two clients

holiman:
This may not be the best discussion forum, but one thing that struck me is that it would be neat if all[*]  tests use the `sender` account as `coinbase`  - in this case `0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b`. If we do that, then the postStates (and poststate dumps) become smaller, and we get less questions about "how did account X (coinbase) get into the poststate?" 

[*] `all` meaning all cases where there isn't some explicit test involving the coinbase address

src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLCODEBlake2fFiller.json
shemnon: It has specific EIP-152 meaning, so I pointed that out in the "// data" comments, and provided a test case by testcase description of what is important in that blob.

These are also the same strings found in the reference implementation, with additions.

shemnon:
It totally demolishes gas calculation comparison because all the fees go right back where they started.  Perhaps the contract becomes the coinbase?  But then you would have to explain why the contract's balance goes up in the same way the coinbase would go up.  `"gasPrice": "0"` may be more what we are looking for.

holiman:
Ah, that's true. So I guess the trick is not really that nifty, unless we explicitly want to avoid letting gas calc be part of the test.

winsvega:
more docs: https://github.com/ethereum/retesteth/wiki/Retesteth-commands

holiman:
Geth doesn't pass these tests... 
```
[user@work go-ethereum]$ build/bin/evm statetest /home/user/workspace/tests/GeneralStateTests/stPreCompiledContracts2/CALLBlake2f.json 
[
  {
    "name": "CALLBlake2f",
    "pass": false,
    "fork": "Istanbul",
    "error": "post state root mismatch: got 56c0c7bd91cff99f993ede0733fa3a38e194d01f5c87a9b882305f61c3a4c338, want b46f92edbed07baacf55c64dc3c9169e1f04966f9ce87e81cc107eebd7ce94af"
  },
...
```
Question: is this PR up to date?

shemnon:
Should be up to date.  What does a retesteth run at --verbosity 6 say?  Given that this is a state root mismatch and not off the filler it could be many things.

winsvega:
Also to see the post state 
'--fullstate --poststate'

shemnon:
Is this off of develop/head?  I don't think the contract is firing as there is nothing stored in storage - https://gist.github.com/shemnon/adf5b5d24f05aadbbcd2272405b1c3e0

winsvega:
The precompiled must be supported by the client implementation. You use Pantheon ?

shemnon:
Yes, Pantheon (as it currently is named). it's an internal build off of head.

shemnon:
The failing vectors are d = {0,1,2,6}, which are all the bad data failure cases.  All the "good data" cases are passing.  Geth has ending balance of 999999999999901559711 for 0,1, and 2 whereas pantheon has multiple balances (caused by intrinsic tx cost changes) of around 999999999999999954832 (at gas 0).  Pantheon charges 45,168 wei for this bad call and Geth 98,440,289 at gas 0 and the full limit at gas 1, 2, and 3.  

When the precompile fails due to validation is the correct failure mode to fail the call or to fail the whole tx and charge the gas limit?

winsvega:
I am afraid geth didn't implement the contract yet. If so then its just a call to empty address. If the precompile is implemented but fails then it just eat the whole gas. If the precompile works need to check the results againt other clients. (By putting the results into storage)

holiman:
Random thought: you're calling the precompile directly, state root fails might be related to emptiness and touch, and whether precompiles have special handling on touch-revert etc...Might be better to set 1 wei on the blake address in the prestateOn Aug 23, 2019 8:31 PM, Danno Ferrin <notifications@github.com> wrote:The failing vectors are d = {0,1,2,6}, which are all the bad data failure cases.  All the "good data" cases are passing.  Geth has ending balance of 999999999999901559711 for 0,1, and 2 whereas pantheon has multiple balances (caused by intrinsic tx cost changes) of around 999999999999999954832.  Pantheon charges 45,168 wei for this bad call and Geth 98,440,289 at gas 0 and the full limit at gas 1, 2, and 3.
When the precompile fails due to validation is the correct failure mode to fail the call or to fail the whole tx and charge the gas limit?

holiman:
As of yesterday, geth has everything implemented

shemnon:
@winsvega - none of the chainparams for retesteth have 1 wei for the contracts.  Should those be added?  my guess is everything will need to be refilled if that is done.

Also, eating the whole gas on a bad param seems very aggressive, it's basically a halt on a missed call.  Is halting the VM and treating it as an invalid transaction correct?  Is this also Aleth's behavior?

@holiman - the state root fails because the balances are different.  Geth halts and consumes all gas and charges the account, not letting the transaction continue.  Pantheon only consumes the precompiled call amount and zero gas and returns zeros, letting the caller return and process the data.  So world state is expected to be different.

holiman:
@shemnon here's the very first testcase, the trace and the dump, that I get with `evm`:
```
[user@work go-ethereum]$ build/bin/evm --json --nomemory  --dump statetest /home/user/workspace/tests/GeneralStateTests/stPreCompiledContracts2/CALLBlake2f.json  
{"pc":0,"op":54,"gas":"0x5f58860","gasCost":"0x2","memory":"0x","memSize":0,"stack":[],"depth":1,"refund":0,"opName":"CALLDATASIZE","error":""}
{"pc":1,"op":96,"gas":"0x5f5885e","gasCost":"0x3","memory":"0x","memSize":0,"stack":["0xd4"],"depth":1,"refund":0,"opName":"PUSH1","error":""}
{"pc":3,"op":96,"gas":"0x5f5885b","gasCost":"0x3","memory":"0x","memSize":0,"stack":["0xd4","0x0"],"depth":1,"refund":0,"opName":"PUSH1","error":""}
{"pc":5,"op":55,"gas":"0x5f58858","gasCost":"0x2d","memory":"0x","memSize":224,"stack":["0xd4","0x0","0x0"],"depth":1,"refund":0,"opName":"CALLDATACOPY","error":""}
{"pc":6,"op":96,"gas":"0x5f5882b","gasCost":"0x3","memory":"0x","memSize":224,"stack":[],"depth":1,"refund":0,"opName":"PUSH1","error":""}
{"pc":8,"op":97,"gas":"0x5f58828","gasCost":"0x3","memory":"0x","memSize":224,"stack":["0x40"],"depth":1,"refund":0,"opName":"PUSH2","error":""}
{"pc":11,"op":54,"gas":"0x5f58825","gasCost":"0x2","memory":"0x","memSize":224,"stack":["0x40","0x3e8"],"depth":1,"refund":0,"opName":"CALLDATASIZE","error":""}
{"pc":12,"op":96,"gas":"0x5f58823","gasCost":"0x3","memory":"0x","memSize":224,"stack":["0x40","0x3e8","0xd4"],"depth":1,"refund":0,"opName":"PUSH1","error":""}
{"pc":14,"op":96,"gas":"0x5f58820","gasCost":"0x3","memory":"0x","memSize":224,"stack":["0x40","0x3e8","0xd4","0x0"],"depth":1,"refund":0,"opName":"PUSH1","error":""}
{"pc":16,"op":96,"gas":"0x5f5881d","gasCost":"0x3","memory":"0x","memSize":224,"stack":["0x40","0x3e8","0xd4","0x0","0x0"],"depth":1,"refund":0,"opName":"PUSH1","error":""}
{"pc":18,"op":90,"gas":"0x5f5881a","gasCost":"0x2","memory":"0x","memSize":224,"stack":["0x40","0x3e8","0xd4","0x0","0x0","0x9"],"depth":1,"refund":0,"opName":"GAS","error":""}
{"pc":19,"op":241,"gas":"0x5f58818","gasCost":"0x5ddb204","memory":"0x","memSize":1088,"stack":["0x40","0x3e8","0xd4","0x0","0x0","0x9","0x5f58818"],"depth":1,"refund":0,"opName":"CALL","error":""}
{"pc":20,"op":96,"gas":"0x17d614","gasCost":"0x3","memory":"0x","memSize":1088,"stack":["0x0"],"depth":1,"refund":0,"opName":"PUSH1","error":""}
{"pc":22,"op":85,"gas":"0x17d611","gasCost":"0x320","memory":"0x","memSize":1088,"stack":["0x0","0x1"],"depth":1,"refund":0,"opName":"SSTORE","error":""}
{"pc":23,"op":97,"gas":"0x17d2f1","gasCost":"0x3","memory":"0x","memSize":1088,"stack":[],"depth":1,"refund":0,"opName":"PUSH2","error":""}
{"pc":26,"op":81,"gas":"0x17d2ee","gasCost":"0x3","memory":"0x","memSize":1088,"stack":["0x3e8"],"depth":1,"refund":0,"opName":"MLOAD","error":""}
{"pc":27,"op":96,"gas":"0x17d2eb","gasCost":"0x3","memory":"0x","memSize":1088,"stack":["0x0"],"depth":1,"refund":0,"opName":"PUSH1","error":""}
{"pc":29,"op":85,"gas":"0x17d2e8","gasCost":"0x320","memory":"0x","memSize":1088,"stack":["0x0","0x2"],"depth":1,"refund":0,"opName":"SSTORE","error":""}
{"pc":30,"op":97,"gas":"0x17cfc8","gasCost":"0x3","memory":"0x","memSize":1088,"stack":[],"depth":1,"refund":0,"opName":"PUSH2","error":""}
{"pc":33,"op":81,"gas":"0x17cfc5","gasCost":"0x3","memory":"0x","memSize":1088,"stack":["0x408"],"depth":1,"refund":0,"opName":"MLOAD","error":""}
{"pc":34,"op":96,"gas":"0x17cfc2","gasCost":"0x3","memory":"0x","memSize":1088,"stack":["0x0"],"depth":1,"refund":0,"opName":"PUSH1","error":""}
{"pc":36,"op":85,"gas":"0x17cfbf","gasCost":"0x320","memory":"0x","memSize":1088,"stack":["0x0","0x3"],"depth":1,"refund":0,"opName":"SSTORE","error":""}
{"pc":37,"op":0,"gas":"0x17cc9f","gasCost":"0x0","memory":"0x","memSize":1088,"stack":[],"depth":1,"refund":0,"opName":"STOP","error":""}
{"output":"","gasUsed":"0x5ddbbc1","time":483660}
{"stateRoot": "56c0c7bd91cff99f993ede0733fa3a38e194d01f5c87a9b882305f61c3a4c338"}
[
  {
    "name": "CALLBlake2f",
    "pass": false,
    "fork": "Istanbul",
    "error": "post state root mismatch: got 56c0c7bd91cff99f993ede0733fa3a38e194d01f5c87a9b882305f61c3a4c338, want b46f92edbed07baacf55c64dc3c9169e1f04966f9ce87e81cc107eebd7ce94af",
    "state": {
      "root": "56c0c7bd91cff99f993ede0733fa3a38e194d01f5c87a9b882305f61c3a4c338",
      "accounts": {
        "0x1000000000000000000000000000000000000000": {
          "balance": "0",
          "nonce": 0,
          "root": "56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
          "codeHash": "af6114367b68d5e075d6a789df044b268cbbb0b2aac57b9d7b72aabc8b0808e8",
          "code": "36600060003760406103e8366000600060095af16001556103e85160025561040851600355"
        },
        "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba": {
          "balance": "98440289",
          "nonce": 0,
          "root": "56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
          "codeHash": "c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"
        },
        "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b": {
          "balance": "999999999999901559711",
          "nonce": 1,
          "root": "56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
          "codeHash": "c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"
        }
      }
    }
  }
]
```

Notably, the call to `blake` is here:

```
{"pc":19,"op":241,"gas":"0x5f58818","gasCost":"0x5ddb204","memory":"0x","memSize":1088,"stack":["0x40","0x3e8","0xd4","0x0","0x0","0x9","0x5f58818"],"depth":1,"refund":0,"opName":"CALL","error":""}
{"pc":20,"op":96,"gas":"0x17d614","gasCost":"0x3","memory":"0x","memSize":1088,"stack":["0x0"],"depth":1,"refund":0,"opName":"PUSH1","error":""}
```
Gas is `0x5f58818` (`99977240`). It leaves a `0x0` on stack, which is failure -- thus it loses all the gas it sent along. And then it continues, as expected.

shemnon:
I think you missed that my question was if a hard failure (return code zero) vs a soft failure (empty return) is what is appropriate.  The other precompilers lean towards soft failures (ec recovery returns an empty string on an invalid signature, by yellow paper spec, some short inputs are zero extended).  But it is not without precedent, altbn128 paring returns a hard failure if not on curve or in group.

Test cases are updated so the input failures produce a hard failure.  Pantheon and Geth passed this set.

axic:
> I think you missed that my question was if a hard failure (return code zero) vs a soft failure (empty return) is what is appropriate. The other precompilers lean towards soft failures (ec recovery returns an empty string on an invalid signature, by yellow paper spec, some short inputs are zero extended).

I'd opt for hard failures. In the case of soft failure the caller must inspect the returned data to see if it failed. In the case of a hard failure the caller can rely on the result code of the call.

axic:
What is blake2f btw?

holiman:
Yes, so the eip is pretty well specced about what are errors, and thus cause what you call hard failure

holiman:
> Test cases are updated so the input failures produce a hard failure. Pantheon and Geth passed this set.

@shemnon updated where? Did you forget to push?

shemnon:
They got pushed to the other PR.  That's what I get for test-driving VSCode - https://github.com/ethereum/tests/pull/627/commits/502a10c93e3c47763765c83e528f0cf4081921bc - I'll cherry pick them

holiman (approved):
Geth passes all these tests (and sheesh, that maxrounds is truly timeconsuming)

winsvega:
Looks good. Just for the test sheet: What are the cases covered with this tests?
I will add this node to the table of coverage

shemnon:
I moved the tests to match the EIP ordering.  It's all 9 tests from the EIP plus an all zeros test and an 8 million rounds test.

winsvega:
Have you regenerated the tests? 
```
== test.py: ERROR: Error getting hash of the yml/json: src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLCODEBlake2fFiller.json
378== test.py: ERROR: Exception:  Expecting value: line 166 column 5 (char 4869)
379== test.py: ERROR: Filler hash is different: ./GeneralStateTests/stPreCompiledContracts2/CALLCODEBlake2f.json  HashSrc: ' error '
380== test.py: ERROR: Error getting hash of the yml/json: src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLBlake2fFiller.json
```

winsvega:
@axic  do you suggest a change to this precompile implementation? If that change to be implemented this tests would be affected.

winsvega:
@axic posting new proposals again for this precompile: 
https://github.com/ethereum/EIPs/issues/152#issuecomment-527436786

holiman:
Nothing new decided, please merge this

winsvega:
hmm. @shemnon  how have you created those tests? 
I am missing genesis description for Istanbul in Retesteth configuration. 
perhaps you used an older version of retesteth. 

I've moved genesis configuration to config files. Now you can set it in
`/tests/Retesteth/default/genesis`

And I also added more checks for test validity. Here I get some transaction report from your tests:
transaction that are missing the expect section: 

```
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 0, g: 1, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 0, g: 2, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 0, g: 3, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 1, g: 1, v: 0 (CALLBlake2f)
```

Please validate your tests and regenerate using latest verion of retesteth 
from branch bcgen (https://github.com/ethereum/retesteth/pull/48)

add the missing expect sections and use latest solidity lllc which is now supported.
latest lllc support chainid and selfbalance opcodes.

winsvega (changes requested):
please add expect section for uncovered transactions
and refill with latest retesteth (https://github.com/ethereum/retesteth/pull/48)
and with latest lllc version

rebase on tests:develop to fetch the retesteth config first

shemnon:
I think the no TXes complaint is from the lack of the Istanbul config.  I pulled over the one I had in place for Martin's changes.

winsvega:
hmm. how you created this tests? did you get any errors? 
I see this errors on that retesteth version:

```
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 0, g: 1, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 0, g: 2, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 0, g: 3, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 1, g: 1, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 1, g: 2, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 1, g: 3, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 2, g: 1, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 2, g: 2, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 2, g: 3, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 3, g: 1, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 3, g: 2, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 3, g: 3, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 9, g: 1, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 9, g: 2, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 9, g: 3, v: 0 (CALLBlake2f)
Error: Test has transaction uncovered with expect sections!  (CALLBlake2f)
```

meaning there are some transactions that does not have post condition checks

shemnon:
```
dannos-mbp:build dannoferrin$ PATH+=":../../../ethereum/solidity/build/lllc/" ./retesteth/retesteth -t  GeneralStateTests/stPreCompiledContracts2 -- --testpath ../../../shemnon/ethereum-tests/ --clients besu --singletest CALLBlake2f --filltests --fillblockchain
Unknown option: --fillblockchain
dannos-mbp:build dannoferrin$ PATH+=":../../../ethereum/solidity/build/lllc/" ./retesteth/retesteth -t  GeneralStateTests/stPreCompiledContracts2 -- --testpath ../../../shemnon/ethereum-tests/ --clients besu --singletest CALLBlake2f --filltests --fillchain
Running tests using path: "../../../shemnon/ethereum-tests/"
Running 1 test case...
Active client configurations: 'besu '
Filter: 'CALLBlake2f'
Running tests for config 'Hyperledger Besu on TCP' 2
Test Case "stPreCompiledContracts2": 
100%

*** No errors detected
dannos-mbp:build dannoferrin$ PATH+=":../../../ethereum/solidity/build/lllc/" ./retesteth/retesteth -t  GeneralStateTests/stPreCompiledContracts2 -- --testpath ../../../shemnon/ethereum-tests/ --clients besu --singletest CALLCODEBlake2f --filltests --fillchain
Running tests using path: "../../../shemnon/ethereum-tests/"
Running 1 test case...
Active client configurations: 'besu '
Filter: 'CALLCODEBlake2f'
Running tests for config 'Hyperledger Besu on TCP' 2
Test Case "stPreCompiledContracts2": 
100%

*** No errors detected
dannos-mbp:build dannoferrin$ PATH+=":../../../ethereum/solidity/build/lllc/" ./retesteth/retesteth -t  GeneralStateTests/stTimeConsuming -- --testpath ../../../shemnon/ethereum-tests/ --clients besu --singletest CALLBlake2f_MaxRounds --filltests --fillchain
Running tests using path: "../../../shemnon/ethereum-tests/"
Running 1 test case...
Skipping stTimeConsuming because --all option is not specified.

*** No errors detected
dannos-mbp:build dannoferrin$ PATH+=":../../../ethereum/solidity/build/lllc/" ./retesteth/retesteth -t  GeneralStateTests/stTimeConsuming -- --testpath ../../../shemnon/ethereum-tests/ --clients besu --singletest CALLBlake2f_MaxRounds --filltests --fillchain --all
Running tests using path: "../../../shemnon/ethereum-tests/"
Running 1 test case...
Active client configurations: 'besu '
Filter: 'CALLBlake2f_MaxRounds'
Running tests for config 'Hyperledger Besu on TCP' 2
Test Case "stTimeConsuming": 
100%
```

winsvega:
No. I don't understand why. the tests state that retesteth used commit is from master branch. but the behaviour of your retesteth is way outdated. It is filling the blockchain tests into a wrong folder name. and it does not perform the state test validity check. 

could you double check that you running the tool on latest commit from the master branch? 
https://github.com/ethereum/retesteth/commit/eeb48052b47173ca7f18585bd1a3b609b930732d

shemnon:
at @eeb4805 I get a whole lot of `WARNING: Error getting LLLC Version`

shemnon:
lllc version is hand fed because the lllc version logic is broken on retesteth on a mac.

winsvega:
this is very strange. you must at least see this errors:
```
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 0, g: 1, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 0, g: 2, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 0, g: 3, v: 0 (CALLBlake2f)
Error: A transaction was specified, but there is no execution results in a test! CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0 (Errored TR: d: 1, g: 1, v: 0 (CALLBlake2f)
```

how you deal with lllc version problem ?

shemnon:
I don't see those errors, never have.

I hand edited the LLLC versions into the JSON.  I think there is some problems with the way you detect if LLLC is present on mac.

winsvega:
I've just compiled lllc and retesteth on mac and created a test for it: 
https://circleci.com/gh/ethereum/retesteth/665?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

the build went through just fine. smth is wrong on your end. 

> I hand edited the LLLC versions into the JSON. I think there is some problems with the way you detect if LLLC is present on mac.

You can't just edit those fields. You provide misinformation about the test. If smth went wrong with lllc then the test is not filled correctly. lllc version is only a warning if when you fill the tests there is no lllc code in the test.

shemnon:
LLLC version - when I put the LLLC setting in the exported environment instead of as a shell parameter the error stops occuring.  You will note the version is exactly the same as was copied in by hand.

https://github.com/ethereum/retesteth/pull/52 version includes the v,s,r values that were complained about in another thread.  So I re-ran the tests to get those in the blockchain tests.

The transaction errors are flakey, so I think it's something inside of retesteth.  I think it is a timing issue as I didn't see them until today and I see them less than 10% or the time and only under load (running a compile in another thread or some ad-filled chrome webpage).  The JSON-RPC traces are mostly the same, except that a success will continue with more JSON-RPCs to do the fill.  https://gist.github.com/shemnon/72cd75520148dc546e995e63a5ecd938

Have you tried to local test against geth or besu and do you see the same errors consistantly?

winsvega:
yes the process is fully deterministic.  how can I build pantheon/besu on ubuntu?

shemnon:
The full version is [here](https://github.com/hyperledger/besu/blob/master/docs/development/building.md).

The short version for this case is
* Install Java JDK 11+ (make sure `javac` and `java` are on the path)
* clone the repo
* `./gradlew installdist`
* `./build/install/besu/bin/besu retesteth`

src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLCODEBlake2fFiller.json
```diff
@@ -0,0 +1,217 @@
...
+        "3 - f flag wrong value",
+        "4 - RFC 7693 Appendix-A, zero rounds, final",
+        "5 - RFC 7693 Appendix-A, twelve rounds, final",
+        "6 - RFC 7693 Appendix-A, twelve rounds, not final",
+        "7 - RFC 7693 Appendix-A, one round, final",
+        "8 - all zero input, correct length, zero rounds, not final",
+        "9 - RFC 7693 Appendix-A, 8 million rounds"
+      ],
+      "data": [
+        "0x",
```
gumb0, line 230: Another test suggestion: non-empty data, but fewer than 4 bytes, to catch possible errors of trying to calculate the precompile cost out of it

shemnon: added as data 10

shemnon:
Two things

(a) the docs and errors were not particularly clear that all permutations must be covered.  The error message could be re-tooled to communicate this better

`Missing transaction execution results.  All permutations for a test transaction must be covered. Please add the validation data for CALLBlake2f, fork: Istanbul, TrInfo: d: 9, g: 0, v: 0`

(b) This error must be produced all the time on all platforms.  I only rarely see it on my macbook pro, and then only when I put the system under load.  This feels like a memory fencing or threading issues, like data is getting overwritten before it is moved to a more permanent location.

winsvega:
(b) 
how is this possible? 

the error is added to error list upon test execution. then all errors from the list are stdeerored once threads are merged. so the output is done at the main thread.

shemnon:
I don't know how it's possible, but I have observed that it does occur.  Perhaps a memory fence needs to be raised prior to the data check?

PR #626, opened 2019-08-14
fix some flaws in transaction test fillers  + EIP2028 fillers

holiman:
This PR addreses some quirks in the fillers, namely
- removes the use of arithmetic within json fields, by replacing all instances of e.g. `2^63-1` with the hex equivalent, 
- fixes some cases of shadowed fields, where e.g. `s` is present in two locations, both representing the secretkey and also the actual `s` value

PR #629, opened 2019-08-27
EIP2028 transaction tests - testeth-compatible version

gumb0:
I adapted Transaction tests from https://github.com/ethereum/tests/pull/626 to be used in testeth and added filled tests, too.

Comments in fillers unfortunately had to go for now, testeth doesn't support those, but I guess that is fixable.


winsvega:
I can work on this issue this weekend

winsvega:
moved to https://github.com/ethereum/tests/pull/630

PR #631, opened 2019-09-01
UpdateReadme

winsvega:
make an actual description for the test repo users


holgerd77:
Sorry, still reading here but wasn't quick enough with a review.

PR #642, opened 2019-09-25
correct postState to postStateHash in LegacyTests

winsvega:


holgerd77:
Is this really the correct PR description (title)? I am not seeing any reference to the LegacyTests location in the paths of the files touched?

winsvega:
Look at commits title rather

https://github.com/ethereum/tests/commit/0c021653ad1db360a6d42017507e47a485eca517

holgerd77:
Still completely not getting what has happened here, If one is doing a [search](https://github.com/ethereum/tests/search?q=postStateHash&unscoped_q=postStateHash) now there are 17 `BlockchainTests` now having a `postStateHash` field instead of `postHash`?

Anyhow, will include in the release notes with some additional linking.

winsvega:
Just renaming.
Hash instead of post state when state is too huge. Like megabytes

holgerd77:
Ah ok. This can be problematic though when people rely on the `state` name being there, so this is technically a breaking change.

Anyhow, have added to the release notes if someone needs to be aware of this.

holiman:
It's also pretty useless, since the `postStateHash` is, by definition, the `stateRoot` of the last successfully imported block. 
See https://github.com/ethereum/tests/blob/b5eb9900ee2147b40d3e681fe86efa4fd693959a/LegacyTests/Constantinople/BlockchainTests/GeneralStateTests/stQuadraticComplexityTest/Create1000Byzantium_d0g1v0.json#L25-L69

PR #649, opened 2019-10-29
Tests for EIP-1706/EIP-2200

shemnon:
Test for Out of Gas for EIP-1706/EIP-2200.  Each of the three mutating call
operations is called with enough gas to hit the non-mutatiing SSTORE with
2299, 2300, and 2301 gas.  Because it is a non-mutating SSTORE the gas
should be SLOAD cost - 800 in Istanbul.  But the security claues in
EIP-1706/EIP-2200 will apply and only the 2301 gas call should succeed
and not the other two calls.

Some clients did less than the stipend and not less than or equals, which is
why a test at the stiped is part of this suite.


winsvega (approved):

winsvega:
I've added the log for better release process: 
https://github.com/ethereum/tests/blob/develop/PRLOG.md

you could leave some comment in that

PR #656, opened 2019-11-12
Enable Istanbul in aleth's retesteth configs

gumb0:


winsvega:
I have a PR log now ) 
https://github.com/ethereum/tests/blob/develop/PRLOG.md

PR #659, opened 2019-11-13
Fix ripemd160 precompile name in retesteth configs

gumb0:


gumb0:
forgot Istanbul.json...

PR #663, opened 2019-11-21
Removing execution permission from json test

evertonfraga:
That's an addition to #662.

`chmod -x difficultyEIP2384_random.json`.

Permission diff: `100755 → 100644`


winsvega:
git preserve the permission on files ?

evertonfraga:
I didn't know that too!

PR #665, opened 2019-11-26
Added KeyStore test based on MyCrypto file

spetz:
After updating the AES-128-CTR implementation in Nethermind client, I found out that it would make sense to include an additional Keystore test in order to ensure that the generated file is compatible with MyCrypto and Metamask.


winsvega:
@gumb0 , does aleth run keystore tests ?

gumb0:
@winsvega looks like we do run them (try `testeth -t Crypto/KeyStore`)

PR #685, opened 2020-04-27
[EIP2315] Add tests for subroutines

matkt:
Addition of tests for the https://eips.ethereum.org/EIPS/eip-2315. Tests generated with the current Besu implementation of the EIP2315.

The tests are those present in the description of the EIP.

Need update of retesteth https://github.com/ethereum/retesteth/pull/90

PR #688, opened 2020-05-13
Reduce balances in SHA3 tests

shemnon:
Reduce the balances in some SHA3 vm tests from 2^256-1 to 2^248-1 to
avoid issues relating to UInt256 rollover.


winsvega:
I actually find it an interesting case. do we have a balance limit officially? 
YP says it can be u256.
do we have a protocol overflow behaviour?

shemnon:
No formal one, but several informal:
* Both OpenEtherem and Besu fail because we use UInt256 for value.
* The BALANCE operation in the EVM will truncate to the lower 256 bits
* Most persuasive, the reference tests themselves truncates to 256 bits in it's post conditions: https://github.com/ethereum/tests/pull/688/files#diff-ce232a19b558d3780ea5e8ed874f05e7R106
  * (geth passes this test!)

shemnon:
As for the travis failures, I have no idea why the source hash is failing.  Do I need to use a newer build of retesteth?

winsvega:
The failures are testeth version of VMTests. I can update those.

holiman:
> The failures are testeth version of VMTests. I can update those.

Did  you already or will you, @winsvega ?

holiman (approved):

shemnon:
Where can on find and build a current version of testeth from scratch?  Ideally on mac.

winsvega:
Its in ethereum aleth repo

winsvega:
Hold on, we actually have a new intern to be able to do this )
@sr9000

sr9000:
Be patient, I'll working on it.

holiman:
> I actually find it an interesting case. do we have a balance limit officially?

In any case, if it's worthwhile to have a test for the "unofficial gas limit", then that test should be called e.g. `TestGasAboveUint64` and `TestGasAboveInt63` etc. It's odd/quirky to have tests which doesn't match the actual description.

sr9000:
@holiman please, check pr to your repository. All tests passed successfully.

adria0:
> > I actually find it an interesting case. do we have a balance limit officially?
> 
> In any case, if it's worthwhile to have a test for the "unofficial gas limit", then that test should be called e.g. `TestGasAboveUint64` and `TestGasAboveInt63` etc. It's odd/quirky to have tests which doesn't match the actual description.

What about the https://eips.ethereum.org/EIPS/eip-1985? It will be nice to have something like this in this cases.

sr9000:
> What about the https://eips.ethereum.org/EIPS/eip-1985? It will be nice to have something like this in this cases.

@adria0 
Did evm already deal with fields overflow? I see explicit field ranges as part of guaranteed termination/flag setting behavior in such scenarios.

adria0:
@sr9000, it deals in a very explicit way, arithmetic operations are prefixed with `saturating_`,  `checked_` or `overflowing_`, where mainly all types are U256. But not sure, am I answsering your question?

shemnon:
So... what's on travis doesn't like berlin?  Will aleth/testeth get berlin cofings?

winsvega:
There is additional .py script layer that verify the test schema. Apparently it checks that Berlin is in allowed list. It isbin json schema folder

shemnon:
So it turns out these are not even my failure. #685 was merged in with broken tests.  I'm finding the whole process of trying to contribute reference test very frustrating between the lack of follow up and the lack of consistency.

shemnon:
I fixed the travis bits to add Berlin.  But now the subroutine hashes in that PR are wrong. Because #685 was merged when the build was red.

sr9000:
> @sr9000, it deals in a very explicit way, arithmetic operations are prefixed with `saturating_`, `checked_` or `overflowing_`, where mainly all types are U256. But not sure, am I answsering your question?

@adria0 yes, you do, thanks!

winsvega:
> I fixed the travis bits to add Berlin. But now the subroutine hashes in that PR are wrong. Because #685 was merged when the build was red.

yeah, sorry it is a bit chaotic at this moment. with all new test suites and new fork name. 
merged:
https://github.com/ethereum/tests/pull/695

those .py scripts are the second layer that detects issues in tests: such as if test is not generated from filler (hash check).

PR #692, opened 2020-05-28
New v8.0.0-beta.1 Berlin HF release

holgerd77:
Now that https://github.com/ethereum/tests/pull/685 got merged I would like to directly continue with a first `Berlin` HF release after the preparatory final [v7.0.0](https://github.com/ethereum/tests/releases/tag/v7.0.0) release.

Then clients have a first basis and some comprise information to integrate on the subroutine tests. This might also help on foster reflection on additional edge and error cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,54 @@ Here is an example how a follow-up release line could look like:
 
 # Release Notes
 
+## v8.0.0-beta.1
+
+First `Ethereum` tests release with partial support for the
+[Berlin](https://eips.ethereum.org/EIPS/eip-2070) hardfork with first tests
+for EVM subroutines support (thanks @matkt 😀).
+
+[Here](https://github.com/ethereum/tests/blob/develop/BlockchainTests/GeneralStateTests/stSubroutine/beginSubAtEndOfCode.json)
+is an example for an added `Berlin` HF test file to the `BlockchainTests` with
+a `*_BERLIN` postfix on the testname and "Berlin" added to the `network` parameter.
+
+The corresponding [state test](https://github.com/ethereum/tests/blob/develop/GeneralStateTests/stSubroutine/beginSubAtEndOfCode.json) file has a `Berlin` entry in the `post` section of the test file.
+
+There is no change in the `LegacyTests` folder structure yet and test files
+there contain tests until `ConstantinopleFix` (aka `Petersburg`).
+
+### Subroutine Tests
+
+Subroutine support for the EVM has been specified in 
+[EIP-2315](https://eips.ethereum.org/EIPS/eip-2315) (Status: `Draft`) and accepted
+for `Berlin` HF inclusion in
+[ACD Call #84](https://github.com/ethereum/pm/blob/master/All%20Core%20Devs%20Meetings/Meeting%2084.md#decisions-made).
+
+PR [#685](https://github.com/ethereum/tests/pull/685) adds first subroutine tests
+to the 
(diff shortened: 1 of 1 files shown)


holgerd77:
//cc @matkt

holiman:
Hold it :) 
I just proposed some changes; https://github.com/ethereum/EIPs/pull/2676 . It's implemented in geth already, should be fairly trivial for clients to update. We need to regenerate the testcases though :/

holgerd77:
@holiman No prob, already valuable to have this info collected here. Will wait until you give a go respectively update here then. 😄

matkt:
I just added a new test for subroutines here https://github.com/ethereum/tests/pull/693 (not merged yet)

holgerd77:
@holiman Not completely followed all conversations, is the current `develop` branch with the merge of https://github.com/ethereum/tests/pull/693 now in a stable, consistent and correct state and also integrates correct tests for a final EIP-2315 version?

So would this be a good time to update here and do a release?

winsvega:
For the release all tests should have Berlin generation. I am preparing an update for that.

holgerd77:
Ok

holiman:
I would propose not Berlin, but Yolo1Whatever Berlin will contain, it will probably not be exactly like YoloV1

winsvega:
Regards to naming it is possible to assign any kind of name for the tests. 
Then in a client configuration use actual name.

The reason we regeenrate all tests is to make sure that existing cases are not affected by new eips. 
The same way it is possible to create tests for fork+eip rules.

winsvega:
Wasn't it 7.0.0 release?

holgerd77:
@winsvega No, the current concept is to always have a major version bump for every major hardfork and [v7.0.0](https://github.com/ethereum/tests/releases/tag/v7.0.0) was the finalizing release of the "pre-Berlin era" 😄 , so on a `Petersburg` (and `MuirGlacier`) state.

The first `v8` beta release introduces the `Berlin` release series ending up at some point in a stable `v8` release with all finalized `Berlin` EIPs included (in our case with `v7`we were just pretty pretty late, maybe that's some source for confusion here).

holgerd77:
@winsvega Any updates on the Berlin test regeneration?

winsvega:
No, only transition tool support. small test modifications. 

Subroutines tests I see.  No BLS12-381 curve operations was PR'ed

holgerd77:
Hi guys, just as a pre-announcement: I will try to pick this up again this week or latest next week. 😄  To reiterate from what we've discussed on our call: then we can release this here as a major release `v8.0.0` (so no `beta` any more) and you guys can pick up from there.

As some summary for @holiman and others: in the call with @qbzzt and @winsvega together with the JavaScript team we came to the conclusion that we want to move away from these HF-centric releases - which just turned out to not be practicable with these unpredictable HF dates - and instead do "normal" semantic versioning releases. So we simply do major releases every time there is a breaking change on the consumer (client developers) side, so e.g. a directory name change, a HF moved from "active" to "legacy", stuff like that. In this new scenario a new HF addition might become rather a minor release in the sense of a new feature added - at least if there is nothing breaking added along.

For us as a JavaScript team these releases are still important since they significantly help us to follow up with the changes on this repo and ensures a smooth upgrade process without breaking anything on our side. So it would be great if we are ending up with process of just doing regular releases evey 3-4 weeks or so, independent from the level of changes being done. Then we as a client team can easier follow-along and we will likely have shorter upgrade cycles.

Comments from representatives/other devs from other client teams highly welcome!

holgerd77:
Superseded by the latest releases, will close.

PR #698, opened 2020-06-14
Update test documentation

winsvega:
Updated:
 * Blockchain Test
 * General State Test
 * VM Test

Files redisigned. It's a preparation step for other instructions on how to generate tests with retesteth
an Filler format description.

@holgerd77 
I am trying to organize the docs. would need you help with rst

with this basic files describing each section separately it is possible to construct general state test description. I think docs should have less text (the current version has a lot of unreadable text) and more diagrams so it would be possible to easily grasp the meaning just by looking at the docs


holgerd77:
@winsvega That's a good idea!

Will have a look within the next 1-3 days.

holgerd77:
What kind of help do you need here, do you need help with the RST format and how you link this into the broader docs? Or do you rather need a general review here and help on the admin side to generate the docs after merge?

Yeah, cutting on too extensive text parts is likely a good idea, you should nevertheless also be careful not too throw away valuable information. Always an act of balance to find a good middle point here.

winsvega:
How does docs are geberated? Is it automatically update with this repo?

Is it possible to display contents of rst file in another rst?

holgerd77:
If we have a webhook from ReadTheDocs added to this repo it should build automatically - not sure - in doubt ping me once your changes are merged. You can also re-trigger the build yourself since you are added to ReadTheDocs (just checked).

Have never embedded an rst in an rst and also can only search the docs, [this](https://docutils.sourceforge.io/docs/ref/rst/directives.html#including-an-external-document-fragment) might work here, eventually some security considerations. If you don't want re-use text parts it might be easier to put everything into the respective outer files.

winsvega:
take a look at this style. I created gstate_test.rst and it has include fileds. so don't have to describe same sections all over again.

holgerd77:
(sorry that it took longer than expected)

Issue #683, opened 2020-04-22
test STATICCALL violation in call to precompile


MrChico:
Found a [bug in hevm](https://github.com/dapphub/dapptools/pull/360) where CALLs with nonzero values to precompiles were violating the staticmode flag correctly. This does not seem to be covered by any tests right now. As clients often have special logic for dealing with calls to precompiles, it could be valuable to test various CALLs to precompile cases explicitly.

PR #700, opened 2020-06-18
fix issue #683 test STATICCALL violation in call to precompile

sr9000:
fix issue #683 test STATICCALL violation in call to precompile
- staticcall from transaction
- staticcall from called contract
- staticcall from initialization contract code


winsvega:
excelent. 

* add comment section to the fillers with a description of whats happenning with what transaction in the test.
```
  "_info" : {
     "comment" : "Description"
}
```
* commit --filltests results
* add one more test with bytecode from issue 683
* update PRLOG file in the tests repo

PR #707, opened 2020-06-29
TransactionTest doc update

winsvega:
clarify field types more

Have warnings of dublicated topics. is that ok to ignore? 
(because we include same file multiple types)


holgerd77:
Overlapping topics is not ideal because internal linking won't work properly then.

I would suggest to move the topic directives (e.g. `.. block:`) out of the included files (like `test_types/TestStructures/info.rst`) to the outer file where the inclusion is done (like `blockchain_tests.rst`) and then give this a name-spaced/context aware topic name like `..blockchain_tests_block` or similar. This should solve it.

holgerd77:
Every topic needs to be on top of the respective `.. include::` directive then, so that internal linking is jumping to the correct section of the text.

PR #712, opened 2020-07-09
EIP 2537 state tests

shamatar:
Based on the original test vectors from [my repo](https://github.com/matter-labs/eip1962/tree/master/src/test/test_vectors/eip2537) (linked in EIP text).

All test pass on monkey-patched (add new test YoloV1 network for `retesteth` command) Geth based on commit `fa01117498a580f8de35e026f7e0ab38313868dd` from https://github.com/ethereum/go-ethereum master branch.

Test fillers are committed without other changes (I suppose).

PR #713, opened 2020-07-10
stEIP2537

winsvega:


jochem-brouwer:
Hey there. This PR looks like it only creates tests for Berlin. It is thus not explicitly tested that pre-Berlin one cannot access the EIP 2537 precompiles. Pre-Berlin the storage slot should yield the string of an empty keccak (as the staticcalls will return an empty string, but they will succeed).

jochem-brouwer:
What's also not in these state tests are tests for empty inputs on the G1/G2 multiexponentation and pairing. These should be added. There are also no tests with inputs which do not have the expected length (i.e. the length of the input `mod` some constant is nonzero).

CC @shamatar

jochem-brouwer:
Also what might be a good idea is to add 4 tests for `MapToG1` and `MapToG2` where the `Fp` and `Fp2` points are the `Base field modulus` (error) and `Base field modulus - 1` (no error).

shamatar:
There are negative tests for FpToG1 mapping (I think for Fp2 too) where field element is larger than modulus. You can make extra ones and PR to 1962 repo, there is a code there to make positive/negative tests and dump them to CSV.

winsvega:
source for contract bytecode in fillers or generator script ?

shamatar:
I've posted it in the original PR https://github.com/ethereum/tests/pull/712#issuecomment-656246408

winsvega:
The generator script? The tests txs also contain data.

PR #716, opened 2020-08-22
Getting Started with State Tests

qbzzt:
A tutorial that explains how to write state tests with LLL (Solidity will be added when it is available)


qbzzt:
I'll add a solidity test and then PR again.

PR #717, opened 2020-08-23
State Transition Test Tutorial

qbzzt:
Now with Solidity tests.


qbzzt:
I'm going to integrate it with the Sphinx documentation.

PR #719, opened 2020-08-26
Added the retesteth and state transition test tutorials, integrated into sphinx

qbzzt:
1. Retestheth tutorial
2. State transition tests tutorial
3. Integrate into Sphinx and looks OK
4. Sample tests included

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -10,13 +10,32 @@ Ethereum Tests
 .. note::
    See :ref:`contribute` if you want to help improve this documentation.
 
+
+
 .. toctree::
    :maxdepth: 3
-   :caption: Contents:
+   :caption: Tutorial:
+
+   retesteth-tutorial.rst
+   state-transition-tutorial.rst
+
+
+
+.. toctree::
+   :maxdepth: 3
+   :caption: Reference:
 
    testsuite.rst
-   generating-tests.rst
    contribute.rst
+
+
+
+
+.. toctree::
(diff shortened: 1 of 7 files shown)


winsvega (approved):
this looks really greate. now we have updated tutorials on how to make a simple test with newest tools

holgerd77 (approved):
Not able to give this an in-depth review since I am not deeply enough in the process. But I build this locally and this looks fantastic 🎊, thanks for the wonderful work @qbzzt!

I would say this can nevertheless definitely merged for now since this will also improve over time and "on usage". Some of our team members (JavaScript/TypeScript) will very likely give this a try along the way towards new EIP implementations and then we will for sure also give some feedback and eventually also contribute with new tests.

So, congrats on this new documentation milestone! ✨

qbzzt:
Thank you. The way I see it in theory there are three tutorials left:

1. Writing blockchain tests
2. Getting a new client to run with retesteth
3. Dealing with Ethereum forks (transition tests, etc).

#3 is probably not worth doing, because forks are rare and Dimitri will
cover them for the foreseeable future.

#2 is a judgement call I'm not knowledgeable enough to make. How often do
we expect new clients? From what I understand there are concerns about
client diversity, so this might be something we want to encourage and make
as easy as possible.

#1 is my next task

After I finish the tutorials I'll know enough to document actual tests.

qbzzt:
I need to add --singletest (and probably --singlenet)

PR #720, opened 2020-08-27
Support to forks >=Istanbul in the expectException

qbzzt:
Now that the docker also runs Berlin tests, this test fails without that.


winsvega:
If the test is not generated for Berlin it wont be run on Berlin.
And many tests will show errors if regenerated with Berlin rules. 

Also changing a filler require test regeneration. The python script failed to detect this.

But retesteth must show an error when running the test

qbzzt:
I ran it with my change, and it worked for both Istanbul and Berlin. If it isn't supposed to run on Berlin, then we should replace

**"network" : [">=Istanbul"]** 

with

**"network" : ["Istanbul"]**

Also, retesteth does show an error.

PR #722, opened 2020-09-07
Added the reference section

qbzzt:
Also added to the state transition tutorial an example of a multitest file

### DIFF
--- a/docs/eip-tutorial.rst
+++ b/docs/eip-tutorial.rst
@@ -40,7 +40,6 @@ To run this test case, execute these commands:
 
 The Test Source
 ---------------
-
 Normally we don't want raw
 machine language code in tests if that can be avoided, so **retesteth** emits a 
 warning. However, in this case we are taking the code directly from the EIP and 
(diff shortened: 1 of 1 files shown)


holiman:
Nice! Well done, this is going to be very useful!

I have some comments about hte EIP tutorial. First of all, I think you're doing good here, so I hope you don't take my comments in a negative way..



- I don't understand why there is talk about "first test case" having that `0x1111...` without any code? 
- In general, it's better to not link externally too much, but try to copy-paste relevant snippets so the reader doesn't have to jump away from the text. 
- It becomes a bit confusing when two tests are assembled at the same time. Maybe better to do the first, then expand it with a brief section to also cover the second test. 


> To see these two test cases tested, run these commands: ...

That's jumping ahead, and breaking the narrative. Let's skip that for later, and concentrate on how to produce the test first. 

I would write like this: 

// Start of my example

The first testcase just excecutes a snippet of code

> This should jump into a subroutine, back out and stop.
> Bytecode: 0x60045e005c5d (PUSH1 0x04, JUMPSUB, STOP, BEGINSUB, RETURNSUB)


The test produces no output and the only state change they cause is a cost in gas. Executing this code will cost `18` gas if the operations executed correctly, and something else if they did not. Therefore, we can rely on the gas cost to validate execution. 

To execute the code, we need to place it in the pre-state, for example:

```
0x1111111111111111111111111111111111111111:
  balance: 0
  nonce: 0
  storage: {}
  code: :raw 0x6800000000000000000c5e005c60115e5d5c5d
  
```

We also need an EOA account to call into the destination that has the code: 

```
0xa94f5374Fce5edBC8E2a8697C15331677e6EbF0B:
  balance: 1000000000000000000
  nonce: 0
  code: ''
  storage: {}

```

/// END of my example

qbzzt:
Thank you @holiman , I think I fixed the issues in the latest commit. I made sure to tell people to only do one test case at a time.

winsvega:
@holgerd77 is it autoupdated for read the docs?

holgerd77:
@winsvega you can always go to https://readthedocs.org/projects/ethereum-tests/ and see what's going on, you are added as a developer there (let me know if you have problems to log in for any reasons). @qbzzt I can add you there as well if you want.

If a build is not triggered automatically, it is always possible there to re-trigger the build manually. In the current case the build (from 5 days ago, that should be it, right?) is showing there as "failed" under the "Builds" tab. Reason is a bit unclear, I just re-triggered for now, we'll see if it passes through.

qbzzt:
It works now, thank you.

If nobody minds, I think it will be useful to make me a developer on
readthedocs.org . Dimitry, is that OK?

holgerd77:
@qbzzt I would very much assume that @winsvega is ok with that. Can you register on ReadTheDocs or sign up with GitHub (if you haven't already) and then give me your profile name from there? Thanks! 😄

winsvega:
I am ok. 
That depends if ori find an agreement with Joseph and sign a contract to work on the tests regularly

PR #725, opened 2020-09-26
Added the retesteth reference

qbzzt:


qbzzt:
This also includes by bad opcodes test. I'm not good enough with github to just have the reference here.

qbzzt:
Subsumed in a different pull request

PR #726, opened 2020-09-26
Three changes:

qbzzt:
1. opcodeTest
2. retesteth reference
3. the new and simplified --vmtrace

### DIFF
--- a/src/GeneralStateTestsFiller/stCreateTest/ResultOfCREATE_OPCODE_SolidityFiller.yml
+++ b/src/GeneralStateTestsFiller/stCreateTest/ResultOfCREATE_OPCODE_SolidityFiller.yml
@@ -8,10 +8,10 @@ ResultOfCREATE_OPCODE_Solidity:
 
   env:
     currentCoinbase: 2adc25665018aa1fe0e6bc666dac8fc2697ff9ba
-    currentDifficulty: 0x20000
-    currentGasLimit: 100000000
+    currentDifficulty: '0x20000'
+    currentGasLimit: '100000000'
     currentNumber: 1
-    currentTimestamp: 1000
+    currentTimestamp: '1000'
     previousHash: 5e20a0453cecd065ea59c37ac63e079ee08998b6045136a8ce6635c7912ec0b6
 
   _info:
@@ -103,7 +103,7 @@ ResultOfCREATE_OPCODE_Solidity:
     cccccccccccccccccccccccccccccccccccccccc:
       balance: '0x0ba1a9ce0ba1a9ce'
       code: ':solidity Creator'
-      nonce: 0
+      nonce: '0'
       storage: {}
 
     a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
@@ -116,13 +116,13 @@ ResultOfCREATE_OPCODE_Solidity:
     data:
     - :abi createChildren()
     gasLimit:
-    - 80000000
-    - 302000  # Run out of gas in the first CREATE
+    - '80000000'
+    - '302000'  # Run out of gas in the first CREATE
     gasPrice: '1'
     nonce: '0'
     to:     cccccccccccccccccccccccccccccccccccccccc
     value:
-    - 0
+    - '0'
     secretKey: 45a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065ff2d8
     
     
@@ -136,23 +136,23 @@ ResultOfCREATE_OPCODE_Solidity:
       result:
         cccccccccccccccccccccccccccccccccccccccc:
           storage:
-            0x00: 0x01     # Two different addresses for two different 
+            '0x00': '0x01'     # Two different addresses for two different 
                            # deployments of the contract
-            0x01: 0x00     # No return data for contract creation
-            0x02: 0xce     # Code size of the Created contract
+            '0x01': '0x00'     # No return data for contract creation
+            '0x02': '0xce'     # Code size of the Created contract
 
             # Normal CREATE
 
             # Result after CALL
-            0x10: 0xFF00FF00FF00FF00   # Value from Created contract
-            0x11: 0x20     # Return data for Created.getValue() is 256 bits
+            '0x10': '0xFF00FF00FF00FF00'   # Value from Created contract
+            '0x11': '0x20'     # Return data for Created.getValue() is 256 bits
 
             # DELEGATECALL
-            0x20: 0xbad   
+            '0x20': '0xbad'   
 
             # Result after STATICCALL
-            0x30: 0xFF00FF00FF00FF00   # Value from Created contract
-            0x31: 0x20     # Return data for Created.getValue() is 256 bits
+            '0x30': '0xFF00FF00FF00FF00'   # Value from Created contract
+            '0x31': '0x20'     # Return data for Created.getValue() is 256 bits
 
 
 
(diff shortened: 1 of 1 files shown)


docs/config-dir.rst
```diff
@@ -0,0 +1,96 @@
...
+#################################
+
+The retesteth **config** directory contains the **retesteth** configuration. If it is
+empty **retesteth** creates one with the default values. Every directory under it 
+contains either the default configuration, or configuration for a specific 
+client (to override the default for that client).
+
+These directories can contain this information:
+
+- **genesis/<forkname>.json**, this is the genesis config for the client, primarily
```
winsvega: note that `forkname` is as defined in the tests

docs/config-dir.rst
```diff
@@ -0,0 +1,96 @@
...
+  because different clients refer to the forks by different names. This file may
+  also contain an **accounts** field. This is legacy and can be ignored.
+
+- **genesis/correctMiningReward.json**, a file that includes the mining reward for
+  each fork. 
+
+- **start.sh**  the meaning of this script varies depending on the
+  method used to communicate with the client. 
+
+  - With **tcp** and **ipc** clients the script
```
winsvega: note "see existing aleth ipc / geth tcp configuration"

docs/config-dir.rst
```diff
@@ -0,0 +1,96 @@
...
+    starts the client and possibly provides it with the port or pipe on 
+    which it should listen. In both cases it is possible to start multiple clients 
+    to run tests in parallel.
+
+  - With **ipc-debug** clients the script is ignored, because it is assumed that the
```
winsvega: the script is also ignored when there is no script. if there is script, retesteth would call it. 
if there is no script. you can start the client manually

docs/config-dir.rst
```diff
@@ -0,0 +1,96 @@
...
+    that the client is already running in debug mode.
+
+  - With **transition-tool** clients the script is executed for every test, and the
```
winsvega: it is also set in config file where to look for this script. in socketAddress field. 
for tcp clients the socket addres contains sockets. 
for t8ntool contains the script path which will run the tool. 
for ipc also the script that will start the instance of a client

but really note "see existing configurations"  it's quite clear.

docs/config-dir.rst
```diff
@@ -0,0 +1,96 @@
...
+    But not additional forks such as **ByzantiumToConstantinopleFixAt5**.
+    
+  - **exceptions**, the exception messages that the client emits for blocks that
+    are invalid in various ways. The key is the string used to identify the exception
+    in the **exceptException** field of invalid block tests. The value is the message
+    the client emits.
+
+    .. note::
+
+       The exception is only checked if the **-\\-filltests** is specified. Without it
```
winsvega: expect exception is used by invalid blockchain tests

docs/retesteth-ref.rst
```diff
@@ -0,0 +1,287 @@
...
+   suite and all the other options.
+   
+
+Set The Suite
+-----------------------
+========================================= ===================================
+Option                                    Meaning
+========================================= ===================================
+-t <TestSuite>                            Run all the tests in that suite
+-t <TestSuite>/<TestCase>                 Run a specific test case
```
winsvega: test case folder. the test case sounds like a single test. but actually testCase here is a whole folder. and in case of blockchains it has 3 levels

docs/retesteth-ref.rst
```diff
@@ -0,0 +1,287 @@
...
+Setting the Test Suite and Test
+----------------------------------
+========================================= ===================================================
+Option                                    Meaning
+========================================= ===================================================
+-\\-testpath <PathToTheTestRepo>          Set path to the test repo
+-\\-testfile <TestFile>                   Run tests from a file. Requires -t <TestSuite>
+-\\-singletest <TestName>                 Run on a single test. `Testname` is filename 
+                                          without Filler.json
+-\\-singletest <TestName>/<Subtest>       `Subtest` is a test name inside the file
```
winsvega: note:  subtest is used in blockchain tests that has many test cases in one file

docs/retesteth-ref.rst
```diff
@@ -0,0 +1,287 @@
...
+============================= ===================================================
+-d <index>                    Set the transaction data array index when running 
+                              GeneralStateTests
+-g <index>                    Set the transaction gas array index when running 
+                              GeneralStateTests
+-v <index>                    Set the transaction value array index when running 
+                              GeneralStateTests
+-\\-vmtraceraw                Trace transaction execution
+-\\-vmtrace                   Trace transaction execution, simplified version
+-\\-limitblocks               Limit the block exectuion in blockchain tests for
```
winsvega: the option has an argument as well

docs/retesteth-ref.rst
```diff
@@ -0,0 +1,287 @@
...
+   filled test files. However, you can also use the test source code
+   (a.k.a. the filler version).
+
+   Fill (and run) a test that is part of the test suite (in this case,
+   **tests/GeneralStateTests/stExample/add11**, whose source code is
+   **tests/src/GeneralStateTestsFiller/stExample/add11Filler.json**):
+
+   ::
+ 
+      ./dretesteth.sh -t GeneralStateTests/stExample -- \
```
winsvega, line 260: here .json would give an error. should be just add11

docs/retesteth-ref.rst
```diff
@@ -0,0 +1,287 @@
...
+      ./dretesteth.sh -t GeneralStateTests -- --testpath ~/tests --filltests \
+         --testfile /tests/docs/tutorial_samples/04_multitestFiller.yml -d 1
+
+#. Run a test and produce `a trace of the Ethererum Virtual Machine:
+   <https://ethereum-tests.readthedocs.io/en/latest/internals-tutorial.html#virtual-machine-trace>`_:
+
+   ::
+
+      ./dretesteth.sh -t GeneralStateTests/stExample -- \
+           --testpath ~/tests --vmtraceraw
```
winsvega, line 311: or --vmtrace

winsvega (commented):
the solidity test we can have. but perhaps need to mention it in the tutorial 
that solidity insert a lot of opcodes that make debug trace very messy and hard to get whats going on in the test. 

solidity is ok to use when you just want to execute plain byte sequence with asm code from a conract. 

but when there is a tricky scenario with create collisions for example and the returndatasize behavior it is safer to use just lllc.

winsvega:
the build is red because python script can't handle the yml format you provided. check that all string fields are string and that all int fields are int as in .json format.

qbzzt:
> the solidity test we can have. but perhaps need to mention it in the tutorial
> that solidity insert a lot of opcodes that make debug trace very messy and hard to get whats going on in the test.
> 
> solidity is ok to use when you just want to execute plain byte sequence with asm code from a conract.
> 
> but when there is a tricky scenario with create collisions for example and the returndatasize behavior it is safer to use just lllc.

I added a note to that effect. It will be in the next commit.

docs/retesteth-ref.rst
qbzzt: Added

docs/config-dir.rst
qbzzt: I think telling people to see the existing documentation once, in the config (which I moved to be above start.sh), is enough.

PR #723, opened 2020-09-18
update gtests for Berlin/Yolov2 WIP

winsvega:
regenerate all the tests using geth t8n tool with yolov2 support. 
* update expect sections in tests where its needed. 
* review tests, improve expect sections
* review :raw opcode places

work in progress

PR #727, opened 2020-09-27
Berlin zk

shemnon:
All of these are gas reductions to ensure they OOG in the berlin gas
configuration.


winsvega:
thanks, I included some of the changes in https://github.com/ethereum/tests/pull/723
please note me if you are making any changes like this so we don't do double work. 

did you use retesteth configuration to run the filltest comand  with  besuRPC ? 
we recently add documentation about this. please share your experience implementing a new fork and enabling it in the configs

PR #728, opened 2020-10-02
The Retesteth reference, including the config directory.

qbzzt:
As requested, just the documentation.

### DIFF
--- a/docs/config-dir.rst
+++ b/docs/config-dir.rst
@@ -106,6 +106,13 @@ These directories can contain this information:
 
 - **stop.sh** stop the client.
 
+- In the case of **transition-tool** clients, this directory also contains the 
+  script that runs the client for each test. This script's name is specified in the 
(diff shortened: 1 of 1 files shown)


docs/config-dir.rst
```diff
@@ -0,0 +1,121 @@
...
+
+    .. note::
+
+       If there is no **start.sh** script at all **retesteth** assumes that it 
+       needs to connect to an existing client rather than run its own.
+
+  - With **ipc-debug** clients the script is ignored, because it is assumed that the
+    that the client is already running in debug mode.
+
+  - With **transition-tool** clients the script is executed for every test, and the
```
winsvega, line 103: not the start.sh 
the script which is defined in socketAddress in a config file

qbzzt: Fixed, thank you for noticing.

PR #733, opened 2020-10-06
Made the test work without --filltests

qbzzt:
1. Changed from state test to blockchain test
2. Split the test into 256 (one per opcode) so the failure will tell us what opcode is bad

It now works correctly on geth.


qbzzt:
Use PR #736 instead.

PR #735, opened 2020-10-10
Test for the result of creating a new contract:

qbzzt:
1. CREATE / CREATE2 the contract
2. Verify RETURNDATASIZE = 0
3. Verify EXTCODESIZE and EXTCODECOPY give the expected results (the contract we created)
4. CALL / CALLCODE / DELEGATECALL / STATICCALL the contract
5. Verify the call returned 1 (0 would mean failure, revert, etc.)


src/GeneralStateTestsFiller/stCreateTest/CreateResultsFiller.yml
```diff
@@ -0,0 +1,171 @@
...
+        - !!int -1
+        gas:  !!int -1
+        value: !!int -1
+      network:
+        - '>=Istanbul'
+      result:
+        cccccccccccccccccccccccccccccccccccccccc:
+          storage: {}
+
+            # Zero if all goes well
```
winsvega, line 341: put all the storage in pre section to 12 for example. so that if all goes well. 12 is overwritten with 0.
in case tx fails the storage would be 0 too

src/GeneralStateTestsFiller/stCreateTest/CreateResultsFiller.yml
```diff
@@ -0,0 +1,171 @@
...
+          ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+          ; EXTCODE should match what we returned in the constructor
+          ;
+
+          [extLength] (extcodesize @addr1)
+          (extcodecopy @addr1 extCode 0 @extLength)
+
+          [[0x11]] (- @contractLength @extLength)
+          [[0x12]] (- @contractCode   @extCode)
+
```
winsvega, line 159: check returndatasize/returndatacopy/extcodesize/extcodecopy/pc

src/GeneralStateTestsFiller/stCreateTest/CreateResultsFiller.yml
```diff
@@ -0,0 +1,171 @@
...
+          ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+          ; CALL the created contract
+
+          (if (= @callType 1)
+             [callRet] (call 0xFFFF @addr1 0 0 0 0 0)
```
winsvega, line 164: put some code in contract like 
sstore [100] = 12 
so that you know that created contract got executed
(I do not suggest to touch the memory in created contract code and return because it will effect the retrundatasize/returndatacopy)

winsvega (changes requested):
comments

src/GeneralStateTestsFiller/stCreateTest/CreateResultsFiller.yml
qbzzt: I am checking extcodesize and extcodecopy (lines 78-88):

          [[0x10]] (returndatasize)
          [extLength] (extcodesize @addr1)
          (extcodecopy @addr1 extCode 0 @extLength)
          [[0x11]] (- @contractLength @extLength)
          [[0x12]] (- @contractCode   @extCode)

If storage[0x10-0x12] are not all zeros, it means an error in one of those values. Since returndatasize is zero, what would be the point of returndatacopy ?

What value of pc (the program counter) do you want me to check?

src/GeneralStateTestsFiller/stCreateTest/CreateResultsFiller.yml
qbzzt: The problem with just modifying storage is that I don't necessarily know the contract address (it's consistent right now, even with CREATE, but I don't think that's promised). So instead I'm going to call a contract with a fixed address from the created contract.

qbzzt:
**create/create2 - > ogg during create - > check opcodes -> call/delegate/static -> see results**

Can we do out of gas in a GeneralStateTest, or does it need to be in a BlockchainTest? Should we move the entire test, or just this particular subtest?

**create/create2 - > collision during create - > check opcodes -> call/delegate/static -> see results**

1. How do I create a collision in CREATE?

2. For CREATE2, I don't see the point in detecting a collision and verifying that the created contract can be called four ways, it would be less efficient. How about I create a separate test that verifies:
    A. We can CREATE2 two identical contracts with different salts
    B. We can CREATE2 two different contracts with the same salt
    C. If we attempt to CREATE2 two identical contracts with the same salt, it fails
    D. If a contract SELFDESTRUCTs, we can then CREATE2 an identical contract with the same salt and it works

winsvega:
for collision you trace the create result on an empty state to see which address was created. 
then you copy that address in to a pre state in the filler.
chage the code. 

after collision you call the contract code to see which code was executed. the original one or the one that could potentially override. 

regardless of the override we check the values of returndatasize/returndatacopy/extcodesize/extcodecopy
this I see can potentially detect errors in those opcode implementation and collision/oog behavior 

**==>actually one more case can be here. create might have a revert instruction in its constructor.**

src/GeneralStateTestsFiller/stCreateTest/CreateCollisionResultsFiller.yml
```diff
@@ -0,0 +1,188 @@
...
+          ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+          ; Create the contract and a constructor to pass to CREATE[2]
+          ;
+
+          [constructorLength] 
+            (lll {   ; successful constructor
+               (return contractCode
+                  ; write the contract to constructor memory 
+                  (lll
+                     (sstore 0 0xFF)
```
winsvega: is this constructor code?

src/GeneralStateTestsFiller/stCreateTest/CreateCollisionResultsFiller.yml
```diff
@@ -0,0 +1,188 @@
...
+                     contractCode
+                  ) ; inner lll
+               )  } ; return contractCode (lll ...)
+                constructorCode
+            )       ; outer lll, if successful
+
+          [contractLength]
+            (lll
+               (sstore 0 0xFF)
```
winsvega: is this the contract actual code

src/GeneralStateTestsFiller/stCreateTest/CreateCollisionResultsFiller.yml
```diff
@@ -0,0 +1,188 @@
...
+          (if (= @creation 1)
+            [addr1] (create 0 constructorCode @constructorLength)
+            [addr1] (create2 0 constructorCode @constructorLength 0x5a17)
+          )
+
+          ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+          ; RETURNDATASIZE should be zero
+          ;
+
+          [[0x10]] (returndatasize)
```
winsvega, line 110: + returndatacopy, extcodesize, extcodecopy and pc

src/GeneralStateTestsFiller/stCreateTest/CreateCollisionResultsFiller.yml
```diff
@@ -0,0 +1,188 @@
...
+          [[0x11]] @addr1
+
+          ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+          ; CALL the address where the contract should have been created
+
+          [callRet] (call 0xFFFF 
+                         0x8af6a7af30d840ba137e8f3f34d54cfb8beba6e2 0 0 0 0 0)
+
+          [[0x12]] (- @callRet 1)
+          [[0x13]] (returndatasize)
```
winsvega, line 123: same returndatacopy, extcodesize, extcodecopy and pc

src/GeneralStateTestsFiller/stCreateTest/CreateCollisionResultsFiller.yml
```diff
@@ -0,0 +1,188 @@
...
+
+        cccccccccccccccccccccccccccccccccccccccc:
+          storage: {}
+            # In the case of a collision CREATE[2] just pops the 
+            # parameter values from the stack and returns zero, so the
+            # results should all be zero
+
+        # The address CREATE would have chosen
+        8af6a7af30d840ba137e8f3f34d54cfb8beba6e2:
+          storage:
```
winsvega, line 221: can also check the code field

src/GeneralStateTestsFiller/stCreateTest/CreateCollisionResultsFiller.yml
```diff
@@ -0,0 +1,188 @@
...
+            0: 0x001D
+            # 0x60A7 - no contract ran
+            # 0x001D - the old contract ran
+            # 0x00FF - the new contract ran
+
+        # The address CREATE2 would have chosen
+        40f1299359ea754ac29eb2662a1900752bf8275f:
+          storage:
```
winsvega, line 230: can also check the code field

src/GeneralStateTestsFiller/stCreateTest/CreateResultsFiller.yml
```diff
@@ -0,0 +1,327 @@
...
+                 )  } ; return contractCode (lll ...)
+                  constructorCode
+              )       ; outer lll, revert
+              NOP
+            )
+
+
+          ; Write the contract to our own memory (as the 0xcc...cc contract)
+          [contractLength] (lll 
+             (call 0xFFFF 0x60A7 0 0 0 0 0)
```
winsvega: what is this call  ?

src/GeneralStateTestsFiller/stCreateTest/CreateCollisionResultsFiller.yml
qbzzt: Yes.

winsvega (changes requested):

winsvega:
Looks a little bit too heavy. I think the test can be cleaner to read. 
also can check pc. 
in lllc you can use labels in the code as subroutines

src/GeneralStateTestsFiller/stCreateTest/CreateCollisionResultsFiller.yml
qbzzt: No. What happens is that we need the contract code in two places with different memories:

1. The constructor, to create the new contract
2. The LLL contract, to verify that EXTCODECOPY returns the same values as the ones we sent

The inner lll inside the constructor creates the code in the constructor memory. This lll creates the code in our contract's memory.

src/GeneralStateTestsFiller/stCreateTest/CreateCollisionResultsFiller.yml
qbzzt: 1. When RETURNDATASIZE is zero do you still need me to have RETURNDATACOPY?
2. I already checked EXTCODE in the regular create test. Do you want me to also check it here?
3. What do you want me to check with PC?

src/GeneralStateTestsFiller/stCreateTest/CreateCollisionResultsFiller.yml
qbzzt: I didn't know the expect: field can include code. Still, this method works so I think we might as well keep it.

src/GeneralStateTestsFiller/stCreateTest/CreateResultsFiller.yml
qbzzt: Same as above:

No. What happens is that we need the contract code in two places with different memories:

The constructor, to create the new contract
The LLL contract, to verify that EXTCODECOPY returns the same values as the ones we sent
The inner lll inside the constructor creates the code in the constructor memory. This lll creates the code in our contract's memory.

src/GeneralStateTestsFiller/stCreateTest/CreateCollisionResultsFiller.yml
winsvega: Yes. Check everything regardles of the result of other checks

Extcode opcodes also to be check in all cases with create (success/oog in constructor/revert in constructor/collision)

Pc just put into storage

src/GeneralStateTestsFiller/stCreateTest/CreateCollisionResultsFiller.yml
winsvega: The expect can include all fields of the account. Or just sone of them.
Did you describe it in the docs?

src/GeneralStateTestsFiller/stCreateTest/CreateResultsFiller.yml
winsvega: But why is the call in constructor?

qbzzt: It isn't in the constructor. It's in the code of the contract being constructed. Anyway, that will be a lot clearer in the next version, where it's all hidden in a macro.

src/GeneralStateTestsFiller/stCreateTest/CreateCollisionResultsFiller.yml
qbzzt: Done

src/GeneralStateTestsFiller/stCreateTest/CreateCollisionResultsFiller.yml
qbzzt: As of when you'll accept the pull request, yes.

src/GeneralStateTestsFiller/stCreateTest/CreateResultsFiller.yml
```diff
@@ -42,38 +43,34 @@ CreateResults:
           [callType]       (and (shr $ 0 240) 0xFF)
           [constructor]    (and (shr $ 0 232) 0xFF)
 
+          ; The contract code          
+          (def 'contractMacro
+                    (lll
+                       (call 0xFFFF 0x60A7 0 0 0 0 0)
```
winsvega, line 45: can you explain a logic around. here.  
a contract creating happens. then a created contract has this call 
as a body (call 0xFFFF 0x60A7 0 0 0 0 0).

qbzzt: I added comments to explain it:

I did not want to rely on knowing the address at which the contract will get created, because the algorithm might change in the future (for CREATE, anyway - CREATE2 is supposed to give fixed results). So to know that the created contract has been called successfully it calls the 0x60a7 account, and that account changes its storage in a way that the expect: section can identify.

qbzzt:
Replaced by pr 746

PR #736, opened 2020-10-10
Fixed testOpcodes

qbzzt:
1. Moved testOpcodes to BlockchainTests/ValidBlocks
2. Split it into 256 individual tests to identify the bad opcode without --filltests
3. Removed some sanity checks that claimed we violate a schema when we don't


winsvega (changes requested):
when you put bytecode in tx data. it should also have :raw prefix

winsvega:
the test works great. just fix the :raw in tx data to silent the warning

PR #737, opened 2020-10-10
Syntax for multiple index values in one expect: member

qbzzt:
Added the syntax to have multiple index values (for data, gas, or value) in the same expect list member.

### DIFF
--- a/docs/state-transition-tutorial.rst
+++ b/docs/state-transition-tutorial.rst
@@ -1,4 +1,4 @@
-.. _state_transition_tutorial:
+_.. _state_transition_tutorial:
 
 ###########################################
 State Transition Tests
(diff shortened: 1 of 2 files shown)


holgerd77:
@qbzzt Thanks, great that there is continued work on the docs! 😄

docs/state-transition-tutorial.rst
```diff
@@ -420,13 +421,17 @@ There are two steps to doing that:
                0:    0x10
                0x10: 0x10
 
-  This is the second item in the **data:** subsection above, 
-  equivalent to **val2Storage(0x11, 0x11)**.
+  This is for the second and fourth items in the **data:** subsection above. 
+  When you have multiple values
+  that produce the same test results, you can specify **data**, **gas**, or **value**
+  as a list instead of a single index.
```
winsvega, line 427: and -1 means all

winsvega (approved):

PR #738, opened 2020-10-17
Tests for the gas cost of each opcode

qbzzt:


src/GeneralStateTestsFiller/stEIP150singleCodeGasPrices/gasCostExpFiller.yml
```diff
@@ -0,0 +1,100 @@
...
+          ; of the parameters
+          [powerOf]       $4
+          [expectedCost]  $36
+
+ 
+          ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+          ; Run the operation
+          [gasB4]    (gas)
+          (exp 2 @powerOf)
+          [gasAfter] (gas)
```
winsvega, line 42: here the cost of (GAS) opcode itself gona be recorded. 
you can put this as a constant in a contract and to the formula.

winsvega (approved):
looks very good. I think we can remove all the previous tests in this folder. because yours cover everything

src/GeneralStateTestsFiller/stEIP150singleCodeGasPrices/gasCostExpFiller.yml
qbzzt: That is the reason what I check is the total expected cost.

qbzzt:
> looks very good. I think we can remove all the previous tests in this folder. because yours cover everything
> (not in this pr)

Not quite. I covered all the opcodes that were either:

1. Easy

or 

2. Previously uncovered

So we should preserve at least these tests:

0x3C EXTCODECOPY     stEIP150singleCodeGasPrices/RawExtCodeCopyGasFiller.json
0xF1 CALL     stEIP150singleCodeGasPrices/RawCallGasAskFiller.json
0xF0 CREATE     stEIP150singleCodeGasPrices/RawCreateGasFiller.json
0xF2 CALLCODE    stEIP150singleCodeGasPrices/RawCallCodeGasMemoryFiller.json
0xF3 RETURN     stEIP150singleCodeGasPrices/gasCostReturn.yml
0xF4 DELEGATECALL     stEIP150singleCodeGasPrices/RawDelegateCallGasAskFiller.json

qbzzt:
Superceded by PR #745

PR #757, opened 2020-11-16
Tests for most of EIP2929

qbzzt:
1. The first time in a transaction that a storage cell or an address is referenced,
   the action is more expensive (compared to Istanbul, which is pre EIP2929)
2. Subsequent access is cheaper (again, compared to EIP2929)

This applies to:

Storage access:
SLOAD and SSTORE

Address access:
BALANCE
EXTCODESIZE
EXTCODECOPY
EXTCODEHASH
CALL
CALLCODE
DELEGATECALL
STATICCALL

### DIFF
--- a/GeneralStateTests/stEIP150singleCodeGasPrices/eip2929-ff.json
+++ b/GeneralStateTests/stEIP150singleCodeGasPrices/eip2929-ff.json
@@ -1,0 +1,241 @@
+{
+    "eip2929-ff" : {
+        "_info" : {
+            "comment" : "Ori Pomerantz qbzzt1@gmail.com",
+            "filling-rpc-server" : "evm version 1.9.25-unstable-6b985808-20201117",
+            "filling-tool-version" : "retesteth-0.0.9-berlin+commit.bdf34dce.Linux.g++",
+            "lllcversion" : "Version: 0.5.14-develop.2020.11.17+commit.9189ad7a.Linux.g++",
+            "source" : "src/GeneralStateTestsFiller/stEIP150singleCodeGasPrices/eip2929-ffFiller.yml",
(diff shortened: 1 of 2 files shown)


src/GeneralStateTestsFiller/stEIP150singleCodeGasPrices/eip2929Filler.yml
```diff
@@ -0,0 +1,848 @@
...
+            0x00: 0
+
+
+    ######### STORAGE ################################################
+
+
+    # SLOAD three times
+    - indexes:
+        data:
+        - :abi f(uint,uint,uint) 1 1 1
```
winsvega: Use ":label foo"

winsvega (changes requested):
Plus put a test for selfdestruct

src/GeneralStateTestsFiller/stEIP150singleCodeGasPrices/eip2929Filler.yml
qbzzt: Done

qbzzt:
> Plus put a test for selfdestruct

Done, `eip2929-ff` (because I didn't want to write a long name such as `eip2929-self-destruct`. It checks all the ways in which an address can become accessed, and also that the selfdestruct itself makes the address accessed if it wasn't before.

winsvega (changes requested):
Ok. Now need a retesteth folder update. And --filltests --fillchain

qbzzt:
I already did the --filltests for this test.

I will do the doc updates for the retesteth changes soon, probably over the
weekend.

winsvega:
I see. Checkhash doesn't detect new tests missing in blockchain format

winsvega:
moved to https://github.com/ethereum/tests/pull/760

PR #758, opened 2020-11-19
mention Nimbus-eth1 next to other clients

jangko:


winsvega (commented):
This whole reference needs an update. 

Why remove other clients?

jangko:
> Why remove other clients?

I removed nothing

winsvega:
Sorry from the app it looked like red lines.

winsvega (approved):

PR #759, opened 2020-11-21
Updated the documentation

qbzzt:
1. Fixed inaccuracies created by the fact state and blockchain used the
   same transaction and expect sections.

2. Added the changed in **expect.indexes**:
   * Range support
   * Label support


winsvega (approved):

winsvega:
ready?

qbzzt:
Yes. No update to retesteth because the tutorial already has build
instructions for the docker

PR #761, opened 2020-11-23
Buffer tests

qbzzt:
For every opcode with a buffer, check:

1. That it is successful with buffer length 10
2. That it is successful with buffer length 0
3. That it fails with buffer length -10

Try this with different offsets, etc.

### DIFF
--- a/src/GeneralStateTestsFiller/stMemoryTest/bufferFiller.yml
+++ b/src/GeneralStateTestsFiller/stMemoryTest/bufferFiller.yml
@@ -69,9 +69,20 @@ buffer:
            (def 'length     0x2020)
            (def 'offset     0x2040)
          
-           ; bufferType 0 is normal, 1 is length zero, 2 is negative length
+           ; bufferTypes  0 is normal, 1 is length zero, 2 is negative length
            ; bufferType 3 is excessively long, for opcodes with bounds checking
            ; Add 0 for offset 0x100, 10 for offset 0x0
+
+           ; High offsets:
+           ; 20 for 2^256-10
+           ; 21 for 2^31-1
+           ; 22 for 2^31
+           ; 23 for 2^32-1
+           ; 24 for 2^32
+           ; 25 for 2^63-1
+           ; 26 for 2^63
+           ; 27 for 2^64-1
+           ; 28 for 2^64
            (if (= bufferType 0) {
                    [length] 10
                    [offset] 0x100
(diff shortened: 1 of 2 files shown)


holgerd77:
Was just commenting on the topic "VM tests vs State Tests" in a discussion with @winsvega here https://github.com/ethereumjs/ethereumjs-vm/issues/954#issuecomment-732827621, just as a note here since this fits: I think these kind of tests are generally a good example for test cases which would also fit well being implemented as VM tests since this would allow for more targeted debugging on the VM side.

Generally it's great though that these kind of tests are added (at all) - thanks @qbzzt - these are exactly the implementation nits where client developers (including us) often stumble upon.

holgerd77:
As a follow-up: so - the more I think about that - I have to say that I do like the VM tests as well. 😄

qbzzt:
> Generally it's great though that these kind of tests are added (at all) - thanks @qbzzt - these are exactly the 
> implementation nits where client developers (including us) often stumble upon.

Thank you. This is very good to hear, considering you're our customers. The whole point is to help you develop clients.

src/GeneralStateTestsFiller/stMemoryTest/bufferFiller.yml
```diff
@@ -0,0 +1,442 @@
...
+    a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+      balance: '0x0ba1a9ce0ba1a9ce'
+      code: '0x'
+      nonce: '0'
+      storage: {}
+      
+  transaction:
+    data:
+    # SHA3       
+    - :label ok   :abi f(uint,uint) 0x20 0   # Valid buffer length
```
winsvega: what is the meaning of 0 1 2 10 11 12 12 ?

qbzzt: Sorry, the last one was supposed to be 13. I fixed the comment.

src/GeneralStateTestsFiller/stMemoryTest/bufferFiller.yml
```diff
@@ -220,6 +304,15 @@ buffer:
     - :label ok   :abi f(uint,uint) 0x3c 10   # Valid buffer length
     - :label ok   :abi f(uint,uint) 0x3c 11   # Buffer length zero
     - :label fail :abi f(uint,uint) 0x3c 12   # Negative buffer length
+    - :label fail :abi f(uint,uint) 0x3c 20
```
winsvega, line 307: i think this need coments too.  20/21/22 the meaning

winsvega (approved):

PR #764, opened 2020-11-27
The Solidity self destruct tests, as requested in https://github.com/…

qbzzt:
…ethereum/tests/issues/431

Note that selfdestruct(<contract's own address>) burns the balance.


winsvega:
have you traced with --vmtrace that execution goes well ?

qbzzt:
> have you traced with --vmtrace that execution goes well ?

Now I have. I didn't go opcode by opcode, but I used grep to see the `SELFDESTRUCT` opcodes and every test case has two of them with the correct address in the top of the stack:


[selfdestructs.txt](https://github.com/ethereum/tests/files/5610930/selfdestructs.txt)

winsvega (approved):

PR #766, opened 2020-11-30
When an opcode receives an address > 2^160, what happens?

qbzzt:
Ensure the least significant 160 bits are used and the rest are truncated


winsvega (changes requested):
1)

```
(call 0x1000 @addr1 0 0x2000 0x20 res1 word)
             (call 0x1000 @addr2 0 0x2000 0x20 res2 word)
```

would feel safer to have  
res1 == res2 == "expected result"

2)
please double check with vmtrace that the stack for opcodes does actually contain a value with (2**160 + adress)  for the second case.  I would not blindly trust the lll compiler to do that.

qbzzt:
> please double check with vmtrace that the stack for opcodes does actually contain a value with (2**160 + adress) for the second case. I would not blindly trust the lll compiler to do that.

Done. The values for `addr2` are the expected ones.

qbzzt:
> would feel safer to have
> res1 == res2 == "expected result"

Then you can feel safe now ;-)

winsvega (approved):

PR #769, opened 2020-12-08
Added a warning about uncle headers

qbzzt:


docs/test_filler/test_blocks.rst
```diff
@@ -98,6 +98,10 @@ The fields in each block are optional. Only include those fields you need.
   - **populateFromBlock**: The block number within that chain for the block
     that is an uncle of the block you are specifying.
 
+  However, if you write a test with uncles, you need to run it twice, once
```
winsvega, line 101: this is because retesteth is unable to calculate the post hash as it is not a client.
and retesteth is constructing block with uncle. (because there is no protocol to ask a test generating client to construct block with uncles.

PR #771, opened 2021-01-16
vmArithmetic tests made into GeneralStateTests

qbzzt:


holgerd77:
Nice. 😄

winsvega:
This a great start to review all the VM tests that we had unmaintained for a long time. 
not only we put them in the same format as other tests but add more edge cases.

winsvega:
readt to merge first part ?

qbzzt:
> readt to merge first part ?

Yes. It started as vmArithmetic, but now it has a lot of other ones too. The ones that are here are ready.

Well, kind of ready. The unready part is I don't have the filled versions in the repository, but that will be easiest to do after you add the test suites and I can just run them in place.

winsvega:
new tests are supported by new retesteth, currently in develop branch

PR #772, opened 2021-01-23
access list test example [example]

winsvega:
Retesteth: https://github.com/ethereum/retesteth/pull/114
Geth: https://github.com/ethereum/go-ethereum/pull/21502


holiman:
Reposting this here: I think the `"data" : ":label somelabel :raw 0x1122"`, is really not a good idea. I think it's fine to have 
```
 "data" : "0x..",
 "_dataCommment": "Foo bar bazonk",
 "accessList" : [ ...] 
"_anotherComment": "Baz bar",
```

But inserting logic into the data-field is really making things too complex, and doesn't play well with json-parsers: e.g. in golang, our parsing is typed. As in, we define what the type of a specific input is, and having some mixed, loosely defined meta-format with :-delimited fields just adds another level of quirkyness

GeneralStateTests/stExample/accessList.json
```diff
@@ -0,0 +1,89 @@
...
+                }
+            }
+        },
+        "transaction" : {
+            "data" : [
+                "0x1122",
+                "0x3344"
+            ],
+            "accessLists" : {
+                "0" : [
```
MariusVanDerWijden: I don't really know how to unmarshal this properly as the key depends on the length of data. I would suggest to drop the index and just make `accessLists` an array like data with the same length and leave elements empty if you don't want to have an access list for it. cc @winsvega @holiman 

e.g.
```go
type stTransaction struct {
	GasPrice    *big.Int       `json:"gasPrice"`
	Nonce       uint64         `json:"nonce"`
	To          string         `json:"to"`
	Data        string         `json:"data"`
	AccessLists []types.AccessList `json:"accessLists,omitempty"`
	GasLimit    []uint64       `json:"gasLimit"`
	Value       []string       `json:"value"`
	PrivateKey  []byte         `json:"secretKey"`
}

type AccessList []AccessTuple  `json:"omitempty"`

type AccessTuple struct {
	Address     *common.Address `json:"address"        gencodec:"required"`
	StorageKeys []*common.Hash  `json:"storageKeys"    gencodec:"required"`
}
```

MariusVanDerWijden: So basically my idea would be something like this:
```json
"transaction" : {
            "data" : [
                "0x1122",
                "0x3344",
                "0x3234"
            ],
            "accessLists" : [
                    {
                        "address" : "0x0000000000000000000000000000000000001337",
                        "storageKeys" : [
                            "0x0000000000000000000000000000000000000000000000000000000000000001",
                            "0x0000000000000000000000000000000000000000000000000000000000000002"
                        ]
                    },
                    {
                    
                    },
                    {
                        "address" : "0x0000000000000000000000000000000000001337",
                        "storageKeys" : [
                            "0x0000000000000000000000000000000000000000000000000000000000000001",
                            "0x0000000000000000000000000000000000000000000000000000000000000002",
                            "0x0000000000000000000000000000000000000000000000000000000000000002"
                        ]
                    },
            ],
            "gasLimit" : [
                "0x061a80"
            ],
            "gasPrice" : "0x01",
            "nonce" : "0x00",
            "secretKey" : "0x45a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065ff2d8",
            "to" : "0x095e7baea6a6c7c4c2dfeb977efac326af552d87",
            "value" : [
                "0x0186a0"
            ]
        }

holiman: I agree about that. To clarify: `{}` can be used for empty list, and `null` to signify that this is a legacy-tx.

holiman: > I don't really know how to unmarshal this properly as the key depends on the length of data. 

Unmarshalling shouldn't be a problem, though, it's just a `map[int]AccessList`. But it's still a bit convoluted

winsvega: you think just null for the transaction with data where accessList not present is better ? 

ok. it is just the matter of test format. which one is easier for you guys.

winsvega:
I removed the map and now it is just an array  (same length as data array)
`{}`  for absense of accessList
`[]`  for empty accessList 

ori alsomst finished his tests and will generate soon on this branch 
https://github.com/ethereum/tests/pull/774

winsvega:
Is there a problem with parsing {}?

Martin wants it to be explicit `null`

holiman:
Yes, the problem is I'm parsing the input into a list type. When it reads an object-type, the parsing fails. Null should be fine, but mixing types is not easy to handle.

winsvega:
added "txbytes"  field in gstate post section. so that no need to build a transaction and do the signing operation from transaction section (complicated logic) 

here also a blockchain test example with access list

PR #773, opened 2021-01-25
Documentation for the t8ntool standard

qbzzt:


holiman:
Hi! 
Nice that you got started working on this! I have also been writing some specifications for the tool, don't know if you've seen it? 
It's here: https://gist.github.com/holiman/5a1e7578bf19115d076de3dcd5b3562c 

Would be good if we can combine the efforts, perhaps you can incorporate my material as you see fit

docs/t8ntool-ref.rst
```diff
@@ -0,0 +1,298 @@
...
+---------------
+- **-\\-state.fork** *fork name*
+- **-\\-state.reward** *block mining reward* (appears only in Block tests)
+- **-\\-trace** produce a trace
+
+Input Files
+-----------
+- **-\\-input.alloc** *full path to pretest allocation file*
+- **-\\-input.txs** *full path to transaction file*
+- **-\\-input.env** *full path to environment file*
```
holiman, line 24: It's either the full path to, but any field can be set to `stdin` if user wants to pass that data as stdin (see my gist)

docs/t8ntool-ref.rst
```diff
@@ -0,0 +1,298 @@
...
+
+
+Output Files
+------------
+- **-\\-output.basedir** *directory to write the output files*
+- **-\\-output.result** *file name for test output*
+- **-\\-output.alloc**  *file name for post test allocation file*
```
holiman, line 35: These two can also use `stdout` if user wants them to pass to standard output

holiman: I also recently added a `output.body`, which outputs a list of rlp transactions

docs/t8ntool-ref.rst
```diff
@@ -0,0 +1,298 @@
...
+   }
+
+
+
+
+
+
+Result File
+-----------
+This file is typically called `out.json`. It is the post state after
```
holiman: The 'typically called' is an implementation-specific choice. It's up to the implementation to decide what to do if the user didn't specify a filename or `stdout`

docs/t8ntool-ref.rst
```diff
@@ -0,0 +1,298 @@
...
+If the transaction fails and does not produce a hash, the name of the file is
+still `trace-<transaction number>-<value that is a legitimate hash>.jsonl`.
+
+
+
+Using Standard Input and Output
+===============================
+It should also be possible to run a `t8ntool` with the input coming from `stdin`
+and the output going to `stdout`. In this case, the input is all one object and
+the output is all one object.
```
holiman: Clarification: 
The input/output is all one object, yes, but you can still have `txs` and `alloc` coming as `stdin` and `env` as file, and send `result` to file but send `alloc` to `stdout`.

qbzzt:
> Hi!
> Nice that you got started working on this! I have also been writing some specifications for the tool, don't know if you've seen it?
> It's here: https://gist.github.com/holiman/5a1e7578bf19115d076de3dcd5b3562c
> 
> Would be good if we can combine the efforts, perhaps you can incorporate my material as you see fit

That is my main source. Well, that and editing `start.sh` to see what actually happens.

docs/t8ntool-ref.rst
qbzzt: Are there any users of `t8ntool` other than `retesteth`? If not, why do we even have `stdin` and `stdout`? They only complicate the implementation for client developers.

holiman: The main idea, I guess, is to allow execution that doesn't touch disk, so e.g. retesteth wouldn't have to actually dump things to disk just to run a testcase, but just feed everything via the proc stdin. 
I didn't know how @winsvega would implement it. 
stdin is faster, but files are more convenient if (as I am mostly) you just run a few tests and want it reproducible.

docs/t8ntool-ref.rst
qbzzt: Bug report. I use this `evm` version:

```
evm version 1.9.26-unstable-6ea49a04-20210124
```

With an absolute path (`--output.body /tmp/x.rlp`) it produces a binary file.

However, with a relative file path (`--output.basedir <whatever> --output.body <filename>`), no output body file is created.

qbzzt: Also, I tried to decode an output.body file using https://codechain-io.github.io/rlp-debugger/, and it didn't work. I'd like to show a sample output in the documentation.

docs/t8ntool-ref.rst
qbzzt: Having the filenames would be helpful, so I'm changing it to:

This is an input to the tool, which `retesteth` calls `txs.json`.

OK?

winsvega:
ready ?

qbzzt:
> ready ?

Yes.

docs/t8ntool-ref.rst
holiman: Right, I'll look into it. I suppose it might be easier for various purposes if I output it as hex-encoded data instead of binary?

qbzzt: I prefer hex-encoded data, but either is fine with me.

holiman: Actually, I prefer hex-encoded too, plus it's more in line with how other data types are output

holiman: This one has three transactions: 

```
[user@work evm]$ dir=./testdata/8
[user@work evm]$ ./evm t8n --state.fork=Berlin --input.alloc=$dir/alloc.json --input.txs=$dir/txs.json --input.env=$dir/env.json --trace --output.body=txs.rlp
INFO [02-01|09:06:23.116] Wrote body file                          file=txs.rlp txs=3
INFO [02-01|09:06:23.116] Wrote file                               file=alloc.json
INFO [02-01|09:06:23.117] Wrote file                               file=result.json
```
This is the hex-encoded body: 
```
cat txs.rlp | xxc
f901e8b8c101f8be0180018304ef0094000000000000000000000000000000000000aaaa0180f85bf859940000000000000000000000000000000000000000f842a00000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000000001a09b83bad6d833b401a9ea725611107e64698e1eac6209f6b7af30b2b62baaece0a0603193024f88264bdb6b864ec88a686823dc1b266598a5740a723acb0e45d19ff86001018304ef0094000000000000000000000000000000000000aaaa028026a05db134df45e75f64425dbae768bebea29f07f72e97694e991908a73bed4a003aa066e7b58c1c7b7b6df1b74e2f318b0fdd7d2f86d0d6e84a8941c9b001f30cbd34b8c101f8be0102018304ef0094000000000000000000000000000000000000aaaa0180f85bf85994000000000000000000000000000000000000aaaaf842a00000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000000301a0e54909b986c516419130c638c06f744d0635c319bdc09b3ab78c5c2d73c25ec9a0311ab841daa142d7d4c5446fd80f0a5278542c8fa0b47f11b7baa236f205f844
```

Decoding that list of three transactions yields: 
```
 cat txs.rlp | rlpdump
[
  01f8be0180018304ef0094000000000000000000000000000000000000aaaa0180f85bf859940000000000000000000000000000000000000000f842a00000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000000001a09b83bad6d833b401a9ea725611107e64698e1eac6209f6b7af30b2b62baaece0a0603193024f88264bdb6b864ec88a686823dc1b266598a5740a723acb0e45d19f,
  [
    01,
    01,
    04ef00,
    000000000000000000000000000000000000aaaa,
    02,
    "",
    "&",
    5db134df45e75f64425dbae768bebea29f07f72e97694e991908a73bed4a003a,
    66e7b58c1c7b7b6df1b74e2f318b0fdd7d2f86d0d6e84a8941c9b001f30cbd34,
  ],
  01f8be0102018304ef0094000000000000000000000000000000000000aaaa0180f85bf85994000000000000000000000000000000000000aaaaf842a00000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000000301a0e54909b986c516419130c638c06f744d0635c319bdc09b3ab78c5c2d73c25ec9a0311ab841daa142d7d4c5446fd80f0a5278542c8fa0b47f11b7baa236f205f844,
]
```

So: `[ blob, rlp-decodable legacy tx, blob]`

When I change the format to use hex-encoded, the example will change slightly, and custom tools like that `xxc` above won't be needed

holiman: Ok, this has now been fixed. Please try `dir=./testdata/8 && ./evm t8n --state.fork=Berlin --input.alloc=$dir/alloc.json --input.txs=$dir/txs.json --input.env=$dir/env.json --output.basedir=/tmp --output.body=stdout --output.result=stdout`.

PR #778, opened 2021-02-02
stEIP2537 fillers [BACKUP]

winsvega:
wait for implementation from geth


qbzzt:
It has 1,403 changed files, that seems a bit high. It might be more manageable if we unify some using different labels for the data.

Or were they created automatically, and this is the easiest way to deal with them?

winsvega:
It was generated by script 

https://github.com/ethereum/tests/pull/712

PR #781, opened 2021-02-14
Underflow tests

qbzzt:
For almost all opcodes that need *n* parameters:

1. Push _n-1_ parameters, run the opcode, see the transaction reverts
2. Push _n_ parameters, run the opcode, see the transaction is successful

Use a code generator (`underflowTestGen.js`) because it's 166 test cases and very repetitive.


qbzzt:
I put the tests in the wrong place

PR #782, opened 2021-02-14
Underflow

qbzzt:
For every possible opcode that takes _n_ parameters:

1. Run with _n_ parameters, see the transaction is successful
2. Run with _n-1_ parameters, see the transaction fails


qbzzt:
This got confused with EIP2930 changes.

PR #795, opened 2021-03-04
logRevert and stateRevert

qbzzt:
logRevert has all the reasons for a revert working
stateRevert looks at SSTOREs both in the failed contract and in contracts it calls


winsvega:
can you also add a description here in a comment for the release notes

qbzzt:
`BlockchainTests/ValidBlocks/bcStateTests/logRevert` tests that if a contract runs a `LOGn` opcode and then reverts the log does not get written. Log entries are not perceptible to the evm, so the test checks the the `BLOCKHASH`, which is accessible. Note that this could result in false positives, because the block hash is so sensitive (it changes when anything in the block changes). To check in that case whether the logs really happened, run this command:

```
<run the test> --verbosity 6 | | grep '\"logs\": \['
```

It will show any log entries that are created by the test.


` GeneralStateTests/stRevertTest/stateRevert` tests that an SSTORE doesn't happen if the contract subsequently reverts. It checks the status of three SSTOREs:

1. One in the contract that calls the reverting contract, which should happen.
2. One in the contract that reverts, which should be cancelled.
3. One in a contract called by the reverting contract, which should also be cancelled.


Both of these tests contain these types of revert:

0: REVERT opcode
1: gas cost overflow (> 2^64)
2: invalid opcode (0xBA)
3: stack underflow  
4: stack overflow
5: jump to addr0 (not a JUMPDEST)
6: gas cost overflow (< 2^64)

winsvega:
was there an issue about this case. log<N> something.

Issue #581, opened 2019-01-15
Make proper test out of stBugs/randomStatetestDEFAULT-Tue_07_58_41-15153-575192.json


chfast:
Hi,

The stBugs/randomStatetestDEFAULT-Tue_07_58_41-15153-575192.json description says

"suicide-created-but-empty account has a non-zero codehash in geth".

It is the only test for this case. It would be good to create a manual state test for the CODEHASH of a "suicide-created-but-empty account".

qbzzt:
It's not just suicide created but empty. If a contract is completely empty (all fields are equal to zero) then `EXTCODEHASH` returns zero. But if there's anything there, such as a balance or a nonce, then `EXTCODEHASH` returns the Keccak-256 of an empty buffer, 
0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470 (see https://medium.com/@ConsenSys/are-you-really-using-sha-3-or-old-code-c5df31ad2b0).

I'm going to write a test to prove that this happens whenever a contract has anything in any field.

PR #799, opened 2021-03-06
EXTCODEHASH has weird behavior

qbzzt:
Verify that, in accordance with https://eips.ethereum.org/EIPS/eip-1052, EXTCODEHASH returns:

- Zero for an empty account
- The hash of the empty string for an account that is not empty, but has no code

"Empty" is as defined in https://eips.ethereum.org/EIPS/eip-161:

> An account is considered empty when it has no code and zero nonce and zero balance.

Issue #581 requested we check this (for a specific case, we expanded to multiple cases)


holiman:
See https://eips.ethereum.org/EIPS/eip-1052

winsvega:
Sum from the EIP

```
A new opcode, EXTCODEHASH, is introduced, with number 0x3f. The EXTCODEHASH takes one argument from the stack, zeros the first 96 bits and pushes to the stack the keccak256 hash of the code of the account at the address being the remaining 160 bits.

In case the account does not exist or is empty (as defined by EIP-161) 0 is pushed to the stack.

In case the account does not have code the keccak256 hash of empty data (i.e. c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470) is pushed to the stack.

The gas cost of the EXTCODEHASH is 400.
```

need to check the implementation. 
1) that gasprice is 400
2) that extcode hash is extcode hash
3) that 0 is returned if account empty or non existent.    <=== here need more tests. because empty is tricky.
4) hash of empty data (empty contract code in non empty account)  is  c5d....

qbzzt:
> 1. that gasprice is 400

It isn't, it changed subsequently. In Istanbul it is 700. In >=Berlin it is 100 if the account has already been touched by the transaction.

> 2. that extcode hash is extcode hash

So not just empty hash? I can do that for a few other contracts with values to be hashed.

> 3. that 0 is returned if account empty or non existent.    <=== here need more tests. because empty is tricky.

Need a few more empty cases:
3.1 Post suicide
3.1.5 Post suicide when during the suicide the contract did something increased its nonce
3.2 Contract was CREATEd with zero code but has values in storage
3.3 Contract was CREATEd with zero code but has a balance
3.4 Contract has zero code but a balance

> 4. hash of empty data (empty contract code in non empty account)  is  c5d....

I checked that

holiman:
> 3.2 Contract was CREATEd with zero code but has values in storage

Note to this: after byzantium (?), a created contract has a non-zero nonce. Therefore, it's (currently) not possible to have a contract which "has storage but is otherwise empty". 
If you create contracts in prestate which has storage but zero nonce, value etc (semantically empty), then you might actually find a few consensus failures which aren't actually exploitable.

qbzzt:
> 3.1 Post suicide
> 3.1.5 Post suicide when during the suicide the contract did something increased its nonce

Suicide doesn't happen until the end of the transaction, so a state test can't check this. I'll add this as a block test

> 3.2 Contract was CREATEd with zero code but has values in storage
> 3.3 Contract was CREATEd with zero code but has a balance

Pointless, because even without values in storage or a balance it gives the hash of the empty string because new contracts are creates with `nonce: 1`.

> 3.4 Contract has zero code but a balance

This is already there, `:label balance`, I just forgot it

PR #801, opened 2021-03-07
subRoutine Fillers [BACKUP]

winsvega:


qbzzt:
Why would we resurrent stSubroutine?

winsvega:
In case its gonna be in London

qbzzt:
It's withdrawn (https://eips.ethereum.org/EIPS/eip-2315). Doesn't that mean it won't come back to life?

I think we should preserve these tests in a branch, but not merge them.

PR #803, opened 2021-03-08
The cost of various types of revert

qbzzt:
This test looks of a call to a contract that reverts. If the call encounters a REVERT opcode, it just costs a bit. But all the errors are equivalent to out of gas, and cost all the gas available to the call.


winsvega:
@qbzzt try putting this label here

PR #811, opened 2021-03-19
Identify precompiles by gas cost

qbzzt:
Not sure if this is ready to merge, because some precompiles are recognized in Berlin but not Istanbul.

### DIFF
--- a/GeneralStateTests/stPreCompiledContracts/idPrecomps.json
+++ b/GeneralStateTests/stPreCompiledContracts/idPrecomps.json
@@ -1,0 +1,5668 @@
+{
+    "idPrecomps" : {
+        "_info" : {
+            "comment" : "Ori Pomerantz qbzzt1@gmail.com",
+            "filling-rpc-server" : "evm version 1.10.1-unstable-5aba9461-20210307",
+            "filling-tool-version" : "retesteth-0.1.0-accesslist+commit.4804afbb.Linux.g++",
+            "lllcversion" : "Version: 0.5.14-develop.2021.3.7+commit.14c9d5de.Linux.g++",
+            "source" : "src/GeneralStateTestsFiller/stPreCompiledContracts/idPrecompsFiller.yml",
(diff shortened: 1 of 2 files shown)


qbzzt:
I fixed a bug, and now it works correctly for Istanbul and Berlin (and should work correctly for future forks).

qbzzt:
Test with earlier networks.

**Byzantium:**

Have 1-8 precompiles, what we'd expect based on the geth source (https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go#L56)


```
Error: Check State: Remote account '0xcccccccccccccccccccccccccccccccccccccccc' has incorrect storage [0x00] = `0x01`, test expected [0x00] = `0x00` (stPreCompiledContracts/idPrecomps, fork: Byzantium, TrInfo: d: 8, g: 0, v: 0, TrData: `:label yes :abi f(uint) 0x09..`)
Error: CompareStates failed with errors: IncorrectStorage (stPreCompiledContracts/idPrecomps, fork: Byzantium, TrInfo: d: 8, g: 0, v: 0, TrData: `:label yes :abi f(uint) 0x09..`)

--------
TestOutputHelper detected 2 errors during test execution!
/retesteth/retesteth/TestOutputHelper.cpp(208): error: in "GeneralStateTests/stPreCompiledContracts": 

*** 1 failure is detected in the test module "Master Test Suite"
info: (stPreCompiledContracts/idPrecomps, fork: Byzantium, TrInfo: d: 8, g: 0, v: 0, TrData: `:label yes :abi f(uint) 0x09..`)
*** Total Tests Run: 0

```


**Homestead:**

In Homestead a bad call uses up all the gas available, not just what it used. So the test needs to be changed instead of `(call (gas) $4 ....)` we need `(call (- (gas) 0x10000 $4 ...)`. After I make that change, it find precompiles in 1-4 but not in 5-9, as expected based on https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go#L47 .

Issue #458, opened 2018-04-20
Create script to generate test cases combining all arithmetic/logic/bitwise instructions


axic:
From https://github.com/ethereum/tests/pull/454#issuecomment-382352633:

> Would it make sense to have a script generate test cases bases on the arithmetic/logic/bitwise instruction list and their in/out arguments? The generator would need to combine at least two of those instructions in a single test case. We could also nominate `add` to be always the second op.

PR #812, opened 2021-03-23
Run two numerical/logical operations

qbzzt:
This was requested by #458 

It is missing the `BlockchainTests/GeneralStateTest`. The reason is that it is a 180 MB monster and git only accepts files up to 100MB.


qbzzt:
Can't have a state test without the corresponding block test (it will get auto generated and cause problems in the future). So I'll rewrite this as a reasonable block test.

PR #813, opened 2021-03-24
Run every two arithmetic/logical op combo

qbzzt:
I put all the calculations in a single contract (`0xccc...cccc`), so now all tests are a reasonable size.


winsvega (approved):
There was an issue asking about non const arguments in vm stack. This test now covers it for the basic operations very well

Issue #364, opened 2017-10-23
Some improvements on call_ecrec_success_empty_then_returndatasize


pirapira:
The test call_ecrec_success_empty_then-returndatasize has some possible improvements.

* the prestate can contain non-zero values in the storage.  This way, it's clear that zero is written to the storage.
* the expectations can be writtten in the filler.

https://github.com/ethereum/tests/pull/362/files#diff-14fa7f288ed243634d91b202f38c67df

winsvega:
could this issue be resolved and closed now?

winsvega:
@qbzzt ori need to review the test and possibly extend the cases

PR #814, opened 2021-03-24
Put a value to be detected if the returndatasize isn't written

qbzzt:
As requested in #364 

I added a call to a non-existent contract to `revertRetDataSize`, so we can probably delete this test - but I'd rather not worry about finding what tests are redundant now. An extra test doesn't hurt us, and if we deleted one we needed it would be bad.


src/GeneralStateTestsFiller/stReturnDataTest/call_ecrec_success_empty_then_returndatasizeFiller.json
```diff
@@ -19,6 +19,7 @@
 		"result" : {
 	 	    "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6" : {
                 	"storage" : {
+                            "0x00": "0x00"
```
winsvega, line 22: Why?

winsvega (approved):

PR #817, opened 2021-03-30
Random tests

qbzzt:
Verify that randomly generated programs create the correct stack values


winsvega:
we don't neccessaraly need to check that execution result of a random program creates correct stack. this are the clients evm will check. we need to find when this stack/memory/hash is different on 2+ clients. 

the idea behind random tests was to generate a lot of random evm code and run it through 2 different clients. like 1000 tests per minute. if they disagree on a post hash, then we investigate the case. the scheme is following:

script generates random test filler. 
client A fills the test
clients B runs the test
results are checked (that no errors reported)

repeat as many times as possible.

winsvega:
does this get outdated by https://github.com/ethereum/tests/pull/949 ?

qbzzt:
Yes.

PR #819, opened 2021-04-07
more state tests examples with new retesteth feautures

winsvega:
- example of using a solidity contract
- example of using yml format
- example of using labels and ranges
- exapmle of using a yul contract


qbzzt:
Do you want me to do sanity checks on these? If not, then you can merge them already.

PR #829, opened 2021-04-20
Tests for EIP2929 operations on precompiles

qbzzt:
Check which low address contracts (precompiles) are part of the EIP2929 access lists. This was the Berlin fork OpenEthereum bug (see https://twitter.com/ralexstokes/status/1382750525070868484 and https://twitter.com/ralexstokes/status/1382750938142629889).

According to EIP2929 (https://eips.ethereum.org/EIPS/eip-2929#abstract) precompiles are exempt from the opcode cost increases for the first (cold) access, they are always hot. However, this only happens with some operations in geth:

1. If you send WEI to a precomp with a CALL, it acts as if the precomp is cold. This is true also if you send or receive data in the call. However, if you send the WEI with CALLCODE, it works normally. 
4. Accounts 1-9 act cold for EXTCODECOPY

The expected results in this test are based on observed geth behavior.

**Fixed by using a constant amount to gas as @holiman suggested:**

2. If you send data, 1-7 act hot, 8 acts cold. This is true for CALL, CALLLCODE, DELEGATECALL, and STATICCALL
3. Account 9 acts cold for calls (all four types, all variants I tested), but  acts hot for BALANCE, EXTCODEHASH, and EXTCODESIZE

### DIFF
--- a/GeneralStateTests/stPreCompiledContracts/precompsEIP2929.json
+++ b/GeneralStateTests/stPreCompiledContracts/precompsEIP2929.json
@@ -6,17 +6,17 @@
             "filling-tool-version" : "retesteth-0.1.1-eip1559+commit.c4a37f72.Linux.g++",
             "lllcversion" : "Version: 0.5.14-develop.2021.4.20+commit.14c9d5de.Linux.g++",
             "source" : "src/GeneralStateTestsFiller/stPreCompiledContracts/precompsEIP2929Filler.yml",
-            "sourceHash" : "b860ec78316f15b6537afb7cd02b60236d39f00d8be8e894f1920cb95ce2e0c6",
+            "sourceHash" : "ab2a267aca21c5c7e3e717ea928c5a244ad9b5e2f22eea5c7043f596419a03f5",
             "labels" : {
                 "0" : ":label yes",
                 "1" : ":label yes",
                 "10" : ":label no",
-                "100" : ":label no",
-                "101" : ":label no",
-                "102" : ":label no",
-                "103" : ":label no",
-                "104" : ":label no",
-                "105" : ":label no",
+                "100" : ":label all",
+                "101" : ":label all",
+                "102" : ":label all",
+                "103" : ":label all",
+                "104" : ":label all",
+                "105" : ":label all",
                 "106" : ":label no",
                 "107" : ":label yes",
                 "108" : ":label yes",
@@ -98,7 +98,7 @@
                 "177" : ":label no",
                 "178" : ":label no",
                 "179" : ":label yes",
-                "18" : ":label no",
+                "18" : ":label new",
                 "180" : ":label yes",
                 "181" : ":label yes",
                 "182" : ":label yes",
@@ -109,7 +109,7 @@
                 "187" : ":label yes",
                 "188" : ":label yes",
                 "189" : ":label no",
-                "19" : ":label no",
+                "19" : ":label new",
                 "190" : ":label no",
                 "191" : ":label no",
                 "192" : ":label no",
@@ -121,7 +121,7 @@
                 "198" : ":label yes",
                 "199" : ":label yes",
                 "2" : ":label yes",
-                "20" : ":label no",
+                "20" : ":label new",
                 "200" : ":label yes",
                 "201" : ":label yes",
                 "202" : ":label yes",
@@ -132,7 +132,7 @@
                 "207" : ":label no",
                 "208" : ":label no",
                 "209" : ":label no",
-                "21" : ":label no",
+                "21" : ":label new",
                 "210" : ":label no",
                 "211" : ":label no",
                 "212" : ":label no",
@@ -143,7 +143,7 @@
                 "217" : ":label yes",
                 "218" : ":label yes",
                 "219" : ":label yes",
-                "22" : ":label no",
+                "22" : ":label new",
                 "220" : ":label yes",
                 "221" : ":label yes",
                 "222" : ":label yes",
@@ -154,7 +154,7 @@
                 "227" : ":label no",
                 "228" : ":label no",
                 "229" : ":label no",
-                "23" : ":label no",
+                "23" : ":label new",
                 "230" : ":label no",
                 "231" : ":label no",
                 "232" : ":label no",
@@ -165,7 +165,7 @@
                 "237" : ":label yes",
                 "238" : ":label yes",
                 "239" : ":label yes",
-                "24" : ":label no",
+                "24" : ":label new",
                 "240" : ":label yes",
                 "241" : ":label yes",
                 "242" : ":label yes",
@@ -176,7 +176,7 @@
                 "247" : ":label no",
                 "248" : ":label no",
                 "249" : ":label no",
-                "25" : ":label no",
+                "25" : ":label new",
                 "250" : ":label no",
                 "251" : ":label yes",
                 "252" : ":label yes",
@@ -198,7 +198,7 @@
                 "267" : ":label no",
                 "268" : ":label no",
                 "269" : ":label yes",
-                "27" : ":label no",
+                "27" : ":label all",
                 "270" : ":label yes",
                 "271" : ":label yes",
                 "272" : ":label yes",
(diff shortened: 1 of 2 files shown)


holiman:
> If you send WEI to a precomp with a CALL, it acts as if the precomp is cold. This is true also if you send or receive data in the call. However, if you send the WEI with CALLCODE, it works normally.

That can't be right. I'll look into it

> If you send data, 1-7 act hot, 8 acts cold. This is true for CALL, CALLLCODE, DELEGATECALL, and STATICCALL

I don't think this is rirght. I assume that `8` simply rejects your input. 

> Account 9 acts cold for calls (all four types, all variants I tested), but acts hot for BALANCE, EXTCODEHASH, and EXTCODESIZE

I think it rejects your input. It's blake2f, it requires exactly `213` bytes of input to not error. 


1-7 have always been precompiles

holiman:
So it's better to use e.g. EXTBALANCE for the coldness-check, then you don't get it mixed up with the per-precompile unique cost for the precompile itself.

holiman:
Also, we already have tests which touch the low accounts, at least that was my impression?

qbzzt:
> Also, we already have tests which touch the low accounts, at least that was my impression?

Yes, but we didn't have tests for every action. For example, we didn't have tests for calling a precomp while sending it wei (the thing that broke openethereum)

> > If you send data, 1-7 act hot, 8 acts cold. This is true for CALL, CALLLCODE, DELEGATECALL, and STATICCALL
> I don't think this is rirght. I assume that 8 simply rejects your input.

The issue isn't that my call is rejected, I completely ignore the result. The issue is that first call to account 8 costs more gas than the second call.

> > Account 9 acts cold for calls (all four types, all variants I tested), but acts hot for BALANCE, EXTCODEHASH, and EXTCODESIZE

> I think it rejects your input. It's blake2f, it requires exactly 213 bytes of input to not error.

Again, the issue isn't that my input is invalid. It is that a first call and a second call that should cost the same don't.

holiman:
> Yes, but we didn't have tests for every action. For example, we didn't have tests for calling a precomp while sending it wei (the thing that broke openethereum)

That wasn't what broke it. The issue was that `0x10` was treated as `hot`, because it was considered part of the 'precompiles'. This was due to it being defined (but not yet active) in the particular chain config for mainnet. This was not the case of the `Berlin`-specific test-config. 

> The issue isn't that my call is rejected, I completely ignore the result. The issue is that first call to account 8 costs more gas than the second call.

I'll look into it. I think there's some mistake here

holiman:
> The issue isn't that my call is rejected, I completely ignore the result. The issue is that first call to account 8 costs more gas than the second call.

If you know which indexes are relevant here, it'd be appreciated..

holiman:
Possible source of error, AFAICT, you send all `gas`: `pop(call(gas())`. This means that you send along everything. Normally, that's fine, and it sends along only `63/64`ths of the available gas. Now, if you do a call to `blake`, and have erroneous input, it will eat up all that gas, essentially almost all your gas. If you call it a second time, you won't have a much gas left, and this time, you'll only "lose" `63/64`ths of what you have now.

qbzzt:
> Possible source of error, AFAICT, you send all `gas`: `pop(call(gas())`. This means that you send along everything. Normally, that's fine, and it sends along only `63/64`ths of the available gas. Now, if you do a call to `blake`, and have erroneous input, it will eat up all that gas, essentially almost all your gas. If you call it a second time, you won't have a much gas left, and this time, you'll only "lose" `63/64`ths of what you have now.

Thank you. I'll redo the test with fixed amounts to check

Edit: Using a fixed amount of gas did fix the issues with 8 and 9. So two issues down, two to go.

holiman:
> So two issues down, two to go.

Which are still 'strange" ?

qbzzt:
> > So two issues down, two to go.
> 
> Which are still 'strange" ?

1. The first time we send wei to a precompile (with `CALL`, not `CALLCODE`) is a different cost than the second time
2. The first time we `EXTCODECOPY` from a precompile is a different cost than the second time

holiman:
> The first time we send wei to a precompile (with CALL, not CALLCODE) is a different cost than the second time

It's the 25K cost to create an account in the trie. CALLCODE doesn't actually create a new account, so it's just a transfer to self, hence no 25K addon. If you were to "create" the precompiles in the 'state', then you wouldn't see this. 


> The first time we EXTCODECOPY from a precompile is a different cost than the second time

That sounds a bit odd. I'll look into it (any idea about what index I should look at?)

qbzzt:
> > The first time we send wei to a precompile (with CALL, not CALLCODE) is a different cost than the second time
> 
> It's the 25K cost to create an account in the trie. CALLCODE doesn't actually create a new account, so it's just a transfer to self, hence no 25K addon. If you were to "create" the precompiles in the 'state', then you wouldn't see this.

Are precompiles "created" on mainnet's state? Or is this behavior consistent across clients so it doesn't matter?

> 
> > The first time we EXTCODECOPY from a precompile is a different cost than the second time
> 
> That sounds a bit odd. I'll look into it (any idea about what index I should look at?)

In the latest version I committed, `-d 348`

holiman:
> Are precompiles "created" on mainnet's state? Or is this behavior consistent across clients so it doesn't matter?

Hah, no, well, yes. They were incrementally created, as people sent money to their addresses. However, when Byzantium rolled in, and we were doing state cleanup by touching empty accounts, the ripemd precompile was touched, _and_ the tx went OOG. It was a shitshow, and caused a still-remaining issue on how to treat empty precompile touches. 

If we ever get a new empty precompile, that would be a problem, but no, we never will, so it's theoretical. A lot of them were inited by this account (https://etherscan.io/address/0x8db98f07d8bec0038da3d13ed93adcf09316ef66) at least up to `0x00..0f`: https://etherscan.io/address/0x000000000000000000000000000000000000000f. At some point, we need to 'create' https://etherscan.io/address/0x0000000000000000000000000000000000000017 (`0x00..17`). Damn, looks like we need it before bls!

holiman:
I trimmed away all but the `d=348`, and ran it as `./build/bin/evm --json statetest /tmp/precompsEIP2929.json 2> trace.json`, to collect the trace. 
Then use `traceview` from https://github.com/holiman/goevmlab  .Can be installe via `go install github.com/holiman/goevmlab/cmd/traceview`, and do `traceview trace.json`. 

Here's the trace right before the first `EXTCODECOPY`. Note the memory size: 
extcodecopy_1

As we execute the EXTCODECOPY, memory expansion is required: 
extcodecopy_2

But for the last EXTCODECOPY, no further expansion is needed, so it's cheaper: 
extcodecopy_3

qbzzt:
> But for the last EXTCODECOPY, no further expansion is needed, so it's cheaper:

Thank you. Sorry, I thought I dealt with it by touching the memory, but then I made the mistake of reading too much data.

qbzzt:
> > Are precompiles "created" on mainnet's state? Or is this behavior consistent across clients so it doesn't matter?
> 
> Hah, no, well, yes. They were incrementally created, as people sent money to their addresses. However, when Byzantium rolled in, and we were doing state cleanup by touching empty accounts, the ripemd precompile was touched, _and_ the tx went OOG. It was a shitshow, and caused a still-remaining issue on how to treat empty precompile touches.
> 
> If we ever get a new empty precompile, that would be a problem, but no, we never will, so it's theoretical. A lot of them were inited by this account (https://etherscan.io/address/0x8db98f07d8bec0038da3d13ed93adcf09316ef66) at least up to `0x00..0f`: https://etherscan.io/address/0x000000000000000000000000000000000000000f. At some point, we need to 'create' https://etherscan.io/address/0x0000000000000000000000000000000000000017 (`0x00..17`). Damn, looks like we need it before bls!

OK, so there's no chance of this issue being a problem in the real world. I'll keep the tests, but add a note about the reason.

holiman:
>  I'll keep the tests, but add a note about the reason.

SGTM! Glad we cleared all the questions out!

qbzzt:
> > I'll keep the tests, but add a note about the reason.
> 
> SGTM! Glad we cleared all the questions out!

Thank you for answering them.

Issue #831, opened 2021-04-26
Missing memory copy out-of-gas cases


chfast:
Hi there,

I'd like to report some missing test cases for instructions which copy memory and the out-of-gas failure is directly result of the additional memory copy cost:
- `KECCAK256`,
- `CODECOPY`,
- `EXTCODECOPY`,
- `RETURNDATACOPY`.

---
![image](https://user-images.githubusercontent.com/573380/116072321-35518080-a68f-11eb-853f-da0cce9a6393.png)
---
![image](https://user-images.githubusercontent.com/573380/116072387-4ac6aa80-a68f-11eb-881c-31a1cf82f259.png)
---
![image](https://user-images.githubusercontent.com/573380/116072778-d5a7a500-a68f-11eb-867e-86f35d9a876c.png)

The coverage is from https://github.com/ethereum/evmone. 
https://codecov.io/gh/ethereum/evmone/src/eeea491ed8d8cb29c895131a97feabc8a125c542/lib/evmone/instructions.hpp.

chfast:
The case for `RETURNDATACOPY` is still not covered. The rest is, thanks.

qbzzt:
> The case for `RETURNDATACOPY` is still not covered. The rest is, thanks.

The test is in https://github.com/ethereum/tests/blob/a92affb177712ebb9b45cf54aee6aa72d25fdf45/src/GeneralStateTestsFiller/stMemoryTest/oogFiller.yml#L66. The call for it is in https://github.com/ethereum/tests/blob/a92affb177712ebb9b45cf54aee6aa72d25fdf45/src/GeneralStateTestsFiller/stMemoryTest/oogFiller.yml#L306.

chfast:
Maybe it fails for some other reason than the cost of memory copy inside `RETURNDATACOPY`. Do you have a trace?

qbzzt:
You're right, my bad. I forgot to account for the fact that in Berlin calls are more expensive than they were in Istanbul. I'll fix it.

PR #833, opened 2021-04-27
Fixed a bug caused by the higher price of a call in Berlin

qbzzt:
Solving the bug in #831 . 

@chfast , can you please verify that this works on your environment and then I'll close the issue, hopefully not prematurely?

   Thanks,
        Ori


chfast:
First of all, I'm very satisfied how my reports are handled. These are not critical issues and very good process has been made recently. There are still some uncovered cases and I will report these gradually.

In this case, there is still **no hit**. I just realized that in my implementation there are 2 memory checks. First the `RETURNDATACOPY` checks the memory expansion cost. This is covered. But after the memory is successfully expanded, there is additional 3 gas per byte of copying from return data buffer to the memory. The OOG failure at this point is sill not covered.

qbzzt:
As they say in sniper school, "we aim to please". 

> In this case, there is still no hit. I just realized that in my implementation there are 2 memory checks. First the RETURNDATACOPY checks the memory expansion cost. This is covered. But after the memory is successfully expanded, there is additional 3 gas per byte of copying from return data buffer to the memory. The OOG failure at this point is sill not covered.

We don't normally look inside individual opcodes, but I added a test that runs with 3073 gas, one less than the amount actually needed. I hope it fails in the correct place. If it doesn't, please send me a trace so I'll know how much gas to allocate to get the correct failure.

chfast:
This works as expected now. Thanks.

qbzzt:
> This works as expected now. Thanks.

Thank you, I'll close the issue. If you have anything else, ask and we'll try to provide.

winsvega:
why 16k lines removed? just disabling one fork leads to this?

qbzzt:
> why 16k lines removed? just disabling one fork leads to this?

Yes. The file is ~16k for Berlin, and it was ~16k for Instanbul. 

I'll rebase later today after I'm done with dayjob stuff.

PR #834, opened 2021-04-28
The effects EIP2930 on "manual" account creation

qbzzt:
If we add the address where the contract will be created to the access list, it is valid.


src/GeneralStateTestsFiller/stEIP2930/manualCreateFiller.yml
```diff
@@ -0,0 +1,126 @@
...
+    a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+      balance: 0x1000000000000000000
+      code: 0x
+      nonce: 1
+      storage: {}
+
+
+  transaction:
+    data:
+    # To create a contract directly, we send a transaction with the
```
winsvega, line 31: You can use lll code Return

qbzzt: To provide contract code, but what's the point when I can't call it anyway (because after the transaction the call is over)? Or do you mean I should write a blockchain test for this?

winsvega: I mean you can write human readable code instead of bytecode.
like here 
```
(seq (CREATE 0 0 (lll (seq  [[1]] (CALLCODE 500000 0x1000000000000000000000000000000000000001 1 0 0 0 0)) 0)   )           )
```

smth similar might work with return

PR #835, opened 2021-04-29
Tests for EIP-3541

gumb0:
Reject new contracts starting with the 0xEF byte https://eips.ethereum.org/EIPS/eip-3541

Imlementation is merged in geth https://github.com/ethereum/go-ethereum/pull/22809


src/GeneralStateTestsFiller/stCreateTest/CREATE_EOF1Filler.yml
```diff
@@ -0,0 +1,174 @@
...
+    value:
+    - 0
+
+  expect:
+    - indexes:
+       data: ':label deploying_0xef'
+       gas: !!int -1
+       value: !!int -1
+      network:
+        - 'Berlin'
```
axic, line 49: Would this work?

```suggestion
        - '<=Berlin'
```

src/GeneralStateTestsFiller/stTransactionTest/CreateTransactionEOF1Filler.yml
```diff
@@ -0,0 +1,141 @@
...
+    a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+      balance: 100000
+      code: ''
+      nonce: 0
+      storage: {}
+
+  transaction:
+    data:
+      #- ':yul { mstore8(0, 0xef) return(0, 1) }'
+      - ':label deploying_0xef :raw 0x60ef60005360016000f3'
```
winsvega, line 23: Yul didn't  work?

gumb0: It did, I just wanted to show raw code here, maybe include it in EIP text.

src/GeneralStateTestsFiller/stCreateTest/CREATE_EOF1Filler.yml
gumb0: Nonce for created accounts was 0 before state-clearing fork, so we would need either make tests a bit more complicated with another nonce expectation for old forks, or remove nonce expectation, or list only forks starting from SpuriousDragon.

But it looks usual in new tests to check only new behaviour and maximum of one previous fork, not caring about repeating it for old forks with the same rules. @winsvega can correct me if it's not ok approach.

winsvega: Yes because we have copy of the tests in legacy with old forks expect section. However new tests are targeting only recent forks (as also pointed out by Alexy Akhunov)

Since we already have client consensus with existing tests, no really need to cover previous forks.

You can use a sstore operation and check that it worked (instead of nonce)

Another issue when having tests with 5 or more forks it takes more time to generate and run it. So I plan to archive tests from time to time refreshing the expect section to the most recent fork only.

src/GeneralStateTestsFiller/stBadOpcode/undefinedOpcodeFirstByteFiller.yml
```diff
@@ -0,0 +1,760 @@
+# Checks that behaviour of already deployed code, having invalid opcodes as the first byte.
```
gumb0: This could be merged separately already, because EIP-3541 doesn't change this behavior.

src/GeneralStateTestsFiller/stTransactionTest/Opcodes_TransactionInitFiller.json
```diff
@@ -306,7 +369,9 @@
                 ":raw 0x6000600060006000730f572e5295c57f15886f9b263e2f6d2d6c7b5ec6620186a0f45060006000f3", 
                 ":raw 0x6000600060006000730f572e5295c57f15886f9b263e2f6d2d6c7b5ec6612710fa5060006000f3", 
                 ":raw 0x60006000fd60006000f3", 
-                ":raw 0x32ff"
+                ":raw 0x32ff",
+                ":label ef_first_byte :raw 0xef",
```
gumb0: I didn't care to make all possible invalid opcodes here.

And this also can be merged independently, because this behavior is not changing yet.

gumb0:
@winsvega On the last ACD call EIP-3541 was "accepted for London unless anyone objects on the next call", so I think it would be ok to start reviewing it.

I will keep PR in draft until EIP is accepted.

axic:
> I will keep PR in draft until EIP is accepted.

Just to shrink the size, do you think it makes sense splitting this and merging the non-London specific tests? i.e. would that just mean a test creating a contract with every single starting byte from 0x00 to 0xff?

gumb0:
> > I will keep PR in draft until EIP is accepted.
> 
> Just to shrink the size, do you think it makes sense splitting this and merging the non-London specific tests? i.e. would that just mean a test creating a contract with every single starting byte from 0x00 to 0xff?

Everything could be merged if we remove London expectations.
But I guess tests that deploy `0xef00...00` contracts would not be very interesting for Berlin.
Tests that deploy all possible starting bytes might be.
And the test `undefinedOpcodeFirstByte ` which checks already deployed contracts behaviour is kind of independent.

I would leave for @winsvega to decide how it would be convenient to split this, if needed.

winsvega:
If there is no London specific logic we can keep the test under any name. Tests fork configs are now translatable with clients configs in retesteth config dir

winsvega:
Is this eip enabled in Aleut?

gumb0:
> Is this eip enabled in Aleut?

No. https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/client-integration-testnets/aleut.md

But included in Baikal. https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/client-integration-testnets/baikal.md

gumb0:
- Latest geth master supports EIP-3541 as part of London fork, so I replaced `Berlin+3541` with `London` now.
- Pulled out `undefinedOpcodeFirstByte` test into https://github.com/ethereum/tests/pull/836
- Added (generated) blockchain tests.

gumb0:
@winsvega I have added `currentBaseFee` required by the latest change in test format and regenerated tests with the latest geth and retesteth. I think this is ready for review.

The commit `Generate London cases for stTransactionTest/Opcodes_TransactionInit` is separate only to make simpler the diff for the last commit `Init code failing to deploy EF byte and rolling back the side effects.`

src/GeneralStateTestsFiller/stTransactionTest/Opcodes_TransactionInitFiller.json
```diff
@@ -4,6 +4,7 @@
             "currentCoinbase" : "2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
             "currentDifficulty" : "0x20000",
             "currentGasLimit" : "1000000",
+            "currentBaseFee": "1",
```
gumb0: Forgot to set this to `10`

gumb0: Fixed.

src/GeneralStateTestsFiller/stCreateTest/CREATE_EOF1Filler.yml
winsvega: for full node implementation IMHO it is good to have tests that check that new changes does not break anything in the past. so if you really think we should check it here we can make an exception.

winsvega (approved):
I think this looks really good now. 
It is great to have contributions from EIP developers finaly. 
Great thanks to Ori and his documentation.

PR #836, opened 2021-05-06
Calling a contract which has an undefined opcode as first byte of code

gumb0:
Pulled out of #835


gumb0:
I realize now that it should be changed for London, because `0x48` is going to be defined as `BASEFEE`.

@winsvega Can we merge this first as is? Do you already merge changes for London?

winsvega:
Not yet. Eip1559 is in development

gumb0:
> Not yet. Eip1559 is in development

Then I guess this should be fine to merge now. I've added blockchain tests.

Please double check the list of undefined opcodes.

axic:
> Please double check the list of undefined opcodes.

I looked at `undefinedOpcodeFirstByteFiller.yml` and compared against the opcode table in evmeone and it looked correct to me.

axic:
@winsvega @qbzzt could this be merged?

winsvega:
It is good to have a separate test, but now we have so much data copied for every transaction. 
Each blockchain test has a copy of the whole state, but executes just one contract from that state. 

Perhaps its better to design the call loop in one transaction otherwise we have 100k+ of test lines?

winsvega (commented):
It is good to have a separate test, but now we have so much data copied for every transaction.
Each blockchain test has a copy of the whole state, but executes just one contract from that state.

Perhaps its better to design the call loop in one transaction otherwise we have 100k+ of test lines?

also should have all possible opcodes as the first byte, so we have a map of which opcodes are supposed to halt the execution, and make sure that others are fine.
then we change this map over forks.

gumb0:
In the second commit I changed it to be the single transaction calling all contracts in a loop and saving those that succeed (expected none to succeed)

In the third commit I changed it to test all possible bytes as a first byte, but then expected section contains the list of those that don't require values on stack and execution is successful.

Perhaps final version is better, but then it's better to rename the test (it's not about undefined opcodes anymore)

@winsvega any thoughts?

winsvega:
yes this looks much better about space. 
to stay safe you can add one more sstore after the loop so to be sure that test didn't run oog while going through the loop and all opcodes are called

PR #837, opened 2021-05-06
EIP1559 blockchain tests

qbzzt:
**InvalidBlocks**

1. `badBlocks`, checks for blocks with invalid values in the header
2. `feeCap`, checks that the fee cap is implemented correctly
3. `gasLimit20m` and 4. `gasLimit40m`, tests cases from the geth tests https://github.com/ethereum/go-ethereum/blob/c9116a7c19301d6906c1aff87085c2fa345a84bf/consensus/misc/eip1559_test.go#L78

**ValidBlocks**

5. `lowDemand`, checks that when gasUsed << gasTarget  baseFee <- baseFee*(7/8)
6. `medDemand`, checks that when gasUsed = gasTarget  baseFee does not change
7. `highDemand`, checks that when gasUsed = 2*gasTarget  baseFee <- baseFee*(9/8)
8. `checkGasLimit`, check that transactions that cause `gasUsed` to exceed `gasLimit` are invalid
9. `tips`, check that the tipped amount given to the miner is correct
10. `transTypes`, check that all three transaction types (original, type 1 (EIP 2930), and type 2 (EIP 1559)) work.
11. `baseFee`, check the behavior of basefee and selfbalance when called in multiple transactions in multiple blocks.
12. `burnVerify`, verify the correct amount of wei is burned

Note: The tests needed are at https://hackmd.io/NNforJS7SMmjJ1t6jA8pgQ


qbzzt:
I thought that it'll be a while until I can do `gasUsed > gasTarget` blocks. Since we already have them, lets wait until I write the tests that use them too.

winsvega:
I think we can remove the python sanitizer scripts and json schemas from here [in a separate PR]. as retesteth checks structures validity upon test generation. and also checks that tests are updated in automatition.

src/BlockchainTestsFiller/InvalidBlocks/bcEIP1559/badBlocksFiller.yml
```diff
@@ -0,0 +1,213 @@
...
+      secretKey: 41f6e321b31e72173f8ff2e292359e1862f24fba42fe6f97efaf641980eff298
+      to: 0x1111111111111111111111111111111111111111
+      value: 0x0
+
+
+  expect:
+  - network:
+    - ">=London"
+    result: {}
+#      0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC:
```
winsvega: comment forgotten

src/BlockchainTestsFiller/InvalidBlocks/bcEIP1559/badBlocksFiller.yml
```diff
@@ -0,0 +1,213 @@
...
+    mixHash: '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421'
+    nonce: '0x0102030405060708'
+    number: '0'
+    parentHash: '0x0000000000000000000000000000000000000000000000000000000000000000'
+    receiptTrie: '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421'
+    stateRoot: '0xf99eb1626cfa6db435c0836235942d7ccaa935f1ae247d3f1c21e495685f903a'
+    timestamp: '0x03b6'
+    transactionsTrie: '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421'
+    uncleHash: '0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347'
+    gasTarget: 1024000000
```
winsvega: there seem would be an issue as the gasTarget is turning back to gasLimit 
and the gasTarget meaning would be calculated from gasLimit in the block

so not to confuse we will have to rename it to gasLimit again and adjust the values.

src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/feeCapFiller.yml
```diff
@@ -0,0 +1,194 @@
...
+         }
+      storage: {}
+
+
+
+
+  blocks:
+  
+  - transactions:
+    # gasPrice of 875 = 0x375, the current baseFee (after the genesis block)
```
winsvega, line 53: if you need any of blockheader fields fixed here you can use the BlockHeader overwrite section. that will make sure that those fields are fixed and not manipulated by the client. such as baseFee which actually can change as the EIP is ongoing discussion

src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/feeCapFiller.yml
```diff
@@ -0,0 +1,194 @@
...
+        storage:
+          0x36b: 1  # gasPrice 875 (= baseFee)
+          0x36c: 1
+          0x36d: 1
+          0x36e: 1
+          0x36f: 1
+          0x370: 1  # gasPrice 880 (= feeCap)
+
+
+          0x371: 0  # gasPrice 881 (> feeCap, so the transaction doesn't happen)
```
winsvega: it does happen, it seems to be reverted. which means fees are lost (

src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/highDemandFiller.yml
```diff
@@ -0,0 +1,268 @@
...
+      tip: 0x00
+      feeCap: 2000
+      nonce: auto
+      secretKey: 41f6e321b31e72173f8ff2e292359e1862f24fba42fe6f97efaf641980eff298
+      to: 0x1111111111111111111111111111111111111111
+      value: 0x0
+
+
+  - blockHeader:
+      gasTarget: 100000
```
winsvega: is there no formula for gasTarget variable?

src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/lowDemandFiller.yml
```diff
@@ -0,0 +1,781 @@
...
+          0x30: 0x0008 #
+
+          # If we put a block that doesn't use any gas,
+          # the ratio is exactly 7/8's, so the following block
+          # has a baseFee of 7.
+
+          0x32: 0x0007
+
+          # baseFee doesn't go below 7. 7*7/8 is 6.125, but then
+          # it is rounded up
```
winsvega, line 780: this is a very cool test. here you can see the capacity of blockchain format!

winsvega (commented):
very cool! 
lets see if we have to rename gasTarget in our structures too. this will need some modification

src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/highDemandFiller.yml
qbzzt: No. Miners are supposed to play with it so we'll have blocks at a certain frequency, IIRC, but there isn't really one.

src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/feeCapFiller.yml
qbzzt: You're right, I'll clarify the comment.

src/BlockchainTestsFiller/InvalidBlocks/bcEIP1559/badBlocksFiller.yml
qbzzt: Just tell me when retesteth supports the new gasLimit behavior, and I'll modify the tests accordingly. I removed the "ready to merge" because until that happens it's not ready.

src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/lowDemandFiller.yml
qbzzt: Yes.

winsvega:
need to review

src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/transTypeFiller.yml
```diff
@@ -0,0 +1,155 @@
...
+
+
+
+
+  expect:
+  - network:
+    - ">=London"
+
+    # The test here is that all the transactions are valid
+    result: {}
```
winsvega: check the nonce of account `0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b:`

src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/tipsFiller.yml
```diff
@@ -0,0 +1,565 @@
...
+    nonce: '0x0102030405060708'
+    number: '0'
+    parentHash: '0x0000000000000000000000000000000000000000000000000000000000000000'
+    receiptTrie: '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421'
+    stateRoot: '0xf99eb1626cfa6db435c0836235942d7ccaa935f1ae247d3f1c21e495685f903a'
+    timestamp: '0x03b6'
+    transactionsTrie: '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421'
+    uncleHash: '0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347'
+    gasLimit: 10000000000
+    gasUsed:   10000000000
```
winsvega, line 21: gasUsed > 0 in genesis block should produce an error no?

src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/medDemandFiller.yml
```diff
@@ -0,0 +1,560 @@
...
+      secretKey: 41f6e321b31e72173f8ff2e292359e1862f24fba42fe6f97efaf641980eff298
+      to: 0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC
+      value: 0x0
+
+
+
+
+
+
+  expect:
```
winsvega, line 531: perhaps check here that all the gas from transaction is burned from the state.

src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/tipsFiller.yml
qbzzt: I was hoping to keep `baseFee` instead of having it become 7/8'th of the value I specified, but that didn't happen. Instead, it looks like `gasUsed` in the genesis block is just ignored.

Your choice if you want to emit an error on it.

src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/checkGasLimitFiller.yml
```diff
@@ -0,0 +1,178 @@
...
+
+
+
+
+  expect:
+  - network:
+    - ">=London"
+    # No real result, this test is based on whether transactions are 
+    # considered valid or not
+    result: {}
```
winsvega: can check the nonce of how many in total transactions went through

src/BlockchainTestsFiller/InvalidBlocks/bcEIP1559/badBlocksFiller.yml
```diff
@@ -0,0 +1,134 @@
...
+  # else:
+  #    gas_used_delta = parent_gas_target - parent_gas_used
+  #    base_fee_per_gas_delta = parent_base_fee_per_gas * gas_used_delta // parent_gas_target // BASE_FEE_MAX_CHANGE_DENOMINATOR
+  #    expected_base_fee_per_gas = max(parent_base_fee_per_gas - base_fee_per_gas_delta, 0)
+  # assert expected_base_fee_per_gas == block.base_fee_per_gas, 'invalid block: base fee not correct'
+
+  - blockHeader:
+      baseFee: 0x36c
+    expectException:
+      '>=London': baseFeeWrong
```
winsvega: the exception name in the config I put as a standard 
`1559BlockImportImpossible_baseFeeWrong`
that was a while ago, please check that you use the latest retesteth and that all the tests from this PR fill on latest geth.
please rename the exceptions

winsvega (changes requested):
please make sure that geth and retesteth are at latest
fix the exception names. 
address the comments 
and regenerate the tests

src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/medDemandFiller.yml
qbzzt: It's easier to do it in a separate test, `burnVerifier`.

src/BlockchainTestsFiller/InvalidBlocks/bcEIP1559/badBlocksFiller.yml
qbzzt: Oops, sorry. Fixed

winsvega:
what happened with the branch. why changes from another branch here ?

winsvega (approved):

holgerd77:
Congrats to the merge here! 🎉  😀

holgerd77:
Hi there,
I am just integrating these tests here: https://github.com/ethereumjs/ethereumjs-monorepo/pull/1269

Just noted that the base fee is still called `baseFee` here in the tests, the agreement here was to use `baseFeePerGas`, at least for the [JSON RPC](https://github.com/ethereum/eth1.0-specs/blob/master/json-rpc/spec.json#L1489).

Is this due to an outdated Geth version and will be updated?

//cc @holiman

holiman:
Yeah, we have several PRs in the pipeline, and I think rpc naming is that last one, so we're definitely not up to spec at the moment

holgerd77:
@holiman ah ok, thanks, I've temporarily added a hot fix on our side.

holgerd77:
We've just run the tests from here, current failures on our side (see https://github.com/ethereumjs/ethereumjs-monorepo/pull/1269):

## State Tests

2 failures (total 150)

- `basefeeExample`
- `eip1559`

## Blockchain Tests

10 failures (total 1275)

- `badBlocks`
- `gasLimit20m`
- `gasLimit40m`

Haven't gone into any analysis, so totally open if (more likely) things are wrong on our side or if there might be some test bugs, just dropping here for reference and for others to compare if they also stumble upon things.

ryanio:
Update from EthereumJS: state tests are now passing (it was within our own code)

For the Blockchain Tests, I think these failures are due to outdated versions of the spec that test `>=` and `<=` rather than `>` and `<`? ([PR](https://github.com/ethereum/EIPs/pull/3566))

The test comments for [`badBlocks`](https://github.com/ethereum/tests/blob/4d539de6788147a0376b70f2bc71a8f818872164/src/BlockchainTestsFiller/InvalidBlocks/bcEIP1559/badBlocksFiller.yml#L100-L102) still reference the outdated spec:

```
  # From the EIP:
  #
  # assert block.gas_target >= parent_gas_target - parent_gas_target // 1024, 'invalid block: gas target decreased too much'
  ... 
  (this block:)
  - blockHeader:
      gasLimit: 1072693248    # 1024*1024*1023 (1023/1024 of the previous gasLimit)
    transactions: []
```

Block gasLimit: 1072693248
Max gasLimit: 1072693248

---

For [`gasLimit20m`](https://github.com/ethereum/tests/blob/4d539de6788147a0376b70f2bc71a8f818872164/src/BlockchainTestsFiller/InvalidBlocks/bcEIP1559/gasLimit20mFiller.yml#L75-L77), it is on the same boundary:

```
 - blockHeader:
      gasLimit: 20019531
    transactions: []
```

Block gasLimit: 20019531
Max gasLimit: 20019531

---

Same with [`gasLimit40m`](https://github.com/ethereum/tests/blob/4d539de6788147a0376b70f2bc71a8f818872164/src/BlockchainTestsFiller/InvalidBlocks/bcEIP1559/gasLimit40mFiller.yml#L55-L57):

```
  - blockHeader:
      gasLimit: 39960938
    transactions: []
```

Block gasLimit: 39960938
Min gasLimit: 39960938

qbzzt:
> Update from EthereumJS: state tests are now passing (it was within our own code)

Good. That means the tests were useful.


> For the Blockchain Tests, I think these failures are due to outdated versions of the spec that test `>=` and `<=` rather than `>` and `<`? ([PR](https://github.com/ethereum/EIPs/pull/3566))

The tests look at exactly the border line between permitted and not permitted (for example, fail at ...247, succeed at ...248):

```
  - blockHeader:
      gasLimit: 1072693247
    expectException:
      '>=London': 1559BlockImportImpossible_TargetGasLow
    transactions: []
  - blockHeader:
      gasLimit: 1072693248    # 1024*1024*1023 (1023/1024 of the previous gasLimit)
    transactions: []
```

If the borderline was different, even by one gas to either direction, the `InvalidBlocks/bcEIP1559/badBlocks` would have failed on geth. 

To verify I downloaded the latest geth and tried badBlocks on it. It still works.

Can you run the test with verbose output, and see on which block it has an error? 

   Thanks,
            Ori

ryanio:
yes they look good!

I quoted the suspect blocks in my above comment, let me know if there is a better way to identify them

qbzzt:
> yes they look good!
> 
> I quoted the suspect blocks in my above comment, let me know if there is a better way to identify them

OK, sorry I didn't realize that.

```
   - blockHeader:
      gasLimit: 1072693248    # 1024*1024*1023 (1023/1024 of the previous gasLimit)
    transactions: []
```

This block is supposed to be accepted (that's why it doesn't have an `expectException:` field). Does it fail? If so, what is the exception? If not, does the block with a gas limit of 1072693247 fail? If so, what exception is reported to retesteth?

It's possible that the only problem is that retesteth doesn't recognize the exception.

ryanio:
np, yes it's failing with our internal validation of `invalid gas limit` as it can't be exactly equal (https://github.com/ethereum/EIPs/pull/3566)

PR #839, opened 2021-05-10
Generate London cases for stTransactionTest/Opcodes_TransactionInit

gumb0:
This doesn't add any new cases, but generates London expectations for `stTransactionTest/Opcodes_TransactionInit` test using recent go-ethereum master.

It is supposed to help make changes in https://github.com/ethereum/tests/pull/835 smaller.

### DIFF
--- a/JSONSchema/st-schema.json
+++ b/JSONSchema/st-schema.json
@@ -238,6 +238,9 @@
                         },
                         "Berlin": {
                             "$ref": "#/definitions/TransactionResults"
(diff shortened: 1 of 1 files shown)


gumb0:
In discussion @winsvega he pointed out that he's going to regenerate all the tests to include London expectations at some point, probably after main EIPs for London (I guess at least EIP-1559) are merged into geth.

This PR can be closed then, if not useful.

PR #841, opened 2021-05-10
Updated testOpcodes for London:

qbzzt:
1. Remove Istanbul from the forks to test list
2. Added opcode 48 (basefee), starting at London

### DIFF
--- a/src/BlockchainTestsFiller/ValidBlocks/bcStateTests/testOpcodesGen.js
+++ b/src/BlockchainTestsFiller/ValidBlocks/bcStateTests/testOpcodesGen.js
@@ -24,6 +24,9 @@
 // Maximum gas we're allowed to use
 const gasLimit = 'F00000000000'
 
+// "User"'s gas balance
+const userBalance = 'F000000000000000'
+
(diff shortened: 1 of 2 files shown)


src/BlockchainTestsFiller/ValidBlocks/bcStateTests/testOpcodesGen.js
```diff
@@ -109,6 +109,7 @@ const boilerPlate_Head = `
     timestamp: 0x54c98c81
     transactionsTrie: 0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421
     uncleHash: 0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347
+    baseFee: 0x01
```
winsvega: I think we better use value of 10 as by the rules it cant go below 7 so anything below 7 might be marked illigal

qbzzt: Done.

winsvega (approved):

PR #842, opened 2021-05-11
Init code with invalid first byte

gumb0:
Pulled out of #835


gumb0:
Included London expectations in blockchain tests by mistake, will remove.

gumb0:
> Included London expectations in blockchain tests by mistake, will remove.

Removed.

PR #843, opened 2021-05-11
CREATE/CREATE2 a contract with all possible first bytes

gumb0:
Pulled out of #835 removing London expectations, should be useful on its own to check that deploying all undefined bytes is allowed in Berlin.


src/GeneralStateTestsFiller/stCreate2/CREATE2_FirstByte_loopFiller.yml
```diff
@@ -0,0 +1,59 @@
...
+
+  expect:
+    - indexes:
+       data: !!int -1
+       gas: !!int -1
+       value: !!int -1
+      network:
+        - 'Berlin'
+      result:
+       a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
```
winsvega, line 54: can add a few addresses with deployed code here to show and to see that accounts with that bytecode has actually been deployed

src/GeneralStateTestsFiller/stCreate2/CREATE2_FirstByte_loopFiller.yml
```diff
@@ -0,0 +1,59 @@
...
+      balance: 100000000
+      code: ''
+      nonce: 0
+      storage: {}
+    b94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+      balance: 0
+      code: |
+        :yul
+        {
+          // { mstore8(0, 0x00) return(0, 1) }
```
winsvega: this is the contract constructor code.
so you need actually to return the bytecode of a contract that is being created. here it is 0x00

or do you try to create2 a contract with constructor code with invalid code before return?

gumb0: Yes, intention is to create contracts with code that consists of one byre

src/GeneralStateTestsFiller/stCreate2/CREATE2_FirstByte_loopFiller.yml
gumb0: Added

src/GeneralStateTestsFiller/stCreate2/CREATE2_FirstByte_loopFiller.yml
gumb0: Added some comments in yul code.

PR #844, opened 2021-05-14
Check for correct rollback of init code with side effects

gumb0:
This adds a test that checks that exceptional abort during init code correctly rolls back side effects.

The transaction `CALL`s a contract that `SSTORE`s `0 -> 1`, then transaction ends with `INVALID`. Storage change in a contract is expected to be rolled back.

(In #835 I'm adding a similar test that rolls back after exceptional abort due to trying to deploy `0xef` byte in London.)


src/GeneralStateTestsFiller/stTransactionTest/Opcodes_TransactionInitFiller.json
```diff
@@ -323,7 +348,8 @@
                 ":raw 0x6000600060006000730f572e5295c57f15886f9b263e2f6d2d6c7b5ec6612710fa5060006000f3", 
                 ":raw 0x60006000fd60006000f3", 
                 ":raw 0x32ff",
-                ":label invalid_first_byte_ef :raw 0xef"
+                ":label invalid_first_byte_ef :raw 0xef",
+                ":label side_effects_invalid_opcode :yul { pop(call(50000, 0xb94f5374fce5edbc8e2a8697c15331677e6ebf0b, 0, 0, 0, 0, 0)) invalid() }"
```
winsvega, line 374: invalid is like revert?

winsvega: here is tricky to know that the call worked. 
if its not requred to call revert directly in transaction init data field. 
the way to do it is to go subcall. 

transaction init ->   call ->  do smth.  revert.  |  continue execution. sstore smth. 

this way revert should cause only a subcall going oog and we can confirtm that execution happend by putting a sstore at the end of transaction init code.

winsvega: or you can have a control transaction with exactly the same data,  but this time without invalid() instruction at the end. 
so we check that the call (50000, ...  worked as at some point 50000 might be not enough and it will just go OOG not testing what we wanted

gumb0: I don't want to test here reverts inside subcalls, becasue this file is all about init code in the creation transaction.
(Although other similar tests to check reverts inside CREATEs might be useful.)

So I prefer to add another similar transaction with the same side effect, but without revert.

gumb0: Added.

src/GeneralStateTestsFiller/stTransactionTest/Opcodes_TransactionInitFiller.json
```diff
@@ -323,7 +369,9 @@
                 ":raw 0x6000600060006000730f572e5295c57f15886f9b263e2f6d2d6c7b5ec6612710fa5060006000f3", 
                 ":raw 0x60006000fd60006000f3", 
                 ":raw 0x32ff",
-                ":label invalid_first_byte_ef :raw 0xef"
+                ":label invalid_first_byte_ef :raw 0xef",
+                ":label side_effects :yul { pop(call(50000, 0xb94f5374fce5edbc8e2a8697c15331677e6ebf0b, 0, 0, 0, 0, 0)) }"
```
gumb0, line 373: This was missing a comma in the end of the line, how did it work and didn't complain? @winsvega 

Did it consider this a concatenation of two strings?

winsvega: yes apparently I can parse without `,`

winsvega: this PR however fails on master geth. which version did you fill it with?

gumb0: > this PR however fails on master geth. which version did you fill it with?

It was this commit according to generated test
https://github.com/ethereum/go-ethereum/commit/addd8824cf3ad6133c1b1bbc3387a621eafba6a3

gumb0: The current `develop` version of this test passes for me with latest geth https://github.com/ethereum/go-ethereum/commit/017cf71fbd51ede510dd0e40e02f89b5340eea93

PR #846, opened 2021-05-16
More eip1559 documentation:

qbzzt:
1. Added an stExample file for EIP1559
2. Changed the field names in some docs to the new ones


winsvega:
ok. sorry I removed the label. now we can start merging London stuff. 
resolve the conflict

PR #847, opened 2021-05-18
Tests for EIP-3540: EVM Object Format (EOF) v1

gumb0:
Tests for [EIP-3540: EVM Object Format (EOF) v1](https://eips.ethereum.org/EIPS/eip-3540).

Geth implementation is currently in an open PR at https://github.com/ethereum/go-ethereum/pull/24090.


src/GeneralStateTestsFiller/stTransactionTest/CreateTransactionEOF1Filler.yml
```diff
@@ -139,3 +141,42 @@ CreateTransactionEOF1:
...
+         nonce: 1
+         balance: 30000
+       6295ee1b4f6dd65047762f924ecd367c17eabf8f:
+         shouldnotexist: 1
+    - indexes:
+       data: ':label deploying_valid_eof1'
+       gas: !!int -1
+       value: !!int -1
+      network:
+        - 'London+3540'
```
gumb0: @winsvega Weird that it passes CI currently, although it contains this invalid network.

winsvega: CI checks that tests are filled and that the filler hash in the filled test match the hash of Filler file
I disabled python checks

gumb0:
I removed London milestone from PR, because it's not proposed for London.

holgerd77:
Hi there, we are currently working on an implementation of EIP-3540 for the EthereumJS VM here https://github.com/ethereumjs/ethereumjs-monorepo/issues/1704 and - TBH - we have no idea what this `data` section in the EIP even is respectively what we should do with this along code execution? 🤔

Can someone enlighten us and give some concrete example here? This is also kept very much implicit in the EIP https://eips.ethereum.org/EIPS/eip-3540, I guess the EIP itself could also benefit on some more concrete specification and/or context here.

If it plays a role - and is not only a placeholder for some future usages - it would e.g. be good if it would be mentioned in the [Execution Semantics Changes](https://eips.ethereum.org/EIPS/eip-3540#changes-to-execution-semantics) section.

Thanks a lot! 😄

gumb0:
Hi, thanks for the feedback!

It's not a placeholder, it's a part of bytecode that can be useful for a contract to put any data blobs into it. 
From the point of view of execution these bytes never can be executed, but can be accessed with `CODECOPY`/`EXTCODECOPY` (which takes offset absolute in the entire EOF bytecode, so you can point it to copying something from data section.) `CODESIZE`/`EXTCODESIZE` also takes into account this section, returning size of entire EOF bytecode. `EXTCODEHASH` returns hash of entire EOF bytecode, too.

gumb0:
I'll push some examples of codes with data section into this PR soon.

holgerd77:
@gumb0 thanks, that helps a lot! 🙏

BlockchainTests/GeneralStateTests/stCallCodes/EOF1_ContractCalls.json
```diff
@@ -0,0 +1,10801 @@
...
+            "0x104f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
+                "balance" : "0x00",
+                "code" : "0xef000101001d020004006019600160003960005160005500aabbccdd",
+                "nonce" : "0x00",
+                "storage" : {
+                }
+            },
+            "0x114f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
+                "balance" : "0x00",
+                "code" : "0xef0001010041020004007fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff6000526008603c60003960005160005500aabbccdd",
```
acolytec3: Is this `code` correct?  This particular test is failing on the EthereumJS branch implementing EIP3540 and it appears that the `code` section of the header (`010041`) indicates a section length of 65 which is longer than the entire bytecode that is intended to be stored in this account, which is only 64 bytes.

gumb0: You're right, this is invalid, thanks.

acolytec3: Thanks!  I thought I was losing my mind looking at this bytecode trying to figure out what was wrong.

gumb0: Sorry about this! I expect many changes in this PR still, so don't assume it's free of dumb errors :grin:

acolytec3: No worries and thanks!

acolytec3:
I can't get a specific code link because of the size of the raw file but is this test valid -- `CreateTransactionInvalidEOF1_d9g0v0_London+3540+3670`?
The `gasUsed` in the post state in this test is recorded as 53040,  When I run this test with our VM, it computes a transaction base fee of 53040 but then our VM does an exceptional abort and reverts the CREATE message call and uses all remaining gas since the bytecode of the contract being created starts with `0xef` and is invalid EOF code, so our total gas used is 70000 for the transaction.  Just trying to decipher whether it's my code or something else since I'm trying to get all of these tests to pass with our implementation of EIP3540/3670.

gumb0:
@acolytec3 `CreateTransactionInvalidEOF1_d9g0v0_London+3540+3670` is sending creation transaction with init code `0xef000100` which is malformed EOF, so execution doesn't start and only transaction intrinsic gas should be consumed.

jochem-brouwer:
@gumb0 could you clarify in the EIP what "exceptional abort" means, especially in the gas context? So in the create transaction which sends malformed EOF, only consume intrinstic gas. What happens in a CREATE/CREATE2 opcode? Do we do the same (i.e. consume the base gas for CREATE(2) + memory + hash cost etc.) or do we consume all gas sent along with CREATE?

gumb0:
> @gumb0 could you clarify in the EIP what "exceptional abort" means, especially in the gas context? So in the create transaction which sends malformed EOF, only consume intrinstic gas. What happens in a CREATE/CREATE2 opcode? Do we do the same (i.e. consume the base gas for CREATE(2) + memory + hash cost etc.) or do we consume all gas sent along with CREATE?

If initcode is malformed or invalid, initcode execution gas is not consumed. If returned code is malformed or invalid, then all creation gas is consumed.

These failure cases are mentioned in EIP-3670 spec, but without gas rules. 

For now I've suggested a clarification in EIP-3670 https://github.com/ethereum/EIPs/pull/4833/commits/a7f45611cd43ad4d7aa1f10f063a5f6df59a33d4 We'll think how to clarify this in EIP-3540, too.

gumb0:
I think tests are mostly done here and ready for review.

(except fork names would need to be replaced at some point: `London` => `Merge`, `London+3540+3670` = > `Shanghai`)

GeneralStateTests/stCreateTest/CREATE_EOF1_FromEOF.json
```diff
@@ -0,0 +1,216 @@
...
+                },
+                {
+                    "hash" : "0xc61814b3888f35a254cda924d7515af20203bf3e355d6a6fa119c2800aa9670e",
+                    "indexes" : {
+                        "data" : 12,
+                        "gas" : 0,
+                        "value" : 0
+                    },
+                    "logs" : "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
+                    "txbytes" : "0xf865800a83e4e1c094b94f5374fce5edbc8e2a8697c15331677e6ebf0b808560016000f31ba08ded9ae1fd850114385f132b1e74d4e123627865bf35a8ccdd1571c4f019b4c4a005536793e1483242976ce12f1a4e9724391d86f23c8d3bad310fa63d636c3ad0"
```
acolytec3, line 168: According to our implementation, the computed size of the EOF container stored in 0x174f5374fce5edbc8e2a8697c15331677e6ebf0b that is called in this transaction is 126 while the actual bytecode length is 128 so this test is failing in the EthereumJS implementation.

GeneralStateTests/stCallCodes/EOF1_Execution.json
```diff
@@ -0,0 +1,1048 @@
...
+                },
+                {
+                    "hash" : "0xef57395b4856eac81327501e8f4b41d1dde23dc1410a47b6b3a750dc7defa675",
+                    "indexes" : {
+                        "data" : 31,
+                        "gas" : 0,
+                        "value" : 0
+                    },
+                    "logs" : "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
+                    "txbytes" : "0xf874800a830aae6094b94f5374fce5edbc8e2a8697c15331677e6ebf0b8094654f5374fce5edbc8e2a8697c15331677e6ebf0b1ba08b6504c574b41e0d661a61accc0f42ea8c5672fa6911bb76c5f4e42eecca553aa049a5c5fc75958c321ffec2733712582156027b649c3d3aeee5f6becdfa79eeca"
```
acolytec3, line 721: As above, this test is failing in the EthereumJS implememtation due to the code stored in 0x654f5374fce5edbc8e2a8697c15331677e6ebf0b having a computed container size of 43 bytes while the actual bytecode length is 42.

acolytec3 (commented):
See my specific comments.  Two of the tests in the `EOF1_Execution` test file are failing in the EthereumJS implementation.

GeneralStateTests/stCreateTest/CREATE_EOF1_FromEOF.json
gumb0: This sends transaction to contract `0xb94f5374fce5edbc8e2a8697c15331677e6ebf0b`, which creates contract `0xf1ecf98489fa9ed60a664fc4998db699cfa39d40` with 1-byte-long code.
(are you sure problem is in this test file (`CREATE_EOF1_FromEOF`)?)

GeneralStateTests/stCallCodes/EOF1_Execution.json
gumb0: In this case found the problem, you're right, code is invalid, will regenerate the test.

(We should work on making geth test generation tool reject invalid EOF in pre-state, then I'll catch all such problems)

gumb0: This case should be fixed now.

gumb0:
Rebased.

acolytec3:
The [code](https://github.com/ewasm/tests/blob/afe4baf505f81ea990094dfd40ff6ac17fab9051/BlockchainTests/GeneralStateTests/stCallCodes/EOF1_Execution.json#L4553) in the account - 0x174f5374fce5edbc8e2a8697c15331677e6ebf0b appears to be invalid EOF for this test `EOF1_Execution_d12g0v0_London+3540+3670` as the computed code size for the container should be 126 bytes and the actual length of the byte string looks like its 128 bytes.  Can you verify?

jochem-brouwer:
I have an interesting test case which would only be interesting if you activate 3540, 3670 AND 3860 (limit and meter initcode). If you deploy an EOF contract which has a code section which is smaller than the max initcode (per 3860), but the entire EOF container is larger than the initcode, what should happen? (Will also raise this on magicians)

holiman:
@jochem-brouwer here's my answer  (not authorative - just based on my reading the EIP. So I'm curious to see if my interpretation is correct) 

From 3540 re the size of code:

> CODECOPY/CODESIZE/EXTCODECOPY/EXTCODESIZE/EXTCODEHASH keeps operating on the entire container.

So essentially, according to 3540, the size of the code, as evm observable, is "the entire container". Hence, if the size of the entire is larger than max initcode / max code, then it should be rejected. 

It is a good question, and the answer should be lifted into the relevant EIP(s) for clarification.

jochem-brouwer:
@holiman That seems like a reasonable interpretation to me!

jochem-brouwer:
@gumb0 I can verify that we had a bug (which was not tested by this test suite):

It is most likely a niche-problem due to our implementation, but we have a problem if we have a correct header EXCEPT we do not have the 0xEF byte at the zero index of the contract.

Could you try creating a test which has an almost valid header, except the first byte?

E.g. `ED000101000600600160015500`

This is an invalid header (byte 0 is not 0xEF). However, the remainder of the header is all correct. The code itself is simple; `PUSH 1 PUSH 1 SSTORE STOP`, so this puts 1 in slot 1. If you change it to `EF000101000600600160015500` then the code is valid and it should store. We had a bug where the 0xED code actually ran the code as if it was a valid header (note: when CALLing the contract, not CREATEing it). We fixed this in https://github.com/ethereumjs/ethereumjs-monorepo/pull/2381

src/GeneralStateTestsFiller/stCallCodes/EOF1_ExecutionFiller.yml
```diff
@@ -0,0 +1,1375 @@
...
+      code: ':raw 0xEF000101000B02006500600456FE5B600160005500FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF'
+      nonce: 0
+      storage: {}
+
+    174f5374fce5edbc8e2a8697c15331677e6ebf0b:
+      balance: 0
+      # Data section is 0x67 (103) bytes long, so header contains 0x67 (PUSH8) byte,
+      # but it must not affect jumpdest analysis (JUMPDEST stays valid)
+      # code section: 'push1(1) jumpi x invalid x: sstore(0, 1) stop()'
+      code: ':raw 0xEF000101000D020067006001600657FE5B600160005500FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF'
```
chfast: This code is invalid: the data section is longer by 2 bytes from expected, remove `FFFF`.

gumb0:
> The [code](https://github.com/ewasm/tests/blob/afe4baf505f81ea990094dfd40ff6ac17fab9051/BlockchainTests/GeneralStateTests/stCallCodes/EOF1_Execution.json#L4553) in the account - 0x174f5374fce5edbc8e2a8697c15331677e6ebf0b appears to be invalid EOF for this test `EOF1_Execution_d12g0v0_London+3540+3670` as the computed code size for the container should be 126 bytes and the actual length of the byte string looks like its 128 bytes. Can you verify?

This should be fixed now.

src/GeneralStateTestsFiller/stCallCodes/EOF1_ExecutionFiller.yml
gumb0: Fixed.

gumb0:
> I have an interesting test case which would only be interesting if you activate 3540, 3670 AND 3860 (limit and meter initcode). If you deploy an EOF contract which has a code section which is smaller than the max initcode (per 3860), but the entire EOF container is larger than the initcode, what should happen? (Will also raise this on magicians)

Our intention certainly was to apply code limit size of 3860 to entire container.

gumb0:
I updated expectations to be based on `Merge` fork
(`London+3540+3670` => `Merge+3540+3670`)

gumb0:
Removed commented out sections from `EOF1_ExecutionFiller.yml`.

gumb0:
Rebased.

winsvega:
Lets merge the tests. Following updates can be in additional PRs. This one got big already.

axic:
Was the [test case](https://github.com/ethereum/tests/pull/847#issuecomment-1291827144) from @jochem-brouwer included?

gumb0:
> Was the [test case](https://github.com/ethereum/tests/pull/847#issuecomment-1291827144) from @jochem-brouwer included?

It was not. So far I only noted it for later in https://github.com/ethereum/tests/issues/1100

PR #851, opened 2021-05-23
Template for multi-context tests

qbzzt:
Used that template to create stEIP1559/gasprice, which checks GASPRICE
in a large number of contexts (direct, CALL, CREATE and then CALL, etc)


qbzzt:
Got mixed up with #837 stuff. I am going to close this and create a different PR

PR #853, opened 2021-05-25
Test to check the initial values of EIP 1559 parameters

qbzzt:
It also verifies when transaction types 0-2 are supported (London)
and when only 0-1 (Berlin)


qbzzt:
Not ready to merge because the value of gasLimit changed in the EIP. It is now:

```
		if INITIAL_FORK_BLOCK_NUMBER == block.number:
			parent_gas_target = self.parent(block).gas_limit
			parent_gas_limit = self.parent(block).gas_limit * ELASTICITY_MULTIPLIER 
```

And that isn't supported by geth yet so I can't run tests for it.

winsvega:
again this are invalid transactions check. would need to copy those that possible into state test for martin. 

as I don't think hive could check transaction invalidity from blockchain tests (from state test either) 
but devs can check transaction invalidity from state tests as it is easier to execute.

lets wait for invalidity exceptions from t8ntool 
@holiman

holiman:
Coming in this PR: https://github.com/ethereum/go-ethereum/pull/22976/commits/913549aad1f2918699122f5ad60bb3ca29c67a50

winsvega:
implemented in retesteth

src/BlockchainTestsFiller/TransitionTests/bcBerlinToLondon/initialValFiller.yml
```diff
@@ -0,0 +1,480 @@
...
+
+    # This type 0 transaction has gasPrice < baseFee. It won't happen
+    # Invalid transactions mess up the nonce calculation, so I prefer to do them
+    # with otherwise unused accounts  
+    - data: :abi f(uint) 0x6060
+      expectException:
+        BerlinToLondonAt5: TR_FeeCapLessThanBlocks
+      gasLimit: 1000000
+      gasPrice: 875000043
+      nonce: auto
```
winsvega, line 278: the nonce auto would be tricky logic here. because it depends on expectException. 
so I know if the transaction is expected to be rejected then the nonce should not increment. 
but it is not obvious if a client accept the transaction, then the next one will be rejected because of nonce. 

I would rather use explicit nonce. 

but I feel more safe now as we have exception check and we can see that it is not related to nonce. in case somthing mess up

src/BlockchainTestsFiller/TransitionTests/bcBerlinToLondon/initialValFiller.yml
```diff
@@ -0,0 +1,480 @@
...
+    # This type 1 transaction has gasPrice < baseFee. It won't happen
+    # Invalid transactions mess up the nonce calculation, so I prefer to do them
+    # with otherwise unused accounts  
+    - data: :abi f(uint) 0x6090
+      expectException:
+        BerlinToLondonAt5: TR_FeeCapLessThanBlocks
+      gasLimit: 1000000
+      gasPrice: 875000043
+      nonce: auto
+      secretKey: 0x69d66403e6a54b18c2a768af4a8de554df133550d4dfa37a377ec8c0e85b9ef4
```
winsvega, line 314: so the nonce auto wont work here if it was the same secretKey as in valid transactions

qbzzt: It won't work if the same secretKey is used in valid transactions (or possibly invalid ones) later. If this is the last transaction with that secretKey then there's no problem.

PR #859, opened 2021-05-30
Check intrinsic costs of transactions:

qbzzt:
21000 + 16 * non-zero calldata bytes + 4 * zero calldata bytes +
  1900 * access list storage key count + 2400 * access list address count

### DIFF
--- a/BlockchainTests/ValidBlocks/bcEIP1559/intrinsicTip.json
+++ b/BlockchainTests/ValidBlocks/bcEIP1559/intrinsicTip.json
@@ -1,0 +1,2856 @@
+{
+    "intrinsicTip_London" : {
+        "_info" : {
+            "comment" : "",
+            "filling-rpc-server" : "evm version 1.10.4-unstable-966ee3ae-20210528",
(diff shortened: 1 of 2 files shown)


src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/intrinsicFiller.yml
```diff
@@ -0,0 +1,2327 @@
...
+
+
+  blocks:
+  
+  - transactions:
+    # Verify gas cost
+    - data: ''
+      accessList: []
+      gasLimit: 500000	
+      maxPriorityFeePerGas: 0x00
```
winsvega, line 685: one more check to see that maxPriority (tip) does not affect formula calculation

qbzzt: The formula I use to get the formula result depends on having the same gasprice. So it's simplest to create another test file with the same tip for every transaction.

Done.

PR #861, opened 2021-05-31
Added another test for intrinsic costs

qbzzt:


src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/intrinsicOrFailFiller.yml
```diff
@@ -0,0 +1,3747 @@
...
+        storageKeys:
+        - 0x600D
+        - 0x0BAD
+        - 0x60A7
+        - 0xBEEF
+      invalid: 1
+      gasLimit: 129631	
+      maxPriorityFeePerGas: 0x00
+      maxFeePerGas: 100000
+      nonce: auto
```
winsvega: nonce auto does not work with invalid yet

winsvega (changes requested):
you use nonce:auto and invalid indicators together. 
we didn't fix this issue yet

qbzzt:
> you use nonce:auto and invalid indicators together.
> we didn't fix this issue yet

The problem only happens when you use the account after it sent an invalid transaction. In this case, I created sixty accounts: https://github.com/ethereum/tests/blob/558a50f575e2215fb83b891c4cc81d113022d75e/src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/intrinsicOrFailFiller.yml#L44-L584 and used each one only for a valid transaction followed by an invalid transaction (https://github.com/ethereum/tests/blob/558a50f575e2215fb83b891c4cc81d113022d75e/src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/intrinsicOrFailFiller.yml#L592-L613). As long as you don't have other transactions that use the same account subsequently in the test, it's not a problem.

winsvega:
now this whole topic is delayed. I wait to get the reason of transaction rejection in json format so can have expect sections for invalid transaction failures

winsvega:
ok now we have it in blockchain. but also we need it in state tests
(transaction rejection check) 

besu catch up!

BlockchainTests/ValidBlocks/bcStateTests/TransactionToItselfNotEnoughFounds.json
```diff
@@ -125,6 +125,30 @@
                         "value" : "0xc351"
                     }
                 ],
+                "transactionSequence" : [
```
winsvega, line 128: I think you pulled my changes here 
this are already in develop

qbzzt: Oops, yes. I'll just close this PR and create a new, clean one.

PR #873, opened 2021-06-05
Added transaction exceptions

qbzzt:
Should be merged after https://github.com/ethereum/retesteth/pull/133. Ready to merge once that PR is merged.


winsvega:
this cases we would need also to check in state tests

PR #877, opened 2021-06-07
Added a state test for intrinsic costs

qbzzt:
The test filler is created by a generator (the JavaScript file)


src/GeneralStateTestsFiller/stEIP1559/intrinsicGen.js
```diff
@@ -0,0 +1,218 @@
...
+    maxFeePerGas: 1000
+    maxPriorityFeePerGas: 100
+`
+
+
+// data for test cases
+const dataList = [
+    {gas:   0,     data: '0x' },
+    {gas:   4,     data: '0x00' },
+//  {gas:  16,     data: '0x00000000' },
```
winsvega, line 60: is this the test with 10k cases?

qbzzt: No, the one with 1440 cases. If you remove the comments in `dataList` and `accessListList`, then it's 10k.

winsvega: sure no need for expectSection transaction only flag?

qbzzt: Do you mean `expectException`? That is already there, it just gets created later in the file. Or do you mean something else?

winsvega: I mean generate transactions that only covered with expectSection. to reduce the number of combinations

qbzzt: I tried that, but when a combination is not covered in the expect section I get an error:
```
Error: Test has transaction uncovered with expect section! (stEIP1559/intrinsic, fork: all, TrInfo: d: 28, g: 55, v: 0)
```

Do you want me to do it anyway? If so, can we have a field in `env` that tells retesteth "some combinations are not covered by expect, just don't run them"?

PR #881, opened 2021-06-11
baseFee to baseFeePerGas

winsvega:


holiman:
Why did you change that? The term is `baseFee` ? It's in the EIP itself and, afaik, every client implementation.

holgerd77:
We have actually also [named](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/header.ts#L46) this `baseFeePerGas` following changes on the RPC spec like https://github.com/ethereum/eth1.0-specs/pull/192 and in the spec it is also named `base_fee_per_gas`, see https://eips.ethereum.org/EIPS/eip-1559.

Do I miss something here? 🤔

holiman:
> Do I miss something here? thinking

Nope, apparently it was me who missed something. Apologies!

PR #882, opened 2021-06-11
Template tests: replacement to PR #856.

qbzzt:
This includes all the scripts, and all the tests except for the `stBadOpcode/opc<n>Template` ones

PR #856, opened 2021-05-27
Template creator and three tests that use it

qbzzt:
1. `GASPRICE` in multiple contexts
2. `SELFBALANCE` in multiple contexts
3. An invalid opcode (0xFE) in multiple contexts

The contexts are:

//  0. 0x0000  Normal execution
//  1. 0x00F1  CALL a contract with the code
//  2. 0x00F2  CALLCODE a contract ...
//  3. 0x00F4  DELEGATECALL a contract ...
//  4. 0x00FA  STATICCALL a contract ...
//  5. 0xF1F1  CALL a contract that CALLs a contract ...
//  6. 0xF2F1  CALLCODE a contract that CALLs ...
//  7. 0xF4F1  DELEGATECALL ... CALL
//  8. 0xFAF1  STATICCALL ... CALL
//  9. 0xF1F2  CALL a contract that CALLCODEs a contract ...
// 10. 0xF2F2  CALLCODE a contract that CALLCODEs ...
// 11. 0xF4F2  DELEGATECALL ... CALLCODE
// 12. 0xFAF2  STATICCALL ... CALLCODE
// 13. 0xF1F4  CALL a contract that DELEGATECALLs a contract ...
// 14. 0xF2F4  CALLCODE a contract that DELEGATECALLs ...
// 15. 0xF4F4  DELEGATECALL ... DELEGATECALL
// 16. 0xFAF4  STATICCALL ... DELEGATECALL
// 17. 0xF1FA  CALL a contract that STATICCALLs a contract ...
// 18. 0xF2FA  CALLCODE a contract that STATICCALLs ...
// 19. 0xF4FA  DELEGATECALL ... STATICCALL
// 20. 0xFAFA  STATICCALL ... STATICCALL
// 21. 0x00FD  Run the code, call a contract that reverts, then run again
// 22. 0x00FE  Run the code, call a contract that goes out of gas, then run again
// 23. 0x00FF  Run the code, call a contract that self-destructs, then run again
// 24. 0x00F0  CREATE a contract, run the code in the constructor
// 25. 0x00F5  CREATE2 a contract, run the code in the constructor
// 26. 0xF0F1  CREATE a contract with the code and then CALL it
// 27. 0xF5F1  CREATE2 a contract with the code and then CALL it
// 28. 0x60BACCFA57 CALL 1023 times and then run the code at the very end of the call stack

PR #886, opened 2021-06-17
Template Tests

### LINKED ISSUE #845: Some ideas for more tests with invalid bytes
1. Creation transactions deploying a contracts with all possible values in a first byte, similar to tests for `CREATE` in https://github.com/ethereum/tests/pull/843 (probably will require a blockchain test for this)

2. Init codes starting with a byte with all possible values
2.1. In creation transaction (one invalid byte value is checked in https://github.com/ethereum/tests/pull/842)
2.2. in the code passed to CREATE 
2.3 in the code passed to CREATE2

### LINKED ISSUE #848: more tests around SELFBALANCE
we have a contribution but I think this needs to be covered more

btw. lllc supports it

qbzzt:
Including tests for all the invalid opcodes (there are about a hundred of them).

This replaces #882 (and #856). Dimitry's requests from today are added.


winsvega:
does this have an update after our call discussion?

qbzzt:
> does this have an update after our call discussion?

Yes, it has everything we discussed.

qbzzt:
It was just easier to start a new PR

holiman:
See https://github.com/ethereum/tests/blob/develop/BlockchainTests/GeneralStateTests/stBadOpcode/invalidDiffPlaces.json#L28 

```
            "uncleHash" : "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
            "baseFee" : "0x0b"
        },
```
Should be `baseFeePerGas`, not `baseFee`. So a lot of these tests fail on go-ethereum on hive due to that.

PR #892, opened 2021-06-30
Added the ommer (uncle) test tutorial

qbzzt:


docs/ommer-tutorial.rst
```diff
@@ -0,0 +1,97 @@
...
+includes `ommer/uncle blocks <https://ethereum.org/en/glossary/#ommer>`_,
+blocks that are not part of the main chain but still deserve a reward.
+
+
+Uncle Blocks
+============
+Uncle blocks are created because there are multiple miners working at any point
+in time. If two miners propose a follow-up block for the chain, only one of them
+becomes the real block, but the other miner who did the same work also deserves
+a reward (otherwise the chances of getting a mining reward will be too low, and there will
```
winsvega, line 18: this is also done to enhance blockchain securinty.

docs/tutorial_samples/12_ommerGoodFiller.yml
```diff
@@ -0,0 +1,73 @@
...
+  - blocknumber: 2
+    transactions: []
+
+  - blocknumber: 3
+    transactions: []
+
+  - blocknumber: 4
+    transactions: []
+    blockHeader:
+      stateRoot: 0x73c5e0631444838bfc2d8368cc4c11a15596de9b586606ebdfe8ba03c45f1039
```
winsvega, line 44: this is a state root we need to overwrite because now the state has a mining reward for the uncles 
since we don't have a protocol that asks a client to generate a block with uncles. 
we need to correct this on retesteth side as retesteth constructing the block with uncles but it can't calculate the post state hash

winsvega (approved):

PR #893, opened 2021-06-30
remove code from sender account

winsvega:
remoe code from sender account


holiman:
I think the legacytests also need to be refilled, since this is a retroactive change

PR #896, opened 2021-07-14
Created a timestamp test

qbzzt:
Note that this test can only be filled on geth. However, once --filltests is run on geth
the filled test can run in besu

Tests three things:
1. Timestamp equal to parent block fails
2. Timestamp prior to parent block fails
3. Timestamp one microsecond above the parent block is successful

All of these tests come from equation (49) in the yellow paper.


winsvega:
Don't use geth t8ntool as it is retesteth backend that performs the check.

We need to get the exception scenarios from besu implementation

src/BlockchainTestsFiller/InvalidBlocks/bcInvalidHeaderTest/badTimestampFiller.yml
```diff
@@ -0,0 +1,91 @@
...
+      timestamp: 1000000
+    blocknumber: 4
+    transactions: []
+    expectException:
+      '>=London': InvalidTimestampOlderParent
+
+
+  # This block is legitimate
+  - blockHeader:
+      timestamp: 4000000
```
winsvega: Need also check would it be valid if
4000001, 4000002 or it will throw timestamp  diff is too short
At what minimum timestamp diff it is accepted?

And at what distance it is considered to be too far in the future to be accepted?

qbzzt: If the timestamp diff is 8 microseconds or less it fails with a difficulty error in the log. However, that error is not propagated to the RPC, so we can't check for it. I asked on Telegram for this to be fixed.

The maximum timestamp diff seems to be variable. I ran the same test four consecutive times, and each time it gave me in the log a different maximum timestamp difference.

[badTimestampFiller.yml.txt](https://github.com/ethereum/tests/files/6834831/badTimestampFiller.yml.txt)
[timestamp-maxdiff.log.txt](https://github.com/ethereum/tests/files/6834832/timestamp-maxdiff.log.txt)

qbzzt: For what it is worth, the maximum difference seems to grow by 1 ms every second.

BlockchainTests/InvalidBlocks/bcInvalidHeaderTest/badTimestamp.json
```diff
@@ -0,0 +1,234 @@
+{
+    "badTimestamp_London" : {
+        "_info" : {
+            "comment" : "",
+            "filling-rpc-server" : "evm version 1.10.6-unstable-f0b1bdda-20210714",
```
winsvega, line 5: does it pass on besu? since the timestamp check and difficulty done on retesteth which might be wrong on extreme cases

qbzzt: No, it doesn't. I can't run this exact test on besu (because it have invalid blocks and we can't get that on besu yet), but when I ran it without the invalid blocks and difficulties it failed on the exact same issue:

```
2021-07-20 17:19:28.680+00:00 | vert.x-worker-thread-13 | INFO  | CalculatedDifficultyValidationRule | Invalid block header: difficulty 131072 does not equal expected difficulty 131136

```

PR #898, opened 2021-07-22
1559 transaction with value bug

winsvega:


holiman:
LGTM. 
Without the fix: 
```
    "error": "post state root mismatch: got ada2ae7e9aa5f645994d18304bd376f56738da00bce3c758d292b87029e88fbe, want e437f23dde868903471ed39c52d5ddf14e0415c9db185d4f20f5c9de0f5209e8"
```
With the fix, it passes

PR #901, opened 2021-07-23
State test for when the value causes a transaction to be out of funds

qbzzt:


src/GeneralStateTestsFiller/stEIP1559/valCausesOOFFiller.yml
```diff
@@ -0,0 +1,112 @@
...
+      accessList: []
+
+    # Costs 65555 gas
+    - data: :abi f(uint) 2
+      accessList: []
+
+    maxFeePerGas: 1000
+    maxPriorityFeePerGas:  0
+    gasLimit:
+    - 100_000   # The balance is high enough only if we aren't also sending wei
```
winsvega, line 54: add documentation to this format when you have time

PR #904, opened 2021-07-26
Tutorial update

qbzzt:
1. Added the invalid transaction information to the generated tests references
2. Added `_` to the tutorial and reference for the test source files (the filler files)
3. Updated the tutorial examples, where needed, so they'll run in London.


docs/test_types/TestStructures/BlockchainTests/block.rst
```diff
@@ -40,19 +40,58 @@ Valid Block Section
...
+                        "rawBytes" : "0x02f87001018502540be4008502540be40086246139ca800094cc...
+                        "exception" : "TR_NoFunds"
+                    },
+                    {
+                        "valid" : "true",
+                        "rawBytes" : "0x02f87001018502540be4008502540be40086246139ca800094cc...
+                    }
+                ]
+
+Blocks that include invalid transactions have an additional field,
```
winsvega, line 83: Put a comment here.
That this is done for those to test transaction rejection. What we do here is

Lets pretend that the block have a given block header 
And then transactions from transaction sequence come in a given order. Those that are invalid must be rejected. 

Before this was being done only when filling the test. Leaving the block empty. I put it for custom test runners in this format  so they can test transaction rejection cases.

PR #905, opened 2021-07-27
retroactive code removal

winsvega:
remove sender code from legacy tests


holiman:
This one is _not_ fixed by this PR: https://github.com/ethereum/tests/issues/908 (doesn't relate to Legacy though). 
However, this does fix the LegacyTests, so :+1:

holiman (approved):
LGTM

winsvega:
there is still a test with sender having code. will fix in another PR

Issue #910, opened 2021-08-03
Add more tests for SSTORE at Constantinople


chfast:
The case where SSTORE in Constantinople cost 200 is not covered by state tests.

![image](https://user-images.githubusercontent.com/573380/128021611-803f3eba-bae5-4f9c-9aef-b30232e76040.png)

qbzzt:
Is there any point in writing tests for Constantinople? It's over a year and a half in the past at this point.

chfast:
Just for completeness. This is my last uncovered line.

PR #913, opened 2021-08-06
Have operations go OOG with different amounts

### LINKED ISSUE #909: Add test for CREATE2 init code hashing OOG
Add a test with CREATE2 where OOG happens directly because of the cost of hashing the init code.

image

qbzzt:
As requested in #909 to cover the specific case in CREATE2 where the hashing
causes OOG

Closing: 909


chfast:
This covers the case from #909.

qbzzt:
> This covers the case from #909.

Thank you. Do you have any related cases that you think we can also cover which it doesn't?

chfast:
> Do you have any related cases that you think we can also cover which it doesn't?

Only #910 left.

winsvega:
ready?

qbzzt:
Yes

PR #931, opened 2021-08-22
regenerate all test files with fields sorted

winsvega:


holiman:
I'm running the tests now. I think you should modify how these files are handled. Whenever a test is regenerated, you should have a step where you compare the existing file (if present) with the new file. The comparison should be json-based, and

- Should ignore (recursively) certain fields, e.g. dates, generated-by-comments. 
- Should ignore ordering in maps

Ideally, it would be in the filler-to-test pipeline, and it would simply skip overwriting the file in the first place.

holiman:
Test passed btw, lgtm!

winsvega:
ordering in maps is ignorred, thats just for prettiness.

I was thinking about this, however I have some logical problem if test is regenerated then dates and versions should be updated. it would be strange to regenerate the test and make no changes. 

however because of the github when there are really no changes made to the tests, then it actually makes sense. 
this might require one more field in the info section - hash sum of the rest of the test. so if no changes then hash is the same I can quicly compare.

Issue #912, opened 2021-08-05
Transaction potentially rejected for the wrong reasons


norswap:
First of all, I know transactions tests are not being maintained, see #906 

Nevertheless, I thought I'd report some issue with them, in case someone is interested.

The issue is that some transactions in those tests have a "malleable signature", essentially, a signature `s` that is higher than the ellipctic curve parameter `N` divided by 2. This should be rejected after [EIP-2](https://eips.ethereum.org/EIPS/eip-2) (to avoid replay attacks) and so is only valid on Frontier.

The affected transactions are:
- `TransactionTests/ttData/DataTestNotEnoughGAS.json`
- `TransactionTests/ttGasPrice/TransactionWithGasPriceOverflow.json`
- `TransactionTests/ttNonce/TransactionWithNonceOverflow.json`

All of these are illegal transactions, and so should fail to parse or to be accepted in general.

The problem is that currently, an implementation may reject them (except for Frontier) because the signature is invalid, and not because it validated the thing that was supposed to be tested (e.g. for `TransactionWithGasPriceOverflow`: that the gas price does not overflow, i.e. is encoded on more than 32 bytes), making these tests much less useful than they should be.

I regenerated the signature on the transactions so that it is EIP-2 compliant.

What I did to work around this was regenerate the signatures on the transactions, here's [the nanoeth commit](https://github.com/norswap/nanoeth/commit/ccb1199b06ad03a54340de0414179a0a6ad51fa2).

I could submit this change as a PR if it's deemed useful. However then I'd need to know how to update the metadata fields. And probably update the filler files too, for consistency?

winsvega:
Hm. Transaction tests currently not supported as there is no legitimate way to generate it. 

You can open a PR, we could pretend that it is still being filled from changed fillers. 

Ideally there would be a mechanism to provide data from filler to the client and get the result if a transaction is legit. But apparently its much less relevant then the core transition tests.

norswap:
Will do the PR then!

I might make such a tool - are you attached to the current JSON format, or could it be changed?
Of course, it wouldn't be useful immediately, since my "client" is itself being implemented and almost assuredly has bugs.

winsvega:
Also the existing ones are verified by aleth constantinople implementation. Are you sure the signature is incorrect?

norswap:
To clarify: these are negative tests, meaning that the transaction should be rejected for the tests to pass.

[...]

PR #933, opened 2021-08-23
Fix transaction potentially rejected for the wrong reasons (#912)

norswap:
cf. https://github.com/ethereum/tests/issues/912

Questions: what should I put in the `filledwith`, `lllcversion` and `sourceHash` fields of the filler?

I haven't validated this against a running node at the moment, but if it's convenient, I might set one up and do it.

Not sure if immensely useful, but [here is the script](https://github.com/norswap/nanoeth/blob/e312edf13eafba9534083ba81d1418b2738d54f9/test/com/norswap/nanoethdev/FixBadSignature.java) I used to generate the new values.

### DIFF
--- a/TransactionTests/ttData/DataTestNotEnoughGAS.json
+++ b/TransactionTests/ttData/DataTestNotEnoughGAS.json
@@ -18,11 +18,11 @@
         },
         "_info" : {
             "comment" : "",
-            "filledwith" : "testeth 1.7.0-alpha.1-33+commit.9ec9d33e.dirty",
-            "lllcversion" : "Version: 0.5.0-develop.2019.8.22+commit.9709dfe0.Linux.g++",
+            "filledwith" : "nanoeth+commit.TODO",
+            "lllcversion" : "none",
             "source" : "src/TransactionTestsFiller/ttData/DataTestNotEnoughGASFiller.json",
-            "sourceHash" : "a43cbd69b4fc997ab2d2b54ffab4267e92e87450e55906b6c66854537ad9d821"
+            "sourceHash" : "TODO"
         },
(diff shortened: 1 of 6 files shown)


winsvega:
To generate this tests we need to use old testeth that can still do it.

However I remember the rule of s signature validation was implemented there. There for the whole issue must be investigated because the tests must be correct

winsvega (changes requested):
Need to check with testeth
Is there really a bug?

norswap:
It's really not that hard to see:
[EIP-2](https://eips.ethereum.org/EIPS/eip-2): 
> All transaction signatures whose s-value is greater than secp256k1n/2 are now considered invalid. The ECDSA recover precompiled contract remains unchanged and will keep accepting high s-values; this is useful e.g. if a contract recovers old Bitcoin signatures.

https://en.bitcoin.it/wiki/Secp256k1: 
```
n = FFFFFFFF FFFFFFFF FFFFFFFF FFFFFFFE BAAEDCE6 AF48A03B BFD25E8C D0364141
```

in `src/TransactionTestsFiller/ttData/DataTestNotEnoughGASFiller.json`:

```
"s" : "0xefffd310ac743f371de3b9f7f9cb56c0b28ad43601b4ab949f53faa07bd2c804"
```

Together:

```
FFFFFFFF FFFFFFFF FFFFFFFF FFFFFFFE BAAEDCE6 AF48A03B BFD25E8C D0364141
efffd310 ac743f37 1de3b9f7 f9cb56c0 b28ad436 01b4ab94 9f53faa0 7bd2c804
```

Obviously, anything whose high-order hex digit is `E` is going to be more than half of a number whose high-order hex digit (in a string of the same length) is `F`.

Hence `s > n/2`, hence the signature is invalid starting from Homestead.

Hence the transaction will get rejected on Homestead+, indepently of the thing that is supposed to be tested (e.g. "not enough gas").

In a certain sense, the tests are not bugged, because they will always succeed on a correct implementation. But they're not testing what they are supposed to be testing.

winsvega:
just checked this file `DataTestNotEnoughGASFiller.json` and  filled `DataTestNotEnoughGAS` 
the test is correct with s/2 requirements, what is the issue?

winsvega:
but you right that it could be rejected to a wrong reason than assumed by test. 
I talked to Martin, we think we can revive thet transaction suite test to cover trasaction structure

winsvega:
fixed with reworked transaction tests

PR #934, opened 2021-08-24
EIP-2681: Limit account nonce to 2^64-1

gumb0:
https://eips.ethereum.org/EIPS/eip-2681
Geth change is in https://github.com/ethereum/go-ethereum/pull/23853

EIP applies retroactively, therefore I put `>=Frontier` for `CREATE` and `>=Constantinople` for `CREATE2`.
(Also not sure what should be done with already generated legacy transaction tests, probably should be updated)


winsvega:
working to revive ttTransaction Tests

winsvega:
checked now with new transaction tests

winsvega:
ah wait its the state tests

src/GeneralStateTestsFiller/stCreate2/CREATE2_HighNonceFiller.yml
```diff
@@ -0,0 +1,56 @@
...
+       value: !!int -1
+      network:
+        - '>=Constantinople'
+      result:
+       a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+          nonce: 1
+       b94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+          nonce: '0xffffffffffffffff'
+          storage:
+            '1': 1
```
marioevz, line 57: The EIP states that the CREATE opcodes should "abort with an exceptional halt" when the nonce is about to overflow.
I would expect the storage remains unchanged, is this ok?

gumb0: Good question, I interpret the EIP text as opcodes themselves end with an exceptional halt, but not the execution frame they're called from (similar to not having enough balance for CREATE's value transfer, call stack overflow, address already existing etc.)

marioevz:
Test cases look good to me, but I think we should add a positive test case for both CREATE/CREATE2 where the nonce is (2**64) - 2, and the contract is still created.
Also, it seems like the current outcome of CREATE_HighNonce is that the nonce wraps around to zero, and STORAGE[0] is equal to the address as if the contract was actually created, although the EIP does not specify what the top of the stack should look like in this scenario.
I'm currently testing on ethereum/go-ethereum@551bd6e.

gumb0:
> Also, it seems like the current outcome of CREATE_HighNonce is that the nonce wraps around to zero, and STORAGE[0] is equal to the address as if the contract was actually created, 

Yes, current geth's behavior is wrapping around, I linked the possible patch in the description https://github.com/ewasm/go-ethereum/commit/36580c62f54e0f157cef11b918ecea2c472cd33f which I used to generate.

> although the EIP does not specify what the top of the stack should look like in this scenario. 

EIP specifies exceptional halt, I think this implies 0 returned on stack, as in all other exceptional cases.

winsvega:
Reworked by new transaction tests

src/GeneralStateTestsFiller/stCreateTest/CREATE_HighNonceFiller.yml
```diff
@@ -0,0 +1,56 @@
...
+       value: !!int -1
+      network:
+        - '>=Frontier'
+      result:
+       a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+          nonce: 1
+       b94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+          nonce: '0xffffffffffffffff'
+          storage:
+            '1': 1
```
marioevz, line 57: Could you add a check here for:
storage:
'0': 0

and

04e9a8460199e670ffb592f93a2f74bdcb44b0bd:
  'shouldnotexist': '1'

src/GeneralStateTestsFiller/stCreateTest/CREATE_HighNonceFiller.yml
```diff
@@ -0,0 +1,56 @@
...
+      balance: 0
+      code: |
+        :yul
+        {
+          // initcode: { return(0, 1) }
+          mstore(0, 0x60016000f3000000000000000000000000000000000000000000000000000000)
+          sstore(0, create(0, 0, 5))
+          sstore(1, 1)
+        }
+      nonce: '0xffffffffffffffff'
```
marioevz, line 28: Could we add a positive test case CREATE_HighNonceMinus1, where nonce is '0xfffffffffffffffe' and contract '0xd061b08a84ebc70fe797f9bd62f4269ef8274a13' is actually created?

src/GeneralStateTestsFiller/stCreate2/CREATE2_HighNonceFiller.yml
```diff
@@ -0,0 +1,56 @@
...
+       gas: !!int -1
+       value: !!int -1
+      network:
+        - '>=Constantinople'
+      result:
+       a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+          nonce: 1
+       b94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+          nonce: '0xffffffffffffffff'
+          storage:
```
marioevz, line 55: We should also check:

storage:
'0': 0

and

77dd5d2a2b742ca01ee2cfff306445e3741ef744:
            'shouldnotexist': '1'

src/GeneralStateTestsFiller/stCreate2/CREATE2_HighNonceFiller.yml
```diff
@@ -0,0 +1,56 @@
...
+      balance: 0
+      code: |
+        :yul
+        {
+          // initcode: { return(0, 1) }
+          mstore(0, 0x60016000f3000000000000000000000000000000000000000000000000000000)
+          sstore(0, create2(0, 0, 5, 0))
+          sstore(1, 1)
+        }
+      nonce: '0xffffffffffffffff'
```
marioevz, line 28: Could we also add here a positive test case CREATE2_HighNonceMinus1, where nonce is '0xfffffffffffffffe' and contract '0x77dd5d2a2b742ca01ee2cfff306445e3741ef744' is actually created?

marioevz (changes requested):
Request for two extra test cases and two minor changes.

src/GeneralStateTestsFiller/stCreateTest/CREATE_HighNonceFiller.yml
gumb0: storage '0': 0 is implicitly checked I think (error if some storage present that is not in test expectations), but added.

src/GeneralStateTestsFiller/stCreateTest/CREATE_HighNonceFiller.yml
gumb0: Added.

src/GeneralStateTestsFiller/stCreate2/CREATE2_HighNonceFiller.yml
gumb0: Added.

marioevz (commented):
Looks good to me, thanks for the changes.
I think this PR still requires ewasm/go-ethereum@eip-2681 to be merged onto ethereum/go-ethereum though.

holiman:
> Geth change is in https://github.com/ewasm/go-ethereum/tree/eip-2681

@gumb0 there's no open PR about it, is there? If not, would you mind opening one?

src/GeneralStateTestsFiller/stCreateTest/CREATE_HighNonceFiller.yml
winsvega: yes. 
but look after cases where you have 
`sstore(0, test())` and a pre storage is empty. 

if smth goes wrong, the transaction reverted, storage remains unchanged and the expect section not showing any error, but the actual exection of test() was aborted or not performed properly

gumb0: If I understand you correctly, that's why `sstore(1, 1)` is done there in the end
(adding `storage '0': 0` wouldn't help this or change anything)

gumb0:
I have also updated transaction tests for a new rule that `nonce == 0xffffffffffffffff` is forbidden (see update to EIP https://github.com/ethereum/EIPs/pull/4437) The geth PR linked in description includes a change to t9n tool for this.

src/TransactionTestsFiller/ttNonce/TransactionWithHighNonce64Minus1Filler.json
```diff
@@ -3,6 +3,7 @@
     {
         "expectException" : 
         {
+            ">=Frontier": "NonceMax"
```
gumb0: This new error also has to be added to retesteth config.

The corresponding geth string is
```
      "NonceMax" : "nonce exceeds 2^64-1",
```

gumb0: @winsvega should it be called `TR_NonceMax` ?

(Some transaction tests use errors with `TR_` prefix, and others without, so not sure)

src/TransactionTestsFiller/ttNonce/TransactionWithHighNonce64Minus2Filler.json
```diff
@@ -0,0 +1,20 @@
+{
+    "TransactionWithHighNonce64Minus1" :
```
marioevz: Should be TransactionWithHighNonce64Minus2

marioevz (approved):
Looks good to me :+1:

src/TransactionTestsFiller/ttNonce/TransactionWithHighNonce64Minus1Filler.json
marioevz: Added https://github.com/ethereum/retesteth/pull/156 for this.

gumb0:
Rebased.

winsvega (approved):
ok.  looks good.

how about a case where a contract with high nonce calls contract via delegate 
and that subcall creates a contract. but since the context of execution moved. the actual contract creator changes.

holiman:
> how about a case where a contract with high nonce calls contract via delegate
> and that subcall creates a contract. but since the context of execution moved. the actual contract creator changes.

`DELEGATECALL` doesn't move the context of execution, it just borrows some code and continues where it was. New scope, same address.

winsvega:
I mean when  A  having  nonce 0, delegatecall  B which has maxnonce,  and B creates contract what happens? 
should work, no?  A will be the creator, not B.

holiman:
Yes, should work, will work

gumb0:
Rebased, removed change to `TransactionWithHighNonce64Minus1` merged in #988.

winsvega:
[what about transaction coming from acc](https://github.com/ethereum/tests/issues/991)

PR #948, opened 2021-09-23
Blake2B (precomp #9) tests, including with lengths up to 1024

qbzzt:


holgerd77:
For reference: see also https://github.com/ethereumjs/ethereumjs-monorepo/pull/1486

jochem-brouwer:
Hiya, just checking here - the bug we found in JS was when the `message` field input to Blake2B was were 5 or more of the first bytes were nonzero. The test cases in the EIP use the string `abc` (so the first 3 bytes are nonzero, the rest are), which are probably taken from the RFC. If we input the string `abcde` then we had a consensus bug due to wrongly converting the input. If these test cases have any input string where the `m` (message) field of the Blake2B input is 5 nonzero bytes or more, then this should cover our case.

jochem-brouwer:
The test case which should be added to the EIP to cover our case was;

```
  {
    input: '0000000c48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b61626364650000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000500000000000000000000000000000001',
    expected: 'f3e89a60ec4b0b1854744984e421d22b82f181bd4601fb9b1726b2662da61c29dff09e75814acb2639fd79e56616e55fc135f8476f0302b3dc8d44e082eb83a8',
    name: 'vector 9'
  }
```

qbzzt:
> Hiya, just checking here - the bug we found in JS was when the `message` field input to Blake2B was were 5 or more of the first bytes were nonzero. The test cases in the EIP use the string `abc` (so the first 3 bytes are nonzero, the rest are), which are probably taken from the RFC. If we input the string `abcde` then we had a consensus bug due to wrongly converting the input. If these test cases have any input string where the `m` (message) field of the Blake2B input is 5 nonzero bytes or more, then this should cover our case.

Thank you, sorry I misunderstood. I'll add more test cases with different message lengths.

PR #949, opened 2021-10-01
Random tests

qbzzt:
1. Generate a random state test
2. Fill the test on `geth`
3. Run the filled test on `besu`
4. If it fails, copy to `/tmp` under a unique name
5. Go to step 1 (endless loop)

Not included yet (I'll do them when I have time): calls to precompiles.


src/Templates/RandomTests/README.md
```diff
@@ -0,0 +1,27 @@
...
+   ```
+   docker run -p 8545:8545 -p 13001:30303 \
+      hyperledger/besu:develop retesteth --rpc-http-port 8545 \
+      --host-allowlist '*' # --logging ALL
+   ```
+
+1. Edit `config` to put the correct IP for besu and copy it to
+   `~/tests/config/besu/config`. You might need to do this as root.
+
+1. Run `./runTests.sh`. It is an endless loop that generates and runs
```
winsvega, line 22: can you make options for this script 
- testpath  
- docker / retesteth (when retesteth then original datadir by default is used)
- StateTest / BlockchainTest
- rounds <num> / unlimited (limit loop for testing)
- errored test output dir

src/Templates/RandomTests/oneTest.sh
```diff
@@ -0,0 +1,31 @@
...
+dir=`pwd`/../../..
+
+# Create a random test
+./randomTest.js > ../../GeneralStateTestsFiller/stExample/randomFiller.yml
+
+cd ~
+
+# Run the random test on Geth
+./dretesteth.sh -t $suite -- --testpath $dir --singletest $test \
+                --filltests --datadir /tests/config
```
winsvega: what if geth fails on test filling ?
should be reported too

winsvega (changes requested):
add options to the script so I can run it on vps

src/Templates/RandomTests/oneTest.sh
qbzzt: Added

src/Templates/RandomTests/README.md
qbzzt: - [x]  testpath
- [x] docker / retesteth (when retesteth then original datadir by default is used)
- [x] StateTest / BlockchainTest
- [x] rounds / unlimited (limit loop for testing)
- [x] errored test output dir

qbzzt: I created all of those command line options. I still need to:

1. Update the README.md
2. Finish what I'm writing to have an explanation of the code.

winsvega:
should keep in mind that besu require javahome to be set 
    `export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64`
for me it seems `JAVA_HOME` could not be set through `/etc/environment`
so I set it manually every time.

also example command: 
`./runTests.sh -n 1 -e /data/rndtests -r`

unfortunately on vps besu does not work (((

qbzzt:
> should keep in mind that besu require javahome to be set `export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64` for me it seems `JAVA_HOME` could not be set through `/etc/environment` so I set it manually every time.
> 
> also example command: `./runTests.sh -n 1 -e /data/rndtests -r`
> 
> unfortunately on vps besu does not work (((

`oneTests.sh` doesn't run Besu, it connects to be an existing instance (specified in the retesteth config). So Besu can just as easily run in a VM outside the vps.

winsvega:
can you name the test a little bit simplier. and the name inside of a test file should match the name of test file
because thats the rule we made for state tests

qbzzt:
> can you name the test a little bit simplier. and the name inside of a test file should match the name of test file
> because thats the rule we made for state tests

Consider it done. The name is now <type of test>-<uuid>Filler.yml.

winsvega (approved):

PR #955, opened 2021-10-11
TransactionTests Update: ttGasLimit

marioevz:
Contains updated TransactionTests for category ttGasLimit.
Outstanding issues:
- Tests TransactionWithHighGas and TransactionWithHihghGasLimit63m1 expect the transaction to be rejected with GasLimit 0x7fffffffffffffff, but they pass, and this seems expected since the uint64 should allow for this value. Please confirm and I can change the expected result to pass.


src/TransactionTestsFiller/ttGasLimit/TransactionWithHihghGasLimit63m1Filler.json
```diff
@@ -1,19 +1,14 @@
 {
-    "TransactionWithHihghGasLimit63m1" : {
-	     "expect" : [
-            {
-                "network" : ["EIP158", "Byzantium", ">=Constantinople"],
-                "result" : "valid"
```
winsvega, line 6: was gasLimit cap a retroactive change ?

src/TransactionTestsFiller/ttGasLimit/TransactionWithHighGasFiller.json
```diff
@@ -1,26 +1,21 @@
...
+    {
+	    "expectException" : 
+        {
+            ">=Homestead" : "InvalidGasLimit4"
+        },
         "transaction" :
         {
             "data" : "",
-	    "gasLimit" : "0x7fffffffffffffff",
+	        "gasLimit" : "0x7fffffffffffffff",
```
winsvega: this test is identical to `TransactionWithHihghGasLimit63m1Filler.json `
better change to   x, x-1, x+1(if x not fails)  where x is gasLimit cap  (pow(2,63)-1)

src/TransactionTestsFiller/ttGasLimit/TransactionWithGasLimitxPriceOverflowFiller.json
```diff
@@ -1,15 +1,13 @@
...
-        ],
+    "TransactionWithGasLimitxPriceOverflow" :
+    {
+	    "expectException" : 
+        {
+        },
         "transaction" :
         {
             "data" : "",
-            "gasLimit" : "115792089237316195423570985008687907853269984665640564039457584007913129639935",
```
winsvega, line 12: the goal of this test is to make   gasLimit * gasPrice  >  2**256-1, need to change. is there a cap on gasPrice?

src/TransactionTestsFiller/ttGasLimit/TransactionWithGasLimitOverflowZeros64Filler.json
```diff
@@ -0,0 +1,21 @@
+{
+    "TransactionWithGasLimitOverflowZeros64" :
+	{
+		"expectException" : 
+        {
+            ">=Frontier" : "InvalidGasLimit4"
```
winsvega, line 6: the exception here is the same as in overflow, but shouldn't it actually throw leading zeroes exception?

src/TransactionTestsFiller/ttGasLimit/TransactionWithGasLimitOverflow63Filler.json
```diff
@@ -1,22 +1,20 @@
...
-		"r" : "0x48b55bfa915ac795c431978d8a6a992b628d557da5ff759b307d495a36649353",
-		"s" : "0x1fffd310ac743f371de3b9f7f9cb56c0b28ad43601b4ab949f53faa07bd2c804"
-	}
+		"expectException" : 
+        {
+        },
+		"transaction" :
+		{
+			"data" : "",
+			"gasLimit" : "0xffffffffffffffff",
```
winsvega: why not 8000000 ?  
why it passes, it overflows ?

winsvega (commented):
I have some questions
and need a change

winsvega (changes requested):
some exceptions are strange need to double chek

marioevz:
The limit is uint64, I will be renaming all tests which do not overflow (< 2^64) to HighGas instead of Overflow, to make it easier to identify them.
I am not sure why the tests originally expected 2^63 to overflow, but this should not be the case.
I am also removing repeated tests, TransactionWithHihghGasLimit63m1 for example only fails because V=37, but I am not sure what was the original goal of this specific test case.
Will upload a new commit soon.

src/TransactionTestsFiller/ttGasLimit/TransactionWithGasLimitOverflowZeros64Filler.json
marioevz: As soon as the value passes the 64bit length in zeros, the error becomes the same as overflow instead of the leading zeros error, I am unsure if this is ok or not.
Even with 0x000000000000000000, the error is still the overflow error.

src/TransactionTestsFiller/ttGasLimit/TransactionWithGasLimitxPriceOverflowFiller.json
marioevz: I am changing this test to be gasLimit=0xffffffffffffffff, gasPrice=0x1000000000000000100000000000000010000000000000002, where gasLimit is the highest uint64 value, and gasPrice is the lowest value such as the product is equal/higher than a 256bit value.
The test still does not produce an exception.

src/TransactionTestsFiller/ttGasLimit/TransactionWithHihghGasLimit63m1Filler.json
marioevz: I am not sure why the original test contained selective forks, the uint64 limit applies to all forks it seems.

src/TransactionTestsFiller/ttGasLimit/TransactionWithGasLimitxPriceOverflowFiller.json
winsvega: Well the transaction technically valid.
Just that irl no account with enough balance exists.

Is int64max a limit of gasLimit field?

src/TransactionTestsFiller/ttGasLimit/TransactionWithGasLimitOverflowZeros64Filler.json
winsvega: Thats question to martin on how he does the check. @holiman

src/TransactionTestsFiller/ttGasLimit/TransactionWithGasLimitxPriceOverflowFiller.json
marioevz: Yes, it's unsigned so uint64 and 0xffffffffffffffff is the max.
I agree that the transaction is valid, I think is a good test to have just to see that no client does something funny when receiving the transaction.
~~Should I add a state test similar to the one I added for gasPrice in the other PR ?~~
I just checked the test case and is exactly this scenario, so no need to add another state test...

src/TransactionTestsFiller/ttGasLimit/TransactionWithGasLimitOverflowZeros64Filler.json
holiman: I'm sorry, I'm feeling a bit daft -- could you clarify more specifically what the question is?

marioevz: For the input "0x000000000000000001", we were expecting the leading-zero-bytes exception:
> rlp: non-canonical integer (leading zero bytes) for uint64, decoding into (types.Transaction)(types.LegacyTx).Gas

And we are getting the input-too-long exception:
> rlp: input string too long for uint64, decoding into (types.Transaction)(types.LegacyTx).Gas

But after thinking about it, I don't think this is really an issue, both seem to be valid exceptions for the test case.

src/TransactionTestsFiller/ttGasLimit/TransactionWithHighGasLimit65Minus1Filler.json
```diff
@@ -0,0 +1,20 @@
+{
+    "TransactionWithHighGasLimit65Minus1" :
+	{
+		"expectException" : 
+        {
+        },
+		"transaction" :
+		{
+			"data" : "",
+			"gasLimit" : "0xffffffffffffffff",
```
winsvega: this is 64-1

src/TransactionTestsFiller/ttGasLimit/TransactionWithHighGasLimit64Plus1Filler.json
```diff
@@ -0,0 +1,20 @@
+{
+    "TransactionWithHighGasLimit64Plus1" :
+	{
+		"expectException" : 
+        {
+        },
+		"transaction" :
+		{
+			"data" : "",
+			"gasLimit" : "0x8000000000000001",
```
winsvega: this is 63+1

src/TransactionTestsFiller/ttGasLimit/TransactionWithHighGasLimit63Filler.json
```diff
@@ -0,0 +1,20 @@
+{
+    "TransactionWithHighGasLimit63" :
+    {
+	    "expectException" : 
+        {
+        },
+        "transaction" :
+        {
+            "data" : "",
+	        "gasLimit" : "0x7fffffffffffffff",
```
winsvega: this is 63-1

src/TransactionTestsFiller/ttGasLimit/TransactionWithGasLimitxPriceOverflowFiller.json
winsvega: what is the exception? GasLimitPriceProductOverflow

src/TransactionTestsFiller/ttGasLimit/TransactionWithGasLimitOverflow65Filler.json
```diff
@@ -0,0 +1,21 @@
...
+    "TransactionWithGasLimitOverflow65" :
+	{
+		"expectException" : 
+        {
+            ">=Frontier" : "InvalidGasLimit4"
+        },
+		"transaction" :
+		{
+			"data" : "",
+			"gasLimit" : "0x10000000000000000",
```
winsvega: this is 64

winsvega (changes requested):
naming is wrong. did you change the  names or is it a github issue?

marioevz:
@winsvega I changed the naming to reference the number of bits instead of powers of two. It's a mistake, I will change it back.

marioevz:
Fixed naming and TransactionWithGasLimitxPriceOverflow test now correctly fails with GasLimitPriceProductOverflow (gas * gasPrice exceeds 256 bits).

PR #959, opened 2021-10-12
TransactionTests Update: ttRSValue

marioevz:
Contains updated TransactionTests for category ttRSValue.

### DIFF
--- a/src/TransactionTestsFiller/ttRSValue/RightVRSTestVPrefixedBy0_3Filler.json
+++ b/src/TransactionTestsFiller/ttRSValue/RightVRSTestVPrefixedBy0_3Filler.json
@@ -1,11 +1,11 @@
 {
-    "RightVRSTestVPrefixedBy0_3" : {
-	"expect" : [
-            {
-                "network" : [">=Frontier"],
-                "result" : "invalid"
-            }
-        ],
+    "RightVRSTestVPrefixedBy0_3" :
+    {
+	    "expectException" : 
+        {
+            "<=EIP150": "InvalidVRS",
+            ">=EIP158": "InvalidChainID"
+        },
         "transaction" :
         {
             "data" : "0x",
--- a/src/TransactionTestsFiller/ttRSValue/RightVRSTestVPrefixedBy0_3Filler.json
+++ b/src/TransactionTestsFiller/ttRSValue/RightVRSTestVPrefixedBy0_3Filler.json
@@ -3,8 +3,7 @@
     {
 	    "expectException" : 
         {
-            "<=EIP150": "InvalidVRS",
-            ">=EIP158": "InvalidChainID"
+            ">=Frontier": "LeadingZerosV"
         },
         "transaction" :
         {
@@ -14,7 +13,7 @@
             "nonce" : "3",
             "to" : "b94f5374fce5edbc8e2a8697c15331677e6ebf0b",
             "value" : "10",
-            "v" : "0x0030",
+            "v" : "0x:bigint 0x0030",
             "r" : "0x98ff921201554726367d2be8c804a7ff89ccf285ebc57dff8ae4c44b9c19ac4a",
             "s" : "0x1887321be575c8095f789dd4c743dfe42c1820f9231f98a962b210e3ac2452a3"
         }
--- a/src/TransactionTestsFiller/ttRSValue/TransactionWithRvalueHighFiller.json
+++ b/src/TransactionTestsFiller/ttRSValue/TransactionWithRvalueHighFiller.json
@@ -1,11 +1,10 @@
 {
-    "TransactionWithRvalueHigh" : {
-	"expect" : [
-            {
-                "network" : [">=Frontier"],
-                "result" : "invalid"
-            }
-        ],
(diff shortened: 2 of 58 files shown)


src/TransactionTestsFiller/ttRSValue/TransactionWithSvaluePrefixed00Filler.json
```diff
@@ -1,11 +1,9 @@
...
-	"expect" : [
-            {
-                "network" : [">=Frontier"],
-                "result" : "valid"
-            }
-        ],
+    "TransactionWithSvaluePrefixed00" :
+    {
+	    "expectException" : 
+        {
```
winsvega, line 5: here is not exactly prefixed. there was no way to make a prefixed by then. add a copy of this test using big int and add 00 prefixes to s value, but keep it in range.

src/TransactionTestsFiller/ttRSValue/TransactionWithSvalue1Filler.json
```diff
@@ -1,11 +1,9 @@
...
-    "TransactionWithSvalue1" : {
-	"expect" : [
-            {
-                "network" : [">=Frontier"],
-                "result" : "valid"
-            }
-        ],
+    "TransactionWithSvalue1" :
+    {
+	    "expectException" :
```
winsvega, line 4: is it a github bug or ident here is different to the one with transaction?

src/TransactionTestsFiller/ttRSValue/TransactionWithRvaluePrefixed00Filler.json
```diff
@@ -1,11 +1,9 @@
...
-	     "expect" : [
-            {
-                "network" : [">=Frontier"],
-                "result" : "valid"
-            }
-        ],
+    "TransactionWithRvaluePrefixed00" :
+    {
+	    "expectException" : 
+        {
```
winsvega, line 5: same here. need a copy of this test with bigint prefixed 00 and in range. 
(out of range overflow crazy bigint scenarious we can add up later)

src/TransactionTestsFiller/ttRSValue/RightVRSTestVPrefixedBy0_3Filler.json
```diff
@@ -1,11 +1,11 @@
...
-            {
-                "network" : [">=Frontier"],
-                "result" : "invalid"
-            }
-        ],
+    "RightVRSTestVPrefixedBy0_3" :
+    {
+	    "expectException" : 
+        {
+            "<=EIP150": "InvalidVRS",
```
winsvega: oh, I forgot I actully support <= in network translation

winsvega: here it should reject I think not with invalid chain id but with prefixed 00, as the v value limitations. need to check with the v value cap.

src/TransactionTestsFiller/ttRSValue/TransactionWithSvaluePrefixed00Filler.json
marioevz: I created the new test case and it actually fails with:
rlp: non-canonical integer (leading zero bytes) for *big.Int, decoding into (types.Transaction)(types.LegacyTx).S
Which I think is ok.

src/TransactionTestsFiller/ttRSValue/TransactionWithRvaluePrefixed00Filler.json
marioevz: Added test case and it correctly fails like the S case.

src/TransactionTestsFiller/ttRSValue/TransactionWithSvalue1Filler.json
marioevz: The indent was incorrect in the original file, I tried to make indentation consistent in all files.

src/TransactionTestsFiller/ttRSValue/RightVRSTestVPrefixedBy0_3Filler.json
marioevz: Changed the expected to LeadingZerosV, and changed the V value to bigint.
Let me know if this is ok.

src/TransactionTestsFiller/ttRSValue/TransactionWithSvalue1Filler.json
winsvega: I think you added TAB

src/TransactionTestsFiller/ttRSValue/TransactionWithRvalueHighFiller.json
```diff
@@ -14,9 +13,9 @@
             "nonce" : "0",
             "to" : "095e7baea6a6c7c4c2dfeb977efac326af552d87",
             "value" : "11",
-            "v" : "27",
+            "v" : "28",
```
winsvega, line 16: why change?

winsvega (approved):
looks good

PR #963, opened 2021-10-13
Add EIP-3607 tests for Sender not EOA

### LINKED ISSUE #952: Include tests for EIP-3607
Include set of tests that verify that clients have implemented https://eips.ethereum.org/EIPS/eip-3607.

marioevz:
This PR closes #952 .
The following test cases are included:
- An account with non-empty code sends the following transactions: send value, calls a contract, calls itself, deploys contract, init code executes: call, delegatecall
- An account with empty code attempts to init into a non-empty code address: contract deployment to non-empty address, init code executes: call, delegatecall, create, create2

All tests are negative test cases and correctly fail.

This PR excludes tests where create and create2 deploy to an existing address since this is already tested by: GeneralStateTests/stCreate2, GeneralStateTests/stCreateTest


holiman:
Just reading the description, I think this PR might be a bit over-ambitious? 
The thing is, _any_ transaction made from a non-EOA account should be rejected out of hand. So I think it would be sufficient to test whether a basic 'send ether to a recipient', 'make a create-tx' are rejected or not. That said, I'm fine with it being as it is, just noting my thoughts. 

Also, to clarify; this change should be retroactive, so could be extended for pre-berlin aswell, but that's not important either.

marioevz:
Thanks for the feedback, what I had in mind was that perhaps other clients implemented this differently and adding some extra checks just to be safe would be ok, but if everyone is inclined to reduce the number of test cases, that is also ok by me.
Also, I will change the fork to be retroactive.
Thanks again.

qbzzt:
> Thanks for the feedback, what I had in mind was that perhaps other clients implemented this differently and adding some extra checks just to be safe would be ok, but if everyone is inclined to reduce the number of test cases, that is also ok by me. 

My opinion is that having a few extra tests won't hurt anything, and might save us from an unspecified error in the future.

MariusVanDerWijden:
Reminder to self: selfdestruct a contract and send a tx from it after

marioevz:
> Reminder to self: selfdestruct a contract and send a tx from it after

Added this test case in the latest commit.
I first tried the contract to send the self-destruct funds to itself, but since the account is deleted, the funds also are destroyed.

The test case I ended up writing:
- Self-destruct sends the funds to another account
- This other account returns the funds to the address that just self-destructed
- The funds are then finally retrieved with the secretKey

Let me know if this is ok.
Thanks.

holiman:
> Let me know if this is ok.

:+1: Looks good to me!

src/GeneralStateTestsFiller/stEIP3607/nonEmptyAccount_initsFiller.yml
```diff
@@ -0,0 +1,67 @@
...
+    6295ee1b4f6dd65047762f924ecd367c17eabf8f:
+      balance: '0'
+      nonce: '0'
+      code: ''
+      storage: {}
+
+  # Transaction originating from account with code
+  transaction:
+    data:
+    - ':raw 0x00'
```
winsvega: you can define multiple data raws here. 
all the transactions would get rejected. I don't think there is a much need to cover every possible init data code. 
but because each state test is translated into bc test for each fork here. please limit the amount of tests. 
if the pre state is the same you can define it in one file with different data raws.

src/GeneralStateTestsFiller/stEIP3607/nonEmptyAccount_contractDeployFiller.yml
```diff
@@ -0,0 +1,69 @@
...
+  transaction:
+    data:
+    - ':yul {
+         return (0, 32)
+       }'
+    gasLimit:
+    - '400000'
+    gasPrice: '10'
+    nonce: '0'
+    to: ''
```
winsvega: the 2 cases here is really when transaction has destination. 
and the second when it runs creation. (empty to) this are 2 different files

src/GeneralStateTestsFiller/stEIP3607/initNonEmptyAccount_callFiller.yml
```diff
@@ -0,0 +1,80 @@
+# Init into a non-empty account, call from init code
+
+initNonEmptyAccount_call:
```
winsvega: this is an interesting case. collision and wrong sender at the same time.

winsvega (changes requested):
I think can merge many tests into one and use data as array.

winsvega:
yes a good test case but in blockchain test within one block
first transaction destroys the account (selfdestruct) 
second one is coming from the destroyed account. (should be valid now, but fail due to the account empty balance)

marioevz:
> yes a good test case but in blockchain test within one block

Yes this was created as a blockchain test in src/BlockchainTestsFiller/ValidBlocks/bcStateTests/contractSuicideThenSendFiller.yml

Also reduced the number of files as suggested in latest commit, and changed the fork to ">=London" to generate less tests.

src/GeneralStateTestsFiller/stEIP3607/nonEmptyAccount_initFiller.yml
```diff
@@ -0,0 +1,87 @@
...
+
+
+  expect:
+    - indexes:
+        data: !!int -1
+        gas:  !!int 0
+        value: !!int -1
+
+      network:
+        - '>=London'
```
winsvega: ok here could be >=Frontier

src/GeneralStateTestsFiller/stEIP3607/nonEmptyAccount_callsFiller.yml
```diff
@@ -0,0 +1,67 @@
...
+      balance: '1000000000000000000'
+      nonce: '0'
+      code: ':raw 0xdeadbeef'
+      storage: {}
+
+    # Account receiving the funds from the account with code
+    095e7baea6a6c7c4c2dfeb977efac326af552d87:
+      balance: '0'
+      nonce: '0'
+      code: ':raw 0xdeadbeef'
```
winsvega: put something meaningfull like  0x6000600155

src/GeneralStateTestsFiller/stEIP3607/nonEmptyAccount_callsFiller.yml
```diff
@@ -0,0 +1,67 @@
...
+
+
+  expect:
+    - indexes:
+        data: !!int -1
+        gas:  !!int 0
+        value: !!int -1
+
+      network:
+        - '>=London'
```
winsvega: here >=Frontiner

src/GeneralStateTestsFiller/stEIP3607/initNonEmptyAccountFiller.yml
```diff
@@ -0,0 +1,102 @@
+# Account attempts to send tx to create a contract on a non-empty address
+
+initNonEmptyAccount:
```
winsvega: put here that collision involved

src/GeneralStateTestsFiller/stEIP3607/initNonEmptyAccountFiller.yml
```diff
@@ -0,0 +1,102 @@
...
+    - indexes:
+        data: !!int -1
+        gas:  !!int 0
+        value: !!int -1
+
+      network:
+        - '>=London'
+      
+      result:
+        6295ee1b4f6dd65047762f924ecd367c17eabf8f:
```
winsvega: check a94f5374fce5edbc8e2a8697c15331677e6ebf0b  nonce

src/BlockchainTestsFiller/ValidBlocks/bcStateTests/contractSuicideThenSendFiller.yml
```diff
@@ -0,0 +1,92 @@
...
+      balance: '10000000000'
+      nonce: '0'
+      code: ''
+      storage: {}
+
+    # Contract that will self-destruct, but we have its secretKey
+    0xfebc70C40FadB14d796F1d499b3f247e12FBbd94:
+      balance: '10000000000'
+      nonce: '0'
+      code: :raw 0x737ae27a42774a1f034ac7edb93d23cb5a662017d5FF
```
winsvega: source code???

src/BlockchainTestsFiller/ValidBlocks/bcStateTests/contractSuicideThenSendFiller.yml
```diff
@@ -0,0 +1,92 @@
...
+      gasLimit: 500000
+      gasPrice: 100
+      nonce: auto
+      secretKey: 45a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065ff2d8
+      to: 0xfebc70C40FadB14d796F1d499b3f247e12FBbd94
+      value: 0
+
+    # We send back the assets received from the self-destruct opcode
+    - data: ''
+      gasLimit: 21000
```
winsvega: put more gaslimit as it tend to change to higher number in the future forks

src/BlockchainTestsFiller/ValidBlocks/bcStateTests/contractSuicideThenSendFiller.yml
```diff
@@ -0,0 +1,92 @@
+# Contract self-destructs and then a private key with colliding address attempts to retrieve value from the account.
+
+contractSuicideThenSend:
```
winsvega: transaction from selfdestruct

src/BlockchainTestsFiller/ValidBlocks/bcStateTests/contractSuicideThenSendFiller.yml
```diff
@@ -0,0 +1,92 @@
...
+      value: 9995800000
+
+  expect:
+  - network:
+    - ">=London"
+    result:
+      0xd0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0:
+        balance: '9995800000'
+
+      0xfebc70C40FadB14d796F1d499b3f247e12FBbd94:
```
winsvega: check nonce

winsvega (changes requested):

PR #966, opened 2021-10-18
Add Test Cases with Explicit Chain ID to ttVValue

marioevz:
This PR adds 8 test cases to the TransactionTests/ttVValue test suite where the "chainId" field is explicitly set:
- Two positive test cases where CHAIN_ID == 1, and V == {0,1} + CHAIN_ID * 2 + 35
- Two negative test cases where CHAIN_ID == 0, and V == {0,1} + CHAIN_ID * 2 + 35
- Two negative test cases where CHAIN_ID == 1, and V != {0,1} + CHAIN_ID * 2 + 35
- Two negative test cases where CHAIN_ID == 1, and V == {0,1} + CHAIN_ID * 2 + 35, but V has leading zeros

### DIFF
--- a/src/TransactionTestsFiller/ttVValue/ValidChainID1ValidV1Filler.json
+++ b/src/TransactionTestsFiller/ttVValue/ValidChainID1ValidV1Filler.json
@@ -1,10 +1,9 @@
 {
-    "V_invalidChainID0P1" :
+    "ValidChainID1ValidV1" :
     {
 	    "expectException" : 
         {
-            "<=EIP150" : "InvalidVRS",
-            ">=EIP158" : "InvalidChainID"
+            "<=EIP150" : "InvalidVRS"
         },
         "transaction" :
         {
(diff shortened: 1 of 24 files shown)


src/TransactionTestsFiller/ttVValue/ValidChainID1ValidV1Filler.json
```diff
@@ -0,0 +1,22 @@
...
+        },
+        "transaction" :
+        {
+            "data" : "",
+            "gasLimit" : "21000",
+            "gasPrice" : "1",
+            "nonce" : "3",
+            "to" : "b94f5374fce5edbc8e2a8697c15331677e6ebf0b",
+            "value" : "10",
+            "chainId" : "1",
```
winsvega, line 16: this field does not affect anything. (for legacy type transaction)
retesteth allows it because besu return transaction with this field for all transaction types.

PR #967, opened 2021-10-19
Fix failing tests for ttSignature

marioevz:
This PR fixes the following test cases for ttSignature:
- EmptyTransaction: Changed the filler to have empty values for all fields. Test case failing with: "invalid transaction v, r, s values"
- TransactionWithTooFewRLPElements / TransactionWithTooManyRLPElements: Changed to "Copier" format and added the txbytes RLP directly. Test cases correctly failing with "rlp: too few elements" and "rlp: input list has too many elements" respectively, but retesteth is currently not recognizing the exceptions.


winsvega (approved):
this are good tests. too bad t9n does not test the geth logic. 
I suggest not to focus on this scenarios just yet. there is bcForgedTest  folder with this kind of tests that are being fed to client as a block rlp. which can be executed on hive. 
lets work on that next.

Issue #941, opened 2021-09-07
Test wanted: ecRecover failure for py-evm


holiman:
Reported by @guidovranken.  cc @carver @pipermerriam 

Here's a repro of a testcase that apparently fails on py-evm, but succeeds on geth and nethermind. The test calls `ecRecover`, and the signature recovery fails on geth and nethermind, but succeeds on py-evm
```json
{
 "ecRecoverTest": {
  "env": {
   "currentCoinbase": "b94f5374fce5edbc8e2a8697c15331677e6ebf0b",
   "currentDifficulty": "0x20000",
   "currentGasLimit": "0x26e1f476fe1e22",
   "currentNumber": "0x1",
   "currentTimestamp": "0x3e8",
   "previousHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
   "currentBaseFee": "0x10"
  },
  "pre": {
   "0x000000000000000000000000000000000000ff0a": {
    "code": "0x7f6b8d2c81b11b2d699528dde488dbdf2f94293d0d33c32e347f255fa4a6c1f0a9600052601b6020527f79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f817986040527f6b8d2c81b11b2d699528dde488dbdf2f94293d0d33c32e347f255fa4a6c1f0a9606052602060006080600060006001620f4240f150600051600055",
    "storage": {},
    "balance": "0xffffffff",
    "nonce": "0x0"
   },
   "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b": {
    "code": "0x",
    "storage": {},
    "balance": "0xde0b6b3a7640000",
    "nonce": "0x0"
   }
  },
  "transaction": {
   "gasPrice": "0x10",
   "nonce": "0x0",
   "to": "0x000000000000000000000000000000000000ff0a",
   "data": [
    ""
   ],
   "gasLimit": [
    "0x989680"
   ],
   "value": [
    "0x0"
   ],
   "secretKey": "0x45a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065ff2d8"
  },
  "out": "0x",
  "post": {
   "London": [
    {
     "hash": "b520c02c1dedacb96574e47fe27a3053e9d0f2b9d52dadd3702a00466532ac31",
     "logs": "1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
     "indexes": {
      "data": 0,
      "gas": 0,
[...]

PR #973, opened 2021-10-28
Add Tests for ecRecover on Point at Infinity

marioevz:
This PR aims to implement test described in #941.
At the moment, only the state test described in the PR is implemented.
I think it would be interesting to also implement a transaction test where the address recovery of the tx hash + v, r, s, results in the same condition as this original test, but I am still trying to figure out how to hit the condition this way.
Let me know if this is a good idea for a test case.


marioevz:
Added TransactionTests test case for this scenario where ecRecover of a transaction hits this same issue.
The trick was to sign the transaction with 0x0000000000000000000000000000000000000000000000000000000000000000 private key.

src/TransactionTestsFiller/ttSignature/PointAtInfinityFiller.json
```diff
@@ -0,0 +1,21 @@
+{
+    "PointAtInfinity" :
+    {
```
winsvega, line 3: can you please add "_info" : { "comment" : "  DESCRIPTION " }
explaining a bit the math. so to know which point at infinity.

src/GeneralStateTestsFiller/stTransactionTest/PointAtInfinityECRecoverFiller.yml
```diff
@@ -0,0 +1,59 @@
...
+      storage: {}
+
+  transaction:
+    data:
+      - ''
+    gasLimit:
+      - 10000000
+    gasPrice: 10
+    nonce: 0
+    secretKey: 45a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065ff2d8
```
winsvega, line 40: here is also must be supported the v,r,s format. 
try adding up invalid signature transaction test, see if it works with expect exception.

winsvega (approved):

PR #976, opened 2021-10-31
Remove Updated Difficulty Tests from BasicTests Folder

marioevz:
This PR removes the BasicTests folder which has been relocated to the ethereum/legacytests repo.

**Update:** Only remove difficulty tests that are already updated in #977.


winsvega:
there are some files that might still be used by someone. 
those are really legacy tests, cause it was here before the frontier even

BasicTests/blockgenesistest.json
```diff
@@ -1,20 +0,0 @@
-[
```
winsvega: I have no idea how many teams are still using this files. perhaps it is better to leave it here as it is small

BasicTests/crypto.json
```diff
@@ -1,16 +0,0 @@
-{
```
winsvega: I have no idea how many teams are still using this files. perhaps it is better to leave it here as it is small

BasicTests/genesishashestest.json
```diff
@@ -1,5 +0,0 @@
-{
```
winsvega: I have no idea how many teams are still using this files. perhaps it is better to leave it here as it is small

BasicTests/hexencodetest.json
```diff
@@ -1,62 +0,0 @@
-{
```
winsvega: I have no idea how many teams are still using this files. perhaps it is better to leave it here as it is small

BasicTests/keyaddrtest.json
```diff
@@ -1,22 +0,0 @@
-[
```
winsvega: I have no idea how many teams are still using this files. perhaps it is better to leave it here as it is small

BasicTests/txtest.json
```diff
@@ -1,24 +0,0 @@
-[
-    {
```
winsvega: I have no idea how many teams are still using this files. perhaps it is better to leave it here as it is small

winsvega (changes requested):
lets move to legacy only the difficulty tests 
legacytests root / BasicTests / *olddifficulty tests

holiman:
As for go-ethereum, we use the BasicTests in this file: https://github.com/ethereum/go-ethereum/blob/master/tests/difficulty_test.go#L84 -- it is the `difficultyTestDir`. 

We skip these: 
```
	dt.skipLoad("hexencodetest.*")
	dt.skipLoad("crypto.*")
	dt.skipLoad("blockgenesistest\\.json")
	dt.skipLoad("genesishashestest\\.json")
	dt.skipLoad("keyaddrtest\\.json")
	dt.skipLoad("txtest\\.json")
```
And these: 
```
	// files are 2 years old, contains strange values
	dt.skipLoad("difficultyCustomHomestead\\.json")
	dt.skipLoad("difficultyMorden\\.json")
	dt.skipLoad("difficultyOlimpic\\.json")
```

holiman:
For all changes such as these, please make sure to write them up on the release notes, eventually, so we can make sure to update and use the new paths.

winsvega:
yes. the existing difficulty tests we plan to move to LegacyTests subfolder (repo)
and have a copy of them in a new difficulty test format https://github.com/ethereum/tests/pull/970
with new tests to cover arrowGlacier as well. 

JavaScript team reported that they still use some of the tests from this folder

marioevz:
I think that partially removing some of the files might end up being more confusing to the users instead of simply instructing:
"./BasicTests is now at ./LegacyTests/BasicTests, be sure to add '--recurse-submodules' when cloning this repo to obtain these files if you require them", or something similar.
Let me know what you think @holiman @winsvega

winsvega:
for thise little files it does not make much sense to fetch the whole legacy tests repo. 
removing old difficulty tests to legacy repo is what we want. it will make everyone notice that there are new difficulty tests

marioevz:
Got it, sounds good. Will remove only what's being updated in #977, everything else remains in ./BasicTests .
For ethereum/legacytests#1, I added everything under ./BasicTests -> ./LegacyTests/BasicTests, but will update to only add files removed in this PR.

PR #984, opened 2021-11-11
Refill All TransactionTests to Include intrinsicGas

marioevz:
t8ntool was updated to return the "intrinsicGas" field, containing the expected intrinsic gas value for a transaction.
All tests in TransactionTests Copier source files, and the filled tests have been updated to include the field.


winsvega:
you used this branch https://github.com/ethereum/go-ethereum/pull/23889
right?

marioevz:
I re-uploaded the filled tests using the correct version for filling-rpc-server, I had indeed used that branch but my working tree contained changes and the version was incorrect.

PR #986, opened 2021-11-15
Improve code size limit tests

gumb0:
1. Add two new cases:
- Valid: creation transaction returning code with max allowed size (0x6000)
- Invalid: creation transaction returning code with size 0x6001

2. Huge initcode is not needed, because CODECOPY implicitly pads code with 0s, so I suggest to remove the code after RETURN for simplification.

3. (optional) I replaced the bytecode with yul. 

4. New tests similar to 1 for CREATE/CREATE2 opcodes


src/GeneralStateTestsFiller/stCodeSizeLimit/codesizeOOGInvalidSizeFiller.json
```diff
@@ -33,8 +33,11 @@
             }
         },
         "transaction" : {
+            "// code size is 0x600d" :" mstore(0x40, 0x60) codecopy(0x00, 0x0012, 0x600d) return(0x00, 0x600d)",
```
gumb0: Note: it's rather pseudocode than exact disassembly, it actually uses DUP1 for code size constant.
(maybe would make sense to just replace raw bytecode with yul here?)

gumb0: Also I didn't get why MSTORE is needed, probably it's not.

marioevz: Just checked and indeed the MSTORE is not required.

src/GeneralStateTestsFiller/stCodeSizeLimit/createCodeSizeLimitFiller.yml
```diff
@@ -0,0 +1,76 @@
+# CREATE with max allowed deployed code size
+createCodeSizeLimit:
```
winsvega, line 2: very nice!

winsvega:
This branch is out-of-date with the base branch
rebase?

winsvega (approved):

gumb0:
Rebased.

Issue #985, opened 2021-11-15
codesize tests for dynamic contract creation


winsvega:
Test create / create2 contract creation (abort on code size limit)

Plus can check this case:

Create a contract with large code but still valid. Destroy it and then revert the transaction

PR #987, opened 2021-11-15
Implement Create Contract Max codesize Tests

marioevz:
This PR adds two test cases where the max codesize contract is created, implementing #985.
The two new test files implement CREATE and CREATE2, both have 4 test cases each, where two are positive (codesize <= max codesize, contract is created), and two negative (codesize > max codesize, contract is not created).

PR is not ready since the last test case mentioned in #985 is not yet implemented.


marioevz:
Added max codesize contract with selfdestruct test case.

src/GeneralStateTestsFiller/stCreateTest/CreateMaxCodesizeSuicideFiller.yml
```diff
@@ -0,0 +1,79 @@
...
+      code: ''
+      nonce: '0'
+      storage: {}
+
+    00000000000000000000000000000000deadbeef:
+      balance: 2
+      code: |
+        :yul
+        {
+          // initcode: { push2(0x6000), dup1, push1(0x0c), push1(0x00), codecopy, push1(0x00), return,
```
winsvega, line 24: need an example where it does not suicide. to see that this code does actually creates a contract with max allowed codesize

winsvega:
also tests are not regenerated (new tests)

src/GeneralStateTestsFiller/stCreateTest/CreateMaxCodesizeSuicideFiller.yml
marioevz: I think that should be covered by CreateMaxCodesize and Create2MaxCodesize, they both have a test case where the codesize is exactly 0x6000, and another pair of test cases where the codesize is exactly 0x6001.

winsvega: In Andrew's tests it was very clear. here is a bit hard to read

src/GeneralStateTestsFiller/stCreateTest/CreateMaxCodesizeFiller.yml
```diff
@@ -0,0 +1,89 @@
...
+    nonce: '0'
+    to: 00000000000000000000000000000000deadbeef
+    secretKey: "45a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065ff2d8"
+    value:
+    - 0
+
+  expect:
+    - indexes:
+        data:
+        - !!int 0
```
winsvega, line 57: you can use labels

winsvega:
I think Adrew pretty much covered this cases.

marioevz:
Agree that #986 covers this already. Closing.

PR #990, opened 2021-11-17
Tests for EIP-3860: Limit and meter initcode

gumb0:
https://eips.ethereum.org/EIPS/eip-3860
Geth implementation in https://github.com/ethereum/go-ethereum/pull/23847


jochem-brouwer:
Are these tests up to date? The EIP seems to state that once we violate the max initcode on a CREATE or CREATE2, we put 0 on the stack but do not consume all gas (not sure if this is correct?). However, in these tests, that seems to be the case..?

axic:
I don't think these have been updated yet.

jochem-brouwer:
@axic sorry to bother you on this one again, but I think that the EIP is now unclear the gas costs regarding this one. If I try to `CREATE` or `CREATE2` a new contract and the `length` parameter is larger than the max init code size, then what gas cost should be charged? I think this logically leads to multiple options:

1. The base fee of the `CREATE` opcode. We can pop the values from stack, if the length parameter > max init code size then charge the gas and push 0 on the stack
2. Calculate the entire gas cost to start executing the message. So this would involve the initcode jump analysis cost (introduced by the EIP), memory expansion, access costs and hash costs (CREATE2). So this is the upfront gas cost before starting to execute the initcode of the `CREATE` message.
3. Consume the provided gasLimit of the message

I am not sure if this is practically justified, but for gas costs, (1) seems to be the most logical one to me. This charges the CREATE base cost, but we just pop the codesize length item from stack, check if it is too large, and then just continue. (2) seems overkill since we do not execute a message at all (so no memory expansion, access, hashing or jumpdest analysis is needed) and (3) seems extremely overkill with the same arguments as (2).

Regarding the EIP: for me, it is hinting at (2) or possibly (1), but this is not very clear. So the question is: what gas costs should be charged when the initcode is larger than the max init code?

gumb0:
@jochem-brouwer In the tests currently if initcode is larger then limit, 0 is pushed on stack and all gas given to CREATE/CREATE2 (63/64th of caller's gas) is consumed.

We're reconsidering the gas cost of failure and will clarify in EIP.

gumb0:
I've updated now the tests with the rule that all gas is _not_ consumed when initcode is larger than limit, this is the probable direction EIP spec will go.

The cost of CREATE in this case is base static cost + memory expansion cost + initcode charge.
(+ hashing charge for CREATE2)
It is similar to the failures due to call stack limit and not enough balance to transfer.

jochem-brouwer:
Great, this makes sense, thanks @gumb0! :)

src/GeneralStateTestsFiller/stCodeSizeLimit/creationTxInitCodeSizeLimitFiller.yml
```diff
@@ -0,0 +1,97 @@
...
...000000310905080505b505660606040523615610074576000357c010000000000000000000000000000000000000000000000000000000090048063a3b9f3e514610079578063b50345b61461008d578063d4f639ea146100a1578063d9bb316f146100b5578063e22a5d5b146100c9578063f983c0fa146100dd57610074565b610002565b346100025761008b60048050506100f1565b005b346100025761009f6004805050611097565b005b34610002576100b3600480505061203d565b005b34610002576100c76004805050612fe3565b005b34610002576100db6004805050613f89565b005b34610002576100ef6004805050614f2f565b005b60007310000000000000000000000000000000000000019050805073100000000000000000000000000000000000000190508050731000000000000000000000000000000000000002905080507310000000000000000000000000000000000000029050805073100000000000000000000000000000000000000390508050731000000000000000000000000000000000000003905080507310000000000000000000000000000000000000049050805073100000000000000000000000000000000000000490508050731000000000000000000000000000000000000005905080507310000000000000000000000000000000000000059050805073100000000000000000000000000000000000000690508050731000000000000000000000000000000000000006905080507310000000000000000000000000000000000000079050805073100000000000000000000000000000000000'
```
winsvega: are this two samples identical? 
you can have 2 different expect sections of the same data but for different forks, saving the space

src/GeneralStateTestsFiller/stCodeSizeLimit/createInitCodeSizeLimitFiller.yml
```diff
@@ -0,0 +1,120 @@
...
+    - indexes:
+       data: ':label valid'
+       gas: !!int -1
+       value: !!int -1
+      network:
+        - 'London'
+      result:
+       a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+          nonce: 1
+          balance: 198932050
```
winsvega: its better not to check the balance as the gasPrice change a lot by other eips 
you can measure gas before the code you want to check and after the execution and sstore the difference. 
msotre(1, gas()) 
... code ...
msotre(10, gas()) 
sstore(1, sub(mload(1), mload(10)))

src/GeneralStateTestsFiller/stCodeSizeLimit/createInitCodeSizeLimitFiller.yml
```diff
@@ -0,0 +1,120 @@
...
+       b94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+          storage:
+              '0': 'f1ecf98489fa9ed60a664fc4998db699cfa39d40'
+              '1': 1
+       f1ecf98489fa9ed60a664fc4998db699cfa39d40:
+          balance: 0
+          code: '0x00'
```
winsvega: empty code both for valid and invalid creations ?

gumb0: (it's not empty, but one byte `00` of code deployed) and yes before EIP result is the same for valid and invalid.

With EIP account is not created for invalid initcode.

gumb0: > (it's not empty, but one byte `00` of code deployed) and yes before EIP result is the same for valid and invalid.
> 
> With EIP account is not created for invalid initcode.

src/GeneralStateTestsFiller/stCodeSizeLimit/creationTxInitCodeSizeLimitFiller.yml
gumb0: As comments say, they are different in size. (`invalid` is one byte longer than `valid`)

src/GeneralStateTestsFiller/stCodeSizeLimit/createInitCodeSizeLimitFiller.yml
winsvega: I mean isn't it better to deploy some code so not to mistake the created contract with smth else.

See887777 (approved):

src/GeneralStateTestsFiller/stCodeSizeLimit/createInitCodeSizeLimitFiller.yml
gumb0: I reworked it using yul variables and `gas()`

src/GeneralStateTestsFiller/stCodeSizeLimit/createInitCodeSizeLimitFiller.yml
gumb0: I made it deploy the codecopied init code itself (`0x600a80600080396000f3`)

jochem-brouwer:
It does not seem that it is being tested if CREATE/CREATE2 charges the initcode per words gas (of 2 gas per 32-byte word).

What I would also suggest is to add a test which CREATE/CREATE2 a contract, and inside this contract the GASLIMIT is stored (this verifies that the initcode cost is deducted before forwarding the 63/64 of the remaining gas)

src/GeneralStateTestsFiller/stCodeSizeLimit/creationTxInitCodeSizeLimitFiller.yml
```diff
@@ -0,0 +1,97 @@
...
+       a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+          balance: 195600120
+          nonce: 1
+       6295ee1b4f6dd65047762f924ecd367c17eabf8f:
+          balance: 0
+          nonce: 1
+          code: '0x60606040523615610074576000357c01'
+          storage: {}
+
+      # TODO this transaction is considered invalid by the latest EIP rules, it is checked by transaction tests
```
winsvega: better to have it in state tests too
especially we can check the tx exception

src/GeneralStateTestsFiller/stCodeSizeLimit/creationTxInitCodeSizeLimitFiller.yml
```diff
@@ -0,0 +1,97 @@
...
+          nonce: 1
+          code: '0x60606040523615610074576000357c01'
+          storage: {}
+
+    - indexes:
+       data: ':label valid'
+       gas: !!int -1
+       value: !!int -1
+      network:
+        - 'London+3860'
```
winsvega: need to create stEIP3860  folder and generate from those. 
Merge+3860 forkname

winsvega (changes requested):
make a folder stEIP3860
change the forkname Merge+3860
check invalid transaction exception in state tests.

src/GeneralStateTestsFiller/stEIP3860/creationTxInitCodeSizeLimitFiller.yml
```diff
@@ -0,0 +1,89 @@
...
+          nonce: 1
+          code: '0x60606040523615610074576000357c01'
+          storage: {}
+
+    - indexes:
+       data: ':label invalid'
+      network:
+        - 'Merge+3860'
+      expectException:
+        'Merge+3860' : 'TR_InitCodeLimitExceeded'
```
gumb0, line 84: Requires addition to retesteth config:
```
      "TR_InitCodeLimitExceeded" : "max initcode size exceeded:",
```

winsvega:
ready?

gumb0:
Ready, I've noted what would be nice to add later in https://github.com/ethereum/tests/issues/1101

PR #992, opened 2021-11-19
Test Create Max Codesize Contracts and OOG

marioevz:
This PR adds test file with several test cases where contracts are created with a max codesize then:
- The current call stack OOGs
- The calling function calls the contracts to self-destruct
- A combination of both

Two combinations of the test cases with low contract count (10) and high contract count (250) were created.
The low contract call test case already uses 50 million gas, and therefore the high contract count might be too unrealistic.
Please advise on whether we should keep both types.


src/GeneralStateTestsFiller/stCreateTest/CreateOOGafterMaxCodesizeFiller.yml
```diff
@@ -0,0 +1,337 @@
...
+    nonce: '0'
+    to: 00000000000000000000000000000000000c0dea
+    secretKey: "45a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065ff2d8"
+    value:
+    - 0
+
+  expect:
+
+    - indexes:
+        data: !!int 0
```
winsvega: can you comment what happens in this case? which scenario

marioevz: This is scenario:
`A stores 1:1, A calls B, B create contracts with maxcodesize 10 times (no oog upto here), then B oogs the call. creations reverted. But A still stores 1:1 as it is the subcall that OOGed`

We are passing the following parameters:
1) 0x00, the top level contract A creates zero contracts using delegateCall (to reuse the contract creating code)
2) 0x0a, the 2nd level contract B creates 10 contracts
3) 0x01, the contract B will issue an invalid opcode after creating the contracts to induce the OOG.
4) 0x00, contract A calls zero contracts to perform self-destruct (since the Contract B reverted anyway).

winsvega: I mean in the test
ah, here A stores 1:1  after all the operations, not before.

marioevz: I added comments to each contract (0xc0dea, 0xc0deb), also comments to each test case.
Let me know if this makes it more readable.

winsvega: wait, if you do sstore inside delegatecall (which actually I meant not a delegatecall there) that goes oog
why would sstore be still valid?

winsvega: ah, first you do another delegate subcall I see.
so the sstore here basically means that the first delegate call worked. 

I use sstore to signal that contract code has actually worked. 
so you can even put sstore (to a different index) in each contract that you call (at the end of the call) to verify that those have worked correctly

src/GeneralStateTestsFiller/stCreateTest/CreateOOGafterMaxCodesizeFiller.yml
```diff
@@ -0,0 +1,347 @@
...
+        # 00000000000000000000000000000000000c0deb contracts:
+        # 1
+        76e2310dc9277b22ffceb40d49229131674ebe02:
+          shouldnotexist: 1
+        # 10
+        b3210b741a5dfbddc1636521965b3558defa3e60:
+          shouldnotexist: 1
+
+    - indexes:
+        data: !!int 1
```
winsvega: what I mean there is a comment in expect section to have an idea what case is happening here.

src/GeneralStateTestsFiller/stCreateTest/CreateOOGafterMaxCodesizeFiller.yml
```diff
@@ -0,0 +1,347 @@
...
+          storage:
+            0x1: 0x1
+
+        00000000000000000000000000000000000c0deb:
+          nonce: 1
+          storage: {}
+
+        # 00000000000000000000000000000000000c0dea contracts:
+        # 1
+        396b47fae89161b1959b5a833acf816be526b860:
```
winsvega, line 219: you mean the ones created through delegate?

src/GeneralStateTestsFiller/stCreateTest/CreateOOGafterMaxCodesizeFiller.yml
```diff
@@ -0,0 +1,347 @@
...
+        bade62b355fe6b7117f4f7c913321b318ca3a4da:
+          storage: {}
+        # 247
+        d8a8f3569a1d76027f9ece5010489576897014ea:
+          storage: {}
+        # 248
+        fcdccc3b46b9fbe71221061091e8fe82b77e02f2:
+          storage: {}
+        # 249
+        67262aa10552371ec665c2f90c56d9d65c16715e:
```
winsvega, line 294: can put sstore codesize here to verify the size of deployed code

src/GeneralStateTestsFiller/stCreateTest/CreateOOGafterMaxCodesizeFiller.yml
marioevz: Added comments to each expected contract.

src/GeneralStateTestsFiller/stCreateTest/CreateOOGafterMaxCodesizeFiller.yml
marioevz: Added comments to each expected contract, I think should be much clearer now but let know if it's ok.

src/GeneralStateTestsFiller/stCreateTest/CreateOOGafterMaxCodesizeFiller.yml
marioevz: Changed the created max codesize contracts to:
- Always do sstore(0, codesize()), so we get the codesize in storage key 0, and verify their size
- Self-destruct if there is calldata (any value)

And we can now call the contracts twice, first to sstore the codesize, and second to self-destruct if necessary.

marioevz:
Added some more improvements, labels, and clearer comments, let me know if it looks ok.

winsvega:
it's still a bit twisted of a test. but it does what we intended.

PR #1004, opened 2021-12-13
Shift tests along signed integer boundaries

shemnon:
Tests shift operations where sign bits on major integer types (8, 16,
32, 64, 128, and 256 bits) is set to true.


shemnon:
This is a reference test exercising CVE-2021-41272

PR #1008, opened 2022-01-07
Add state tests for RANDOM opcode

marioevz:
Requires: ethereum/retesteth/pull/160

This PR adds state tests for the RANDOM opcode.

Draft only since the required retesteth changes are still not merged.

### DIFF
--- a/GeneralStateTests/stMerge/randomOpcode.json
+++ b/GeneralStateTests/stMerge/randomOpcode.json
@@ -1,0 +1,98 @@
+{
+    "randomOpcode" : {
+        "_info" : {
+            "comment" : "",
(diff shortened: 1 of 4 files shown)


axic:
How are these tests different to tests around `DIFFICULTY`?

marioevz:
> How are these tests different to tests around `DIFFICULTY`?

They are not that much different at the moment. I would recommend using the [hive tests](https://github.com/ethereum/hive/tree/master/simulators/ethereum/engine) to better test PREVRANDAO which comes from the Engine API.

winsvega:
this makes sense as blockchain test as random is a property of blockchain header

winsvega:
The case is covered in bc tests

PR #1013, opened 2022-02-01
signextend(31,+/- 1)

qbzzt:


jochem-brouwer:
I am convinced that these extra cases cover the case I brought up! Thanks!

jochem-brouwer (approved):

PR #1023, opened 2022-04-04
feat(example): Added "not supported in this fork"

qbzzt:
To stExample tests (that use a transaction type > 0)


winsvega:
thanks!

PR #1047, opened 2022-06-16
Separated out the London burnVerify

qbzzt:
In London there is a 2 ETH block reward we need to account for. Post-merge there isn't (in the execution layer).


winsvega:
Why there was a slight burn variation?
Because of the gas used to store mining reward variable?

qbzzt:
Yes. I didn't think of it at when I submitted the PR, but in London I create an ETH variable with 10^18 and multiply that by two. In Merge I just have zero in that variable. At least, I think that's the reason - I could spend the time to verify if you'd like.

winsvega:
well, as long as clients agree we don't really care why it happened

PR #1050, opened 2022-06-29
state tests: Create transaction at Frontier

chfast:
Generate any create transaction for Frontier and Homestead. The cost of
create transaction is different in Frontier than in Homestead and later.

### DIFF
--- a/src/GeneralStateTestsFiller/stTransactionTest/CreateTransactionSuccessFiller.json
+++ b/src/GeneralStateTestsFiller/stTransactionTest/CreateTransactionSuccessFiller.json
@@ -9,18 +9,17 @@
             "previousHash" : "5e20a0453cecd065ea59c37ac63e079ee08998b6045136a8ce6635c7912ec0b6"
         },
         "expect" : [
-    	    {
+            {
                 "indexes" : {
                     "data" : -1,
                     "gas" : -1,
                     "value" : -1
                 },
-                "network" : [">=Istanbul"],
+                "network" : ["Frontier","Homestead",">=Istanbul"],
                 "result" : {
                     "6295ee1b4f6dd65047762f924ecd367c17eabf8f" : {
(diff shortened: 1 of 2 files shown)


src/GeneralStateTestsFiller/stTransactionTest/CreateTransactionSuccessFiller.json
```diff
@@ -9,18 +9,17 @@
...
         "expect" : [
-    	    {
+            {
                 "indexes" : {
                     "data" : -1,
                     "gas" : -1,
                     "value" : -1
                 },
-                "network" : [">=Istanbul"],
+                "network" : ["Frontier","Homestead",">=Istanbul"],
```
winsvega, line 18: we have this test for Frontier and Homestead in legacy, why ?

chfast: I didn't know they existed. Anyway, they have different format so I'm not able to use these.

winsvega: because of 'txbytes' and missing 'baseFee'? 
baseFee is not used on those networks. and 'txbytes' constructable from transaction data. (transaction) 
but yes that requires signing. 

just in case how many cases you would want to use from legacy networks. I try to avoid tests for older networks in test repo.

chfast: I manually added sender and corrected the difficulty: https://github.com/ethereum/legacytests/pull/3. Now I plan to use all of the legacy tests.

winsvega: loos fine. ok. but no need to change fillers there.

PR #1051, opened 2022-06-30
Add touchAndGo state test

chfast:
The empty account 0xe4971 is being called (with gas 0). The test checks
if the touched empty account is removed in the end and enabling point of
this feature.


src/GeneralStateTestsFiller/stCallCodes/touchAndGoFiller.yml
```diff
@@ -0,0 +1,49 @@
...
+    currentNumber: 1
+    currentTimestamp: 1000
+    previousHash: 5e20a0453cecd065ea59c37ac63e079ee08998b6045136a8ce6635c7912ec0b6
+  expect:
+  - indexes:
+      data: !!int -1
+    network:
+    - '>=Frontier'
+    result:
+      095e7baea6a6c7c4c2dfeb977efac326af552d87:
```
winsvega: add 
```
00000000000000000000000000000000000e4971:
       shouldnotexist: '1'
```

chfast: Why? If it exist or not depends on the "network".

winsvega: because this is what you test, that the address is cleaned. if it is not mentioned in result of expect section then no checks are done during filling, only the stateRoot hash will be verified but it might happen that root hash contain the result we don't expect. 

you can have different expect sections for different networks

chfast (commented):
I have fixed the filler (so geth now passes the test it generated).
But I have enabled legacy tests for evmone and this case is covered there so not sure if you still want it.

winsvega:
It must be covered on latest fork too. If its only triggered in legacy we must add this

chfast:
> we must add this

Add what?

winsvega:
the test case. if the bug triggered only by legacy tests but not anymore by develop tests.

src/GeneralStateTestsFiller/stCallCodes/touchAndGoFiller.yml
```diff
@@ -0,0 +1,61 @@
...
+        balance: 0
+      00000000000000000000000000000000000e4971:
+        balance: 0
+        nonce: 0
+  - indexes:
+      data: !!int -1
+    network:
+      - EIP158
```
winsvega: `>=EIP158`

winsvega (changes requested):
Looks good, please always put `>=`

here `>=EIP158` so the test would be generated on new versions of evm too.

chfast:
> the test case. if the bug triggered only by legacy tests but not anymore by develop tests.

I cannot tell any more because this was a week ago. I believe this may have been covered by legacy tests implicitly but not by the develop tests. But I cannot guarantee the claim is true.

winsvega:
its ok. run --filltests --fillchain

winsvega (approved):

PR #1054, opened 2022-07-03
Update the docs for state transition tests with `currentRandom`

qbzzt:


docs/test_filler/test_env.rst
```diff
@@ -60,9 +60,10 @@ transaction tests.
 Name in Env Section Meaning
 =================== ========================
 currentCoinbase     beneficiary of mining fee
-currentDifficulty   difficulty of previous block
+currentDifficulty   difficulty of previous block (pre Merge)
+currentRandom       random value that is supposed to come from the beacon chain (post Merge)
```
chfast, line 64: In the generated tests both `currentDifficulty` and `currentRandom` co-exist. Are you sure this is good idea? Only one of them can be present in practice.

Also, some people put a lot of effort to avoid the "random" name. Where did you get it from?

qbzzt: These is the name @winsvega selected in `retesteth`. @winsvega , what do you think?

@chfast , what's the problem with the name "random"?

chfast: The `prevrandao` should be better name. https://eips.ethereum.org/EIPS/eip-4399.

winsvega: this is done to avoid confusion what is difficulty and what is random. becaue difficulty is expected to be 0
also test env describe env for multiple forks. you just take what you need for current network (fork).

as for random, random is just much simplier to type and remember than prevrandao

PR #1063, opened 2022-07-13
Add state test: double selfdestruct touches different beneficiaries

chfast:
A single contract can execute SELFDESTRUCT multiple times using by being
called multiple times. The second and later SELFDESTRUCTs have little
effect but can touch some new beneficiary addresses.


src/GeneralStateTestsFiller/stSystemOperationsTest/doubleSelfdestructTouchFiller.yml
```diff
@@ -0,0 +1,110 @@
...
+      result:
+        a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+          nonce: 1
+        000000000000000000000000000000000000c0de:
+          nonce: 0
+          balance: 0
+          storage: {}
+        0000000000000000000000000000000000e49701:
+          shouldnotexist: true
+        0000000000000000000000000000000000e49702:
```
winsvega, line 93: so it get deleted because it is being touched by the second selfdestruct?

src/GeneralStateTestsFiller/stSystemOperationsTest/doubleSelfdestructTouchFiller.yml
```diff
@@ -0,0 +1,110 @@
...
+
+    # Tx target contract
+    000000000000000000000000000000000000c0de:
+      nonce: 0
+      balance: 0
+      storage: {}
+      code: |
+        :yul {
+          let r1 := call(70000, 0xdead, callvalue(), 0, 0, 0, 0)
+          let r2 := call(70000, 0xdead, 0, 0, 0, 0, 0)
```
winsvega: what if you transfer a little balance when running the second time as well?

winsvega:
This is an interesting case. What if when you call the contract second time you give it some balance as well? (3rd case on top of what already covered in this test)

src/GeneralStateTestsFiller/stSystemOperationsTest/doubleSelfdestructTouchFiller.yml
chfast: Ah, good point. You can increase the balanace of already selfdestructed account. This is also not covered by tests because I can see my implementations is buggy. I will add more variants then.

chfast: Done.

jochem-brouwer:
OK this is also a super great test :) Thanks for this one, the selfdestructs are not super intuitive.

PR #1068, opened 2022-08-02
Add a state test checking if create address is warm

chfast:
Invokes failing CREATE (because 0xEF code cannot be deployed in EIP-3541) and checks
if the create address is considered warm in the follow up call as required by EIP-2929.


winsvega:
I think this is covered with 
https://github.com/ethereum/tests/pull/835/files#diff-784e3a55679246c0562a4eaf82db111a591dba19e0d1d5ff40aff78153a74d22R24

does those tests not trigger the issue?

src/GeneralStateTestsFiller/stCreateTest/CreateAddressWarmAfterFailureEFFiller.yml
```diff
@@ -0,0 +1,103 @@
...
+      - 1
+
+  expect:
+    - indexes:
+        data: !!int -1
+        gas: !!int -1
+        value:
+          - !!int 0
+      network:
+        - London
```
winsvega: also >= London, to generate on latest evm version

chfast:
> I think this is covered with https://github.com/ethereum/tests/pull/835/files#diff-784e3a55679246c0562a4eaf82db111a591dba19e0d1d5ff40aff78153a74d22R24
> 
> does those tests not trigger the issue?

No. This test is combination of EIP-3541 and EIP-2929: after a create fails because of EIP-3541 we additionally check (not covered before) if the create address is warm as required by EIP-2929.

PR #1069, opened 2022-08-06
Added test for ecrecover with an unusual buffer size

qbzzt:
Anything from 0 to 0xF0


src/GeneralStateTestsFiller/stPreCompiledContracts2/ecrecoverShortBuffFiller.yml
```diff
@@ -0,0 +1,92 @@
...
+             // The expected retval is one, so to avoid specifying every length
+             // in the expect: section we subtract one.
+             sstore(len, sub(sload(len), 1))
+
+             // The returned address should always be zero, because it is not a valid signature
+             sstore(add(0x1000,len), mload(0x100))
+          }
+        }
+      nonce: 1
+      storage: {}
```
winsvega: You expect 0, you should set storage to be dirty

Ah, you do it dynamically, I don't feel safe in case the code compilation issue happens, better you check non zero value, at least in some

winsvega (approved):

jochem-brouwer:
Thanks for this one :+1: :smile:

PR #1073, opened 2022-08-10
Test warm address after create failure (too expensive)

chfast:
Invokes failing CREATE and checks if the create address is considered warm in the follow up call as required by EIP-2929.

This is similar to #1068, but the failure reasons are different:
- initcode failure
- deploy code is too big
- deploy code is too expensive
- the sender nonce is maximum


src/GeneralStateTestsFiller/stCreateTest/CreateAddressWarmAfterInitcodeFailureFiller.yml
```diff
@@ -0,0 +1,96 @@
...
+      storage: {}
+      code: |
+        :yul
+          object "C" {
+            code {
+              let initcode_size := datasize("initcode")
+              datacopy(0, dataoffset("initcode"), initcode_size)
+              sstore(0, create(0, 0, initcode_size))
+              // Call the create address.
+              sstore(1, call(0, 0xd4e7ae083132925a4927c1f5816238ba17b82a65, callvalue(), 0, 0, 0, 0))
```
winsvega: Can you measure in expect section the gas price of call before and after it is warmed up.
By 2929. You can have smth like 
 sstore(10, gas())
Call()
Sstore(10, sub(sload(10), gas()))

src/GeneralStateTestsFiller/stCreateTest/CreateAddressWarmAfterNonceMaxFiller.yml
```diff
@@ -0,0 +1,88 @@
...
+        - ">=London"
+      result:
+        a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
+          nonce: 1
+        00000000000000000000000000000000000c0deC:
+          nonce: "0xffffffffffffffff"    # Create failed because nonce value is maximum.
+          storage:
+            0: 0    # Create failed
+            1: 1    # Call to create address successful
+        f4eb6941ffbbffd112b83bda0d43f1c9d98d450a: # Create address
```
winsvega, line 133: Here as well. The gas price of call we expect to change (become lower) since its warmed up

winsvega (changes requested):
Can you measure in expect section the expected call gasprice for before and after warmup.

src/GeneralStateTestsFiller/stCreateTest/CreateAddressWarmAfterInitcodeFailureFiller.yml
qbzzt: I added the code to check the costs of the `create` and `call`. But the effect of sending value is 34k gas, which is a lot more than the cost of cold vs. warm. See https://www.evm.codes/#f1 (or the comments I added to the filler).

qbzzt:
@winsvega, I added the requested changes on the two tests where you requested them. I wasn't sure if you wanted the other two as well or not.

chfast:
I'm not sure what has changed here. Do you still need anything from me?

winsvega:
No, we got the idea, Ori now works on test cases doing this warming up check. 
Ori, have you moved all this cases in a new PR, is there something to merge from this PR?

winsvega:
moved to https://github.com/ethereum/tests/pull/1077

chfast:
> moved to #1077

Fine, except you removed me as co-author of this contribution.

PR #1077, opened 2022-08-29
Test when an address is warm after creating a contract for it failed

qbzzt:
Data: 0 (REVERT inside the constructor), 1 (constructor returns >max size), 2 (invalid opcode in constructor), 6 (contract starts with 0xEF, which is forbidden)
   First call to created address is warm (although the address isn't there, so if we try to send it ETH it needs to get created)

Data: 3 (OOG in the CREATE), 4 (OOG after the CREATE, still in the call to the contract that has CREATE), 5 (Max nonce) 
   First call to created address is cold

Replacement for #1073


winsvega:
why it removes a file?

src/GeneralStateTestsFiller/stCreateTest/CreateAddressWarmAfterFailFiller.yml
```diff
@@ -0,0 +1,458 @@
...
+  
+  transaction:
+    data:
+      - :abi fail(uint) 0   # Revert in the constructor
+      - :abi fail(uint) 1   # Code too big
+      - :abi fail(uint) 2   # Constructor has invalid opcode
+      - :abi fail(uint) 3   # Constructor OOG because it returns the max contract size
+      - :abi fail(uint) 4   # OOG after the constructor
+      - :abi fail(uint) 5   # Nonce too high to create an account
+      - :abi fail(uint) 6   # Contract starts with 0xEF, which is not allowed
```
winsvega: how about successfull create, just for comparison to see if it warms up the address ?

winsvega (approved):
very good, easy to read and many comments. 
just why it removes one existing test isn't you adding a new file?

and now need a copy of this test but for create2

qbzzt:
> just why it removes one existing test isn't you adding a new file?

Because that test is having the created contract start with 0xEF, which fits great with this test file.

qbzzt:
> and now need a copy of this test but for create2

It's easiest for me to add it to the same test file. Doing that now (may finish today, or it might happen tomorrow - the address that should be created is harder to calculate because it depends on the initialization code)

qbzzt:
@chfast , I didn't mean to remove you - I'm just not very good with github rebase. If you just want to be on the PR, can you do some meaningless edits on https://github.com/ethereum/tests/edit/20220827-create-addr-warm/BlockchainTests/GeneralStateTests/stCreateTest/CreateAddressWarmAfterFail.json?pr=%2Fethereum%2Ftests%2Fpull%2F1077 so you'll be an author? 

I already added your name to the comments.

chfast:
I mostly care about commit authorship so I can track my work easily. I added `Co-authored-by` to the first commit. Try to keep it if you are going to squash commits.

PR #1079, opened 2022-09-08
Tests for EIP-4895: Beacon chain push withdrawals as operations

marioevz:
Withdrawals test fillers based on https://eips.ethereum.org/EIPS/eip-4895


winsvega:
Implemented in pyspecs

PR #1082, opened 2022-09-21
Tests for EIP-3651: Warm COINBASE

marioevz:
## Changes included
Adds tests for Shanghai fork upgrade EIP-3651.

Coinbase account is now warm and is by default added to the `accessed_addresses` list when the transaction starts execution.

We test accessing the account with the following opcodes:
- EXTCODESIZE
- EXTCODECOPY
- EXTCODEHASH
- BALANCE
- CALL
- CALLCODE
- DELEGATECALL
- STATICCALL

Requires `Shanghai` fork to be added to retesteth.


chfast:
What is the status of this?

marioevz:
@winsvega I refilled these tests with latest ethereum/retesteth@faead84 and updated the yul code because it was being optimized by `solc`.
I think this is ready for merge.

winsvega:
Add vectors to /src/readme

marioevz:
> Add vectors to /src/readme

Added, let me know if the format of the descriptions is ok.

src/README.md
```diff
@@ -101,3 +101,19 @@
 - ID: **EOF1C0017** legacy contact > CREATE[2] > legacy init code > nonsense
 - ID: **EOF1C0018** legacy contact > CREATE[2] > eof1 init code > nonsense (fail)
 - ID: **EOF1C0019** legacy contact > CREATE[2] > nonsense init code (fail)
+
+
+## Warm COINBASE EIP-3651 
+
+- ID: **WRMCB0001** check EXTCODESIZE gas cost on coinbase
```
winsvega, line 108: the same key should also be mentioned in the test file in comments or in info section comment field. so we can find the test. 
if one test contains many test vectors ids implementation can put a comment right next to the block that executes a scenario (for bc tests) or transaction data that implements it. 
In yml comments supported by format with #
In json fillers any field started with "//" is deleted

winsvega:
ready?

PR #1084, opened 2022-09-28
Add Gas Leak Test Case

marioevz:
Introduces test to verify a Besu consensus issue described here: https://hackmd.io/@shemnon/besu-gas-leak

Test will use call opcodes with gas values in edge values around `2**63` and `2**64` and verify that the EVM does not run out of gas (due to improper subtraction of the remaining gas) or has more remaining gas than before the call.

Verified using besu evmtool versions: [22.7.3](https://hub.docker.com/layers/hyperledger/besu-evmtool/22.7.3/images/sha256-ff70ba4f485c477f3a19654ba1c6fbc8766b21e5c01cf33c2b26957157b5f09d?context=explore) and [22.7.0-RC3](https://hub.docker.com/layers/hyperledger/besu-evmtool/22.7.0-RC3/images/sha256-509da9443ccb0de6387a3525d53b8cb6c36e96a2098c2292a898912ddd7e3e7a?context=explore), test fails on 22.7.0-RC3, and passes on 22.7.3.

Implements #1083

PR #1085, opened 2022-09-28
Gas Leak Tests

shemnon:
Tests for the 63/64 bit gas leak.
Each call requests values on either side of major signed integer boundaries, attempting to trigger a mis-handling of unsigned data as signed data.


shemnon:
Just saw #1084.  Its interesting how we both settled on the same principal approach to test the leak, with test suite only differences.

marioevz (approved):
Looks good to me, I think we could merge this one and I'll close #1084

shemnon:
@qbzzt bump

PR #1088, opened 2022-11-02
Check what happens when an EOA empties itself

qbzzt:


jochem-brouwer:
Related; https://github.com/ethereumjs/ethereumjs-monorepo/pull/2383

jochem-brouwer (commented):
I quickly checked but it seems that these tests cover `EXTCODEHASH` and `EXTCODESIZE` (and not two other edge cases which we found).

The extra cases involved in the discovered bugs in the aforementioned PR are:

(1) empty EOA account, but `CALL` the account with nonzero value (this charged "create new account" gas without the fix, while it should not charge this)

(2) The same situation, except `SELFDESTRUCT` the contract to the emptied EOA (i.e. tx.origin) and the contract sends nonzero value (should thus not charge "create new account" gas)

qbzzt:
Oops, you're right. I'll fix it, probably this weekend (sorry, life during
the work week is crazy)

jochem-brouwer:
No problem at all, I'm glad we can participate in the coverage of the tests :) Let me know if you need any help or validation!

qbzzt:
> Let me know if you need any help or validation!

I added those cases.

1. `CALL` without a value is 100 (plus 27 for misc expenses)
2. `CALL` with a value is 6800 (plus 27 for misc expenses). 100 for the call, 9000 for the value call - 2300 for the gas stipend the EOA doesn't use
3. `SELFDESTRUCT` is 7600. 31 is probably for misc expenses. 7600 is for:
    - `CALL` to `0xDEAD`, which is cold: 2600
    - `SELFDESTRUCT`, which is 5000. 
    No extra cost for adding to the EOA, regardless of whether the balance on it is zero (d: *, g: 0, v: 0) or 100 (d: *, g: 1, v: 0)

holgerd77:
> > Let me know if you need any help or validation!
> 
> I added those cases.
> 
>     1. `CALL` without a value is 100 (plus 27 for misc expenses)
> 
>     2. `CALL` with a value is 6800 (plus 27 for misc expenses). 100 for the call, 9000 for the value call - 2300 for the gas stipend the EOA doesn't use
> 
>     3. `SELFDESTRUCT` is 7600. 31 is probably for misc expenses. 7600 is for:
>        
>        * `CALL` to `0xDEAD`, which is cold: 2600
>        * `SELFDESTRUCT`, which is 5000.
>          No extra cost for adding to the EOA, regardless of whether the balance on it is zero (d: *, g: 0, v: 0) or 100 (d: *, g: 1, v: 0)

Thanks guys, that is really super-valuable that you are so quickly with adding new test cases!

Super cool! 🤩 👍

PR #1091, opened 2022-11-09
EIP-1153: Transient Storage tests

moodysalem:


marioevz:
I haven't looked at all the tests but one thing that came to mind when hearing about this EIP is that the clients need to implement a new data structure that might be ephemeral but we might still need to test the maximum number of transient elements it can realistically contain within the span of a single transaction, just to make sure that this structure does not break.

moodysalem:
> I haven't looked at all the tests but one thing that came to mind when hearing about this EIP is that the clients need to implement a new data structure that might be ephemeral but we might still need to test the maximum number of transient elements it can realistically contain within the span of a single transaction, just to make sure that this structure does not break.

Mentioned in the EIP security considerations: https://eips.ethereum.org/EIPS/eip-1153#security-considerations

And it should be tested in the file `src/GeneralStateTestsFiller/stEIP1153/15_tstoreCannotBeDosdFiller.yml`

src/GeneralStateTestsFiller/stEIP1153/15_tstoreCannotBeDosdFiller.yml
```diff
@@ -0,0 +1,96 @@
...
+      code: "0x"
+      nonce: 0
+      storage: {}
+
+  transaction:
+    data:
+      - data: :abi doCall(uint) 0x200
+        accessList: []
+    gasLimit:
+      - "400000"
```
moodysalem: should probably be much more than 400k, like 30m gas, and `0x200` -> `0x30d40` (200k)

winsvega:
can this EIP be activated by command `Merge+1153`?

snreynolds:
> can this EIP be activated by command `Merge+1153`?

updated, thank you. can now run them off the develop branch of retesteth with the following commands

GeneralStateTests:

```
/dretesteth.sh -t GeneralStateTests/stEIP1153 -- \
 --testpath <path to tests>  --filltests
```

BlockchainTests:

```
./dretesteth.sh -t BlockchainTests/ValidBlocks/bcEIP1153 -- \
 --testpath <path to tests>  --filltests
```

src/GeneralStateTestsFiller/stEIP1153/15_tstoreCannotBeDosdFiller.yml
snreynolds: fixed here https://github.com/ethereum/tests/pull/1091/commits/d1f8ba85bc0c2c4d150aa7390bc56a7913f0a6e5

snreynolds:
@marioevz Would love to get any other thoughts you might have for these tests

winsvega:
ready to merge? 
for the blockchain test,  t8n tool does not construct a blockchain. so instead it is treated like a chain of state tests. (single state in, single state out.) 
the bc test produced here is valid because you don't expect the cache to be present in the next block. 
however if you wanted to make a test that it does present such thing won't work with t8n.

the client that could run it is besu once it has 1153 support.

src/GeneralStateTestsFiller/stEIP1153/03_tloadAfterStoreIs0Filler.yml
```diff
@@ -0,0 +1,63 @@
...
+
+  pre:
+    000000000000000000000000000000000000000A:
+      balance: 1000000000000000000
+      nonce: 0
+      code: |
+        :yul {
+          verbatim_2i_0o(hex"b4", 0, 30)
+          let val := verbatim_1i_1o(hex"b3", 1)
+          sstore(0, val)
```
marioevz: I think it would be a good idea in this test to sstore into a different location than tstore, just to test at the end that the tstore didn't somehow end up being written to real storage.

E.g.
```
tstore(0, 1)
sstore(1, <some test result>)
expect:
0x00: 0 # With this, we are testing that tstore didn't get stored in real storage
0x01: <some test result>
```

src/GeneralStateTestsFiller/stEIP1153/04_tloadAfterCallFiller.yml
```diff
@@ -0,0 +1,80 @@
...
+          sstore(1, success)
+        }
+      storage: {}
+    000000000000000000000000000000000000000B:
+      balance: 1000000000000000000
+      nonce: 0
+      code: |
+        :yul {
+          let val := verbatim_1i_1o(hex"b3", 0)
+          sstore(0, val)
```
marioevz: Also `tstore` after this `sstore` with a different value than the calling contract, to test that the `tstore` doesn't modify calling contract transient storage.

src/GeneralStateTestsFiller/stEIP1153/08_revertUndoesTransientStoreFiller.yml
```diff
@@ -0,0 +1,100 @@
...
+            doReenter()
+          }
+
+          default {
+            sstore(2, 1)
+          }
+
+          function doStoreAndRevert() {
+            verbatim_2i_0o(hex"b4", 0, 11)
+            revert(0, 0)
```
marioevz: Is revert the only method used throughout the tests?

If so, I think it would be valuable to add at least one test case with other types of failures such as:
- out of gas
- invalid opcode
- stack underflow

src/GeneralStateTestsFiller/stEIP1153/09_revertUndoesAllFiller.yml
```diff
@@ -0,0 +1,93 @@
...
+          function storeTwiceAndRevert() {
+            verbatim_2i_0o(hex"b4", 0, 5)
+            verbatim_2i_0o(hex"b4", 0, 6)
+            revert(0,0)
+          }
+
+          function doReenter() {
+            verbatim_2i_0o(hex"b4", 0, 4)
+            mstore(0, hex"578290e5")
+            let success := call(0xffff, 0x000000000000000000000000000000000000000A, 0, 0, 32, 0, 0)
```
marioevz: Var name should be `fail`.

src/GeneralStateTestsFiller/stEIP1153/09_revertUndoesAllFiller.yml
```diff
@@ -0,0 +1,93 @@
...
+
+          case 0x578290e5 { // storeTwiceAndRevert()
+            storeTwiceAndRevert()
+          }
+
+          default {}
+
+          function storeTwiceAndRevert() {
+            verbatim_2i_0o(hex"b4", 0, 5)
+            verbatim_2i_0o(hex"b4", 0, 6)
```
marioevz: Maybe also do `tstore(1, x)` and check that on the caller too.

src/GeneralStateTestsFiller/stEIP1153/11_tstoreDelegateCallFiller.yml
```diff
@@ -0,0 +1,75 @@
...
+        :yul {
+          verbatim_2i_0o(hex"b4", 0, 1)
+          let success := delegatecall(0xffff, 0x000000000000000000000000000000000000000B, 0, 32, 0, 0)
+          let v := verbatim_1i_1o(hex"b3", 0)
+          sstore(0, success)
+          sstore(1, v)
+        }
+
+      storage: {}
+    000000000000000000000000000000000000000B:
```
marioevz: Maybe add another contract that does `tstore` but reverts.

src/GeneralStateTestsFiller/stEIP1153/13_tloadStaticCallFiller.yml
```diff
@@ -0,0 +1,90 @@
...
+          }
+
+          case 0x611e535a { // tstore3()
+            tstore3()
+          }
+
+          default {}
+
+          function tstore3() {
+            verbatim_2i_0o(hex"b4", 0, 3)
```
marioevz: We are missing a test case where we `staticcall` and then `tload`, which can be verified by returning the value and then saving it on the caller contract.

src/GeneralStateTestsFiller/stEIP1153/15_tstoreCannotBeDosdFiller.yml
```diff
@@ -0,0 +1,96 @@
...
+            let t := call(0xffff, 0x000000000000000000000000000000000000000A, 0, 0, 64, 0, 0)
+            sstore(0, t) // should revert
+
+            let v := verbatim_1i_1o(hex"b3", 0)
+            sstore(1, v)
+          }
+
+          function doNStores(n) {
+            for { let i := 0 } lt(i, n) { i := add(i, 1) } {
+              verbatim_2i_0o(hex"b4", i, 48)
```
marioevz: Could be interesting to add a similar test to this but skipping some slots in between the stored values.

Also one where the slots are written out of order:
$$0, 2^{256}-1, 1, 2^{256}-2, ...$$

src/GeneralStateTestsFiller/stEIP1153/15_tstoreCannotBeDosdFiller.yml
```diff
@@ -0,0 +1,96 @@
...
+          }
+
+          default {
+            sstore(0, 5)
+          }
+          
+          function doCall(n) {
+            mstore(0, hex"f7d6c0eb") // doNStores(n)
+            mstore(32, n)
+            let t := call(0xffff, 0x000000000000000000000000000000000000000A, 0, 0, 64, 0, 0)
```
marioevz: I think the `0xffff` might be a very low value considering that the transaction in total has 30,000,000 gas.
15 seconds for 65535 gas in operations is in my opinion excessively long.

src/GeneralStateTestsFiller/stEIP1153/16_tloadGasFiller.yml
```diff
@@ -0,0 +1,81 @@
...
+      code: |
+        :yul {
+          let extraOpcodesGas := 5
+
+          // calculate warm sload gas
+          let s := sload(5)
+          let g1 := gas()
+          let s2 := sload(5)
+          let g2 := gas()
+          sstore(0, sub(sub(g1, g2), extraOpcodesGas))
```
marioevz: I don't think it's worth it to include the gas costs for `sload` here, since this is an entirely different test suite, if in the future for some reason the `sload` gas cost changes, this test will fail.

src/GeneralStateTestsFiller/stEIP1153/17_tstoreGasFiller.yml
```diff
@@ -0,0 +1,81 @@
...
+      nonce: 0
+      code: |
+        :yul {
+          let extraOpcodes := 8
+
+          // calculate warm dirty sstore gas
+          sstore(5, 1)
+          let g1 := gas()
+          sstore(5, 2)
+          let g2 := gas()
```
marioevz: Same observation as with previous test.

marioevz (changes requested):
I think it looks good overall.

The DoS test seemed a bit worrying to me since it was not that much gas and can take way too long.

Also there should be more tests using more transient slots than 0, almost all tests only check the zero slot, so we should check other interesting keys too.

holiman:
> won't work with t8n.

Really? I think it should, but if it doesn't, we need to get that fixed.

holiman:
Seems to work
```
[user@work evm]$ ./evm t8n --state.fork=Merged+1153 --input.alloc=./testdata/23/alloc.json --input.txs=./testdata/24/txs.json --input.env=./testdata/25/env.json
INFO [11-30|22:04:22.208] rejected tx                              index=0 hash=92ea4a..2ee941 from=0xa94f5374Fce5edBC8E2a8697C15331677e6EbF0B error="nonce too high: address 0xa94f5374Fce5edBC8E2a8697C15331677e6EbF0B, tx: 172 state: 0"
INFO [11-30|22:04:22.209] rejected tx                              index=1 hash=16b1d9..c4f1da from=0xa94f5374Fce5edBC8E2a8697C15331677e6EbF0B error="nonce too high: address 0xa94f5374Fce5edBC8E2a8697C15331677e6EbF0B, tx: 173 state: 0"
INFO [11-30|22:04:22.209] Trie dumping started                     root=b7360b..e3458f
INFO [11-30|22:04:22.209] Trie dumping complete                    accounts=2 elapsed="51.703µs"
INFO [11-30|22:04:22.209] Wrote file                               file=alloc.json
INFO [11-30|22:04:22.209] Wrote file                               file=result.json
```
As opposed to e.g.: 
```
[user@work evm]$ ./evm t8n --state.fork=Merged+1199 --input.alloc=./testdata/23/alloc.json --input.txs=./testdata/24/txs.json --input.env=./testdata/25/env.json
ERROR(3): failed constructing chain configuration: syntax error, invalid eip number 1199

```

src/GeneralStateTestsFiller/stEIP1153/08_revertUndoesTransientStoreFiller.yml
winsvega: Yes. One thing we were working on are templates for new opcodes - a basic series of scenarious to cover for any new opcode. 

Like oogs, reverts, underflow, static context... 

@qbzzt I think you even did a template generator for some of that

holgerd77:
> The DoS test seemed a bit worrying to me since it was not that much gas and can take way too long.

Has this comment from @marioevz been picked up since it might be important to dig in further here and have a somewhat more solid picture? Mario, could you expand a bit on this by citing the numbers (gas used, runtime, machine) to get some first impression here?

@holiman are you able to give a first-round judgement on this how critical this is/might be respectively how much worth to be deeper investigated?

holiman:
Re the Dos-issue. The description says: `CALL(self, ...), TSTORE(x, y), TSTORE(x + 1, y), TSTORE(x + 2, y), ... ,TSTORE(x + n, y), REVERT, TLOAD(x)`. So essentially, it just does a lot of `TSTORE` ? 

Sure, it might be something we need to look into (it ought not be a problem, but implementations can always be surprising) -- but I don't see why it would be a blocking issue for _this_ pr.

src/GeneralStateTestsFiller/stEIP1153/03_tloadAfterStoreIs0Filler.yml
snreynolds: fixed here https://github.com/ethereum/tests/pull/1091/commits/9a54b73adf145c945bfb72d63513e707b4526131 we now expect storage slot 0 to be 0xffff, the same at the start of the txn. And we now expect storage slot 1 to be 0 which is the loaded value from an unused key

src/GeneralStateTestsFiller/stEIP1153/04_tloadAfterCallFiller.yml
snreynolds: fixed https://github.com/ethereum/tests/pull/1091/commits/2aace0267dce263e8439d44e10cef18397c24e8a

src/GeneralStateTestsFiller/stEIP1153/08_revertUndoesTransientStoreFiller.yml
snreynolds: Added some oog tests here
https://github.com/ethereum/tests/pull/1091/commits/2bcc343480ed47b2dd6cd0bf1957a95aae3447bc
https://github.com/ethereum/tests/pull/1091/commits/c86bd00f9ff7f6938b48f7a3767c18d48e54b6ae

src/GeneralStateTestsFiller/stEIP1153/09_revertUndoesAllFiller.yml
snreynolds: added storing/loading from transient slot 1 https://github.com/ethereum/tests/pull/1091/commits/285c951c37e8b9cbcca24249b7bbc8d495a7b97f

src/GeneralStateTestsFiller/stEIP1153/09_revertUndoesAllFiller.yml
snreynolds: good catch thanks https://github.com/ethereum/tests/pull/1091/commits/285c951c37e8b9cbcca24249b7bbc8d495a7b97f

src/GeneralStateTestsFiller/stEIP1153/11_tstoreDelegateCallFiller.yml
snreynolds: adds revert after delegatecall https://github.com/ethereum/tests/pull/1091/commits/31319495e03410614c38351d6498aa7d6c393435

src/GeneralStateTestsFiller/stEIP1153/13_tloadStaticCallFiller.yml
snreynolds: added tload https://github.com/ethereum/tests/pull/1091/commits/11ffdeb108d5379fa9f4bd45d61aba17d677bea1

src/GeneralStateTestsFiller/stEIP1153/15_tstoreCannotBeDosdFiller.yml
moodysalem: should probably change to `call(gas(), ...`

snreynolds: updated https://github.com/ethereum/tests/pull/1091/commits/c07c1485a9b43874b9ee106ebeb18fdaef98836e

src/GeneralStateTestsFiller/stEIP1153/16_tloadGasFiller.yml
snreynolds: makes sense, fixed https://github.com/ethereum/tests/pull/1091/commits/947ebcaeb3e33266f7af456722057536595db53a & https://github.com/ethereum/tests/pull/1091/commits/2556022df9340b4fe1911a64478885b26f22fb9b

src/GeneralStateTestsFiller/stEIP1153/15_tstoreCannotBeDosdFiller.yml
snreynolds: added some out of order stores here https://github.com/ethereum/tests/commit/75e479fb379d84db383fa51247e654211aad5fe3

src/GeneralStateTestsFiller/stEIP1153/15_tstoreCannotBeDosdFiller.yml
```diff
@@ -0,0 +1,91 @@
...
+  pre:
+    000000000000000000000000000000000000000A:
+      balance: 1000000000000000000
+      nonce: 0
+      code: |
+        :yul {
+          switch selector()
+
+          case 0x883264e8 { // doCall(uint)
+            let n:= div(calldataload(32), 0x100000000000000000000000000000000000000000000000000000000)
```
moodysalem: i don't think you want a div here

src/GeneralStateTestsFiller/stEIP1153/15_tstoreCannotBeDosdFiller.yml
```diff
@@ -0,0 +1,91 @@
...
+          }
+
+          default {
+            sstore(0, 5)
+          }
+          
+          function doCall(n) {
+            mstore(0, hex"f7d6c0eb") // doNStores(n)
+            mstore(32, n)
+            let t := call(gas(), 0x000000000000000000000000000000000000000A, 0, 0, 64, 0, 0)
```
moodysalem: could be more clear that it's self-calling if it's `address()` instead of `0x000000000000000000000000000000000000000A`

src/GeneralStateTestsFiller/stEIP1153/21_tstoreCannotBeDosdOOOFiller.yml
```diff
@@ -0,0 +1,105 @@
...
+  pre:
+    000000000000000000000000000000000000000A:
+      balance: 1000000000000000000
+      nonce: 0
+      code: |
+        :yul {
+          switch selector()
+
+          case 0x883264e8 { // doCall(uint)
+            let n:= div(calldataload(32), 0x100000000000000000000000000000000000000000000000000000000)
```
moodysalem: unnecessary div here

src/GeneralStateTestsFiller/stEIP1153/21_tstoreCannotBeDosdOOOFiller.yml
```diff
@@ -0,0 +1,105 @@
...
+            mstore(32, n)
+            let t := call(gas(), 0x000000000000000000000000000000000000000A, 0, 0, 64, 0, 0)
+            sstore(0, t) // should revert
+
+            let v := verbatim_1i_1o(hex"b3", 0)
+            sstore(1, v)
+          }
+
+          function doNStores(n) {
+            let val := exp(2, 256)
```
moodysalem: 2**256 should return 0 i expect

src/GeneralStateTestsFiller/stEIP1153/21_tstoreCannotBeDosdOOOFiller.yml
```diff
@@ -0,0 +1,105 @@
...
+              verbatim_2i_0o(hex"b4", i, 600)
+              let even := mod(i, 2)
+              let index
+              switch even 
+              case 0 {
+                index := i
+                verbatim_2i_0o(hex"b4", i, 600)
+              }
+              case 1 {
+                index := sub(val, i)
```
moodysalem: this probably still works because of underflow

src/GeneralStateTestsFiller/stEIP1153/21_tstoreCannotBeDosdOOOFiller.yml
```diff
@@ -0,0 +1,105 @@
...
+                 verbatim_2i_0o(hex"b4", index, 600)
+               }
+              default {}
```
moodysalem: no default needed

winsvega:
please move tests to EIPTests folder in BlockchainTests and GeneralStateTests accordingly.

src/GeneralStateTestsFiller/stEIP1153/21_tstoreCannotBeDosdOOOFiller.yml
snreynolds: addressed all the above https://github.com/ethereum/tests/pull/1091/commits/bc0dcbc292b8256d9566b05be2eccab60a2e73c9 &
https://github.com/ethereum/tests/pull/1091/commits/eb0a02818c20be023e6407928054d3478b51b899 & 
https://github.com/ethereum/tests/pull/1091/commits/fcd3575118cc621b2ada3f4d0b88070ea9f81263

src/GeneralStateTestsFiller/EIPTests/stEIP1153/15_tstoreCannotBeDosdFiller.yml
```diff
@@ -0,0 +1,93 @@
...
+
+  pre:
+    000000000000000000000000000000000000000A:
+      balance: 1000000000000000000
+      nonce: 0
+      code: |
+        :yul {
+          switch selector()
+
+          case 0x883264e8 { // doCall(uint)
```
moodysalem, line 28: bit strange how the selector is computed from `doCall(uint)` rather than `doCall(uint256)` which is how solidity would do it, and that `:abi doCall(uint)` on line 72 below takes the signature as a literal instead of parsing it

moodysalem: in other words, this is not consistent with solidity. doesn't really matter, but thought i'd point it out. you also have a function selector computed from `keccak256('doNStores(n)')` which is what made me think of it (using `n` instead of the type in the signature)

winsvega (changes requested):
git reset --soft HEAD~60

Can repack this into smaller number of commits. Like test srcs, test filled separate
We don't need history of editing and moving from folder to folder. 
And its ready to merge into EIPTests folder, then we can pr to make it better ib another thread. And eventually update to cancun or when it's gonna be accepted

snreynolds:
> git reset --soft HEAD~60
> 
> Can repack this into smaller number of commits. Like test srcs, test filled separate We don't need history of editing and moving from folder to folder. And its ready to merge into EIPTests folder, then we can pr to make it better ib another thread. And eventually update to cancun or when it's gonna be accepted

@winsvega Can we close this in favor of https://github.com/ethereum/tests/pull/1194/files which squashes these commits and is updated against latest develop branch

winsvega:
Finally ready to merge. Thanks for the contribution!
https://github.com/ethereum/tests/pull/1196

PR #1109, opened 2022-12-01
move stEIP3540 to Shanghai/stEOF

winsvega:
@holiman 
because will require to adjust geth local test runner


winsvega:
https://github.com/ethereum/tests/pull/1110

PR #1110, opened 2022-12-02
EipStateTests

winsvega:


gumb0:
To me it would make more sense to still have it in GeneralsStateTests, e.g.
`src/EIPStateTests` => `src/GeneralStateTests/EIPTests`

winsvega:
The reason it is out of state tests since it represent not a mandatory tests. But yes, can consider that too.

gumb0:
Other top-level folders corrsepond to kinds of tests - State, Blockchain, Transaction... "EIPs" is not a separate kind of tests, it's state tests, too.
(having non-mandatory forks like `Merge+3540` inside makes it possible to "skip" them if not implemented, but yeah you still would run these files if you run all state tests)

holgerd77:
> To me it would make more sense to still have it in GeneralsStateTests, e.g. `src/EIPStateTests` => `src/GeneralStateTests/EIPTests`

Yes, think I would prefer this too (writing this after my comment on our Discord where I expressed general agreement in a first-round).

winsvega:
ok moved there.
and in `/src/BlockchainTestsFiller/EIPTests`  for many tx per block tests too.

PR #1112, opened 2022-12-07
EIP-3540: Test EOF initcode cannot create legacy contracts

hugo-dc:
EOF initcode (in tx data, or CREATE/CREATE2 initcode) creating Legacy contracts are [now expected to be invalid](https://github.com/ethereum/EIPs/pull/6052), previous tests were moved to the corresponding test file testing invalid EOF code/initcode.

Tests were filled using this geth version: https://github.com/ipsilon/go-ethereum/tree/eip-3670


gumb0:
Need to move this to new folder `GeneralStateTests/stEOF/stEIP3540`

winsvega:
`GeneralStateTests/EIPTests/stEOF/`

hugo-dc:
Rebased develop, tests were moved to the correct directory.

src/GeneralStateTestsFiller/EIPTests/stEOF/stEIP3540/CreateTransactionEOF1Filler.yml
```diff
@@ -59,16 +59,6 @@ CreateTransactionEOF1:
       # deployed code section: 'PUSH1 0 STOP' data section: 'aabbccdd'
       - ':label eof1_initcode_with_data_deploying_valid_eof1_with_data :raw 0xef0001010029020002007fef000101000302000400600000aabbccdd00000000000000000000000000000060005260116000f3eeff'
 
-      # EOF initcode => legacy code
```
gumb0, line 62: We still need to have a test for this

> For a create transaction, if _initcode_ or _code_ is invalid, the contract creation results in an exceptional abort. Such a transaction is valid and may be included in a block.

It must check that transaction was valid, but nothing was deployed

gumb0, line 62: We still need to have a test for this

> For a create transaction, if _initcode_ or _code_ is invalid, the contract creation results in an exceptional abort. Such a transaction is valid and may be included in a block.

It must check that transaction was valid, but nothing was deployed

gumb0: Ah nevermind, I see we have this now in CreateTransactionInvalidEOF1

gumb0:
EOF1_Execution and EOF1_Calls filled tests changed, but I don't see a change in fillers, why is that?

gumb0:
> EOF1_Execution and EOF1_Calls filled tests changed, but I don't see a change in fillers, why is that?

Maybe it was in those missing fillers, but I think we're missing here to check that inside EOF contract CREATE/CREATE2 with legacy initcode results in a failure.

gumb0:
> EOF1_Execution and EOF1_Calls filled tests changed, but I don't see a change in fillers, why is that?

@winsvega this could have been detected by CI (test files updated without update in fillers)

winsvega:
Its vice versa. We check the filler hash. If filler is changed test must be updated.

We trust that no commit happens like that

gumb0:
You could addionally check that if test was updated, filler hash was changed.

hugo-dc:
> EOF1_Execution and EOF1_Calls filled tests changed, but I don't see a change in fillers, why is that?

This is interesting: The filler didn't change at all!

The JSON file is changed because, I'm using a different solc version or a not optimized one.

For example: In my solc version, the Yul code `{ sstore(1, 1) }` gets compiled to `PUSH1(1) PUSH1(1) SSTORE`, but in your version this would be compiled to: `PUSH1(1) DUP1 SSTORE`.

gumb0:
I see, I will delete the refilled files without fillers then.

gumb0:
> You could addionally check that if test was updated, filler hash was changed.

I guess my suggestion isn't really straightforward - because someone might want to refill the tests without changing the fillers, e.g. with new implementation or new comilers.
(I still can imagine a warning if neither of tooling nor fillers changed, but filled files did)

gumb0:
> Maybe it was in those missing fillers, but I think we're missing here to check that inside EOF contract CREATE/CREATE2 with legacy initcode results in a failure.

Actually these cases exist in `CREATE_EOF1_FromEOF` and `CREATE2_EOF1_FromEOF` and apparently they succeeded, so implementation was wrong.

We should try next with new geth implementation (and probably move these cases to `_Invalid` part)

hugo-dc:
Closing in favor of: https://github.com/ethereum/tests/pull/1127

PR #1116, opened 2022-12-13
Update docs for new blockchain test related features

qbzzt:


winsvega:
statediff from specific transaction state works only with filltests. Without it only select full blocks. 

Same with poststate.

winsvega:
vmtrace however works in both modes as t8n tool provides access to traces

qbzzt:
> statediff from specific transaction state works only with filltests. Without it only select full blocks.
> 
> Same with poststate.

Wouldn't `--poststate` *always* be at the end of a block? BTW, added that comment on `--statediff`.

winsvega:
Yes, poststate takes at the end of a block. 
But when filling you can choose a transaction. 2:1 that would be at the block 2, after transaction 0 and 1 (and block might have 2,3,4 transactions) thats because when generating I send transactions one by one and when running I only import already generated rlp

winsvega:
Still conflict on the branch?

PR #1126, opened 2022-12-28
feat(EOFlizer): Create valid and invalid EOF1 code for testing purposes

qbzzt:
`node.js EOF1-generator.js > ~/tests/src/GeneralStateTests/<stWhatever>/EOF1ValidInvalidFiller.yml`


jochem-brouwer:
You might be interested in this;

https://github.com/ethereumjs/ethereumjs-monorepo/blob/3487ed5834833e449b9481a29563c98f7700ce0b/packages/vm/test/api/EIPs/eip-4750-eof-functions.spec.ts#L66

It supports multiple functions.

jochem-brouwer:
Ah wait, maybe this is already supported in the generator.

qbzzt:
> Ah wait, maybe this is already supported in the generator.

https://github.com/ethereum/tests/blob/3c1df43a5982c566de46cc4a2b6fe6e01b293fb9/src/Templates/EOF1/EOF1-generator.js does support multiple code segments. I haven't tested a case where one segment calls another, but I should add that.

PR #1103, opened 2022-11-21
EIP-4200 Tests

hugo-dc:
Based on Tests vectors: https://github.com/ethereum/tests/issues/1093

---

#### Valid

- EOF code containing RJUMP (Positive, Negative, Zero)
- EOF code containing RJUMPI (Positive, Negative, Zero)
- EOF with RJUMPV table size 1 (Positive, Negative, Zero)
- EOF with RJUMPV table size 3
- EOF with RJUMPV table size 255 (Target 0, 100, 254, 256)
- Code with branches having the same stack height
- Jump table
- Infinite loop
- Infinite loop using RJUMPV
- CALLF branches with the same total of outputs
- CALLF inputs
- Validate input for all opcode requiring at least 1 stack items
- Containing terminating opcode at the end (RETURN, REVERT)
- Loop ending with unconditional RJUMP
- Functions ending with RETF
- Stack is not required to be empty on terminating instruction (RETURN, REVERT)
- RETF returning maximum number of outputs (127)

#### Invalid

- EOF code containing truncated RJUMP
- EOF code containing RJUMP with target outside code bounds
- EOF code containing RJUMP with target immediate
- EOF code containing truncated RJUMPI
- EOF code containing RJUMPI with target outside code bounds
- EOF code containing RJUMPI with target immediate
- EOF code containing RJUMPV with count 0 and zero immediates
- EOF code containing RJUMPV with count 0 and one immediates
- EOF code containing truncated RJUMPV
- EOF code containing RJUMPV with target outside code bounds
- EOF code containing RJUMPV with target immediate

#### Execution

Test RJUMP/RJUMPI/RJUMPV execution.

- EOF1_RJUMP_RJUMPI_Execution
- RJUMP_MaxOffset
- RJUMPI_MaxOffset

PR #1127, opened 2022-12-29
EIP-3540: Unified Spec Tests

hugo-dc:
Contains tests in https://github.com/ethereum/tests/pull/1112 , updated based on latest EOF Unified Spec


gumb0:
@winsvega Is it ok if we leave expectations with `Shanghai` fork name here?

winsvega:
Yes, but also put a comment in info comment section with url to the geth/client generated the test and its branch and what eips are activated.

winsvega:
Also checkout /src/README.md

There we put map of ids and test vector descriptions. The ids then can be found in tests (also in comments section)

gumb0:
This still needs to move cases with CREATE/CREATE2 in EOF code with leagacy initcode to Invalid files, and refill it with implementation that fails such CREATEs.

src/GeneralStateTestsFiller/EIPTests/stEOF/stEIP3540/EOF1_CallsFiller.yml
```diff
@@ -126,54 +116,44 @@ EOF1_Calls:
...
-      code: ':raw 0xEF000101002B00600060006000600073c94f5374fce5edbc8e2a8697c15331677e6ebf0b620186a0fa600055600160015500'
+      code: ':raw 0xef0001010004020001002b0300000000000006600060006000600073c94f5374fce5edbc8e2a8697c15331677e6ebf0b620186a0fa600055600160015500'
       nonce: 0
       storage: {}
     344f5374fce5edbc8e2a8697c15331677e6ebf0b:
       balance: 0
       # EOF code STATICCALLs legacy code
       # code: ':yul { sstore(0, staticcall(100000, 0xd94f5374fce5edbc8e2a8697c15331677e6ebf0b, 0, 0, 0, 1)) sstore(1, mload(0)) sstore(2, 2) stop()}'
-      code: ':raw 0xEF000101003100600160006000600073d94f5374fce5edbc8e2a8697c15331677e6ebf0b620186a0fa600055600051600155600260025500'
+      code: ':raw 0xef000101000402000100310300000000000006600160006000600073d94f5374fce5edbc8e2a8697c15331677e6ebf0b620186a0fa600055600051600155600260025500'
```
winsvega, line 140: Ori is working on a script that will convert normal yul into eof code. But it is good to have it defined explicitly here for now

gumb0: Solidity already compiles yul to eof in a branch, see https://github.com/ethereum/tests/pull/1103#issuecomment-1366789807

src/GeneralStateTestsFiller/EIPTests/stEOF/stEIP3540/EOF1_CallsFiller.yml
```diff
@@ -498,7 +413,7 @@ EOF1_Calls:
     - indexes:
        data: ':label eof_staticcall_eof'
       network:
-        - 'Merge+3540+3670'
+        - 'Shanghai'
       result:
         a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
           nonce: 1
```
winsvega, line 419: Would be nice to have a little comment for the storage values below, what they intent to indicate.

src/GeneralStateTestsFiller/EIPTests/stEOF/stEIP3540/EOF1_CallsFiller.yml
qbzzt: > Ori is working on a script that will convert normal yul into eof code. But it is good to have it defined explicitly here for now

I gave up on that idea (it was to convert bytecode, not yul). It's a lot harder than it appears, because of the way jumps changed and the way you need to track maxStackHeight.

src/GeneralStateTestsFiller/EIPTests/stEOF/stEIP3540/EOF1_CallsFiller.yml
gumb0: Where exacty would you like this? In every expectation in every filler? If we add a a comment in every expectation, it will be very repetitive. It's basically everywhere:
- slot 0 - result of CALL/CREATE/CREATE2
- slot 1 - always 1 when not OOG

PR #1137, opened 2023-01-08
fix basefee in wt tests

winsvega:


jochem-brouwer:
I am confused here, tests seem to include withdrawals with 0 wei as value, this is invalid per https://eips.ethereum.org/EIPS/eip-4895

(Also checked that the block is valid since latest hash points to that block)

PR #1138, opened 2023-01-09
feat(Templates/DiffPlaces): Update to Shanghai

qbzzt:


winsvega:
I am afraid this needs an update with weitogwei update on Shanghai spec

src/Templates/DiffPlaces/templateGen.js
```diff
@@ -679,7 +679,7 @@ ${testName}:
         value: !!int -1
 
       network:
-        - '>=London'
+        - '>=Merge'
```
winsvega, line 682: why no London?

winsvega:
there are more tests refilled than src changed. please refill only those tests that are affected by changed srcs 
and commit in indderent commits. src changes - 1 commit, filltests - 2 commit.

winsvega:
also to keep it simple. 1 PR for badOpcodes, 2 PR for diffplaces

src/Templates/DiffPlaces/templateGen.js
qbzzt: Just because I don't see the point of checking London, just as I don't see the point of checking Byzantium. I can undo this this.

PR #1142, opened 2023-01-12
feat(validInvalidOpcode): New test (with generator)

qbzzt:
Check if opcodes are valid or not by calling a contract with them. All in a single state test.


winsvega:
this goes to the next release

src/GeneralStateTestsFiller/stBadOpcode/validInvalidOpcodeFiller.yml
```diff
@@ -0,0 +1,6008 @@
...
+                                                  11,12,13,14,15,16,17,18,19,20)
+
+             // If we get here, return one byte. That would be the sign of success
+             return(0,1)
+         }
+
+
+
+
+    0x7E57C0DE00000000000000000000000000000001:
```
winsvega, line 83: is there a reason each time you call a different address? can't the opcode to call be in input argument?

qbzzt:
Replaced by https://github.com/ethereum/execution-spec-tests/pull/54

PR #1143, opened 2023-01-14
feat(CreateEOF1): Update to solc 0.8.17

qbzzt:
Also, add spaces and dashes to EOF1 code to make it clearer


winsvega (changes requested):
this need regeneration with weitogwei branch on also latest retesteth update. 
the withdrawalsRoot hash reported is incorrect. 
rebase the branch first because I think I already fixed this tests

qbzzt:
> this need regeneration with weitogwei branch on also latest retesteth update. the withdrawalsRoot hash reported is incorrect. rebase the branch first because I think I already fixed this tests

I think we can close it. I replaced EOF1ValidInvalid with an EOF test, and if you already fixed the CREATE test, there's nothing else to do.

PR #1144, opened 2023-01-15
Tests for arithmetic opcodes

qbzzt:
9156 tests. The code is ~400k bytes, but it bypasses the tests by specifying it as `code:` rather than CREATE[2] ing it, so it works, at least on geth.


chfast (commented):
image

Why do you think generating huge tests like this is any good idea? Even GitHub refused to show it.

src/Templates/EachOpcode/Yul2Filler.js
```diff
@@ -0,0 +1,86 @@
...
+      code: 0x
+      nonce: 1
+      storage: {}
+
+
+  transaction:
+    data:
+    - 0x
+    gasLimit:
+    - 80000000
```
chfast, line 64: Stop using arbitrary large gas limits.

qbzzt:
> Why do you think generating huge tests like this is any good idea? Even GitHub refused to show it.

Because in this case the Filler is the compiled version. The source code is actually https://github.com/ethereum/tests/blob/bdfd202b89d8b3845fad6a3e4af91d89d4c7897a/src/Templates/EachOpcode/Arith.ts, which is less than 400 lines.

> Stop using arbitrary large gas limits.

Why? It makes it easier to run tests without worrying about the gas limit, when the gas limit is not part of what we test.

chfast:
> Why? It makes it easier to run tests without worrying about the gas limit, when the gas limit is not part of what we test.

Not everyone can run a test with 80M gas limit.

chfast:
> Because in this case the Filler is the compiled version. The source code is actually https://github.com/ethereum/tests/blob/bdfd202b89d8b3845fad6a3e4af91d89d4c7897a/src/Templates/EachOpcode/Arith.ts, which is less than 400 lines.

If you need additional language to generate a filler to generate a test you better contribute it to https://github.com/ethereum/execution-spec-tests.

qbzzt:
> If you need additional language to generate a filler to generate a test you better contribute it to https://github.com/ethereum/execution-spec-tests.

How about I get started with writing docs for how to write tests for that system? It looks similar to retesteth, but there's enough difference to justify more docs.

winsvega:
yes makes sense to look into python specs if filler design is too complex. 
but also Ori, I would love to have the docs on pyspecs on how to regenerate those tests with given t8n tool, and on new forks, and how to analyze their expect section. 
also pyspecs need update on generated test format as it generates incomplete accounts and the fields format is incoherent with the current one we use in tests

qbzzt:
> @winsvega: also pyspecs need update on generated test format as it generates incomplete accounts

Can you give me the name of a test where it does that and I'll try to fix it?

winsvega:
Its fixed now

qbzzt:
> I would love to have the docs on pyspecs on how to regenerate those tests with given t8n tool, and on new forks,

I added that to https://github.com/ethereum/execution-spec-tests/pull/51 - you just update the `evm` in the path to the new one,

winsvega:
Thats not very convenient when I have multiple evms on the same machine

I would need to run mv commands. Would be much nicer if I could provide the path to python script

qbzzt:
You're right. I will make a script to improve that.

winsvega:
Are you redoing this PR with pyspecs too?

qbzzt:
Yes. I think we can close it.

qbzzt:
>

PR #1149, opened 2023-01-19
feat(EIPs): Added a README.md

qbzzt:


winsvega (changes requested):
also in BlockchainTests/EIPTests
and not only in src but alson in filled folders

winsvega (approved):

qbzzt:
Rebased.

PR #1151, opened 2023-01-23
feat(EOF1ValidInvalid): Replaced the generator with an EOF test

qbzzt:


winsvega:
ah. need rebase and resolv conflict now. 
its good to have less generators. the generators mean we need better test src editor. maybe python help.

qbzzt:
Was easiest to close this PR and open a separate one.

PR #1159, opened 2023-02-04
feat(EOF validInvalid): Add to repository

qbzzt:
Replacing PR #1151


winsvega:
what about the old generator? is it removed?

PR #1160, opened 2023-02-09
Update tests to Shanghai (partially)

rodiazet:
Most of the test are updated to Shanghai. Still a couple of errors listed in attached doc.
Testing on evmone master branch

https://hackmd.io/@etgDMnCFQgS7tnZ9yjkOPw/SylTKkypj


winsvega:
this is a matter of revisioning each test that failed like this. because need to be sure that the test didn't fail and still works and affected by shanghai in a good way. (gas change, eip interaction)

for this reason there are warning when expect section uses coinbase balance or origin balance checks.

rodiazet:
> this is a matter of revisioning each test that failed like this. because need to be sure that the test didn't fail and still works and affected by shanghai in a good way. (gas change, eip interaction)
> 
> for this reason there are warning when expect section uses coinbase balance or origin balance checks.

Sure. Just wanted to know which tests I didn't update do Shanghai because it probably takes minutes to someone who is more familiar with the generated tests. I hope it's beneficial in any way.

winsvega:
is latest geth master shanghai ready?

winsvega:
in progress. updating all tests to Shanghai. I've checked all of them seems to be no breaking changes. Ori must fix his vectors now.

PR #1161, opened 2023-02-11
Correct source path in EIP1559 tests

jkilpatr:
All transaction tests in the `ttEIP1559` directory have an invalid source path with `ttValue` instead of `ttEIP1559`. This patch corrects the source path for those files.


winsvega:
the retesteth flag to update it if no changes are in the tests is `--forceupdate`

PR #1162, opened 2023-02-12
feat(doubleSelfdestructTest[2]): Change to YUL

qbzzt:


winsvega (changes requested):
can you add /src/README test vectors descriptions that are covered in this test?

qbzzt:
> can you add /src/README test vectors descriptions that are covered in this test?

Added in PR #1169 (I'm afraid to change the same files in two PRs at the same time)

winsvega:
need to make it a single commit. src files only and put it to shanghai branch.

PR #1163, opened 2023-02-17
EIP-5450: State Tests

hugo-dc:
**EOF Code Validation**

- EOF1V5450_0001 (Valid) Code with branches having the same stack height
- EOF1V5450_0002 (Valid) Jump table
- EOF1V5450_0003 (Valid) Infinite loop
- EOF1V5450_0004 (Valid) Infinite loop using RJUMPV
- EOF1V5450_0005 (Valid) CALLF branches with the same total of outputs
- EOF1V5450_0006 (Valid) CALLF inputs
- EOF1V5450_0007 (Valid) Validate input for all instructions requiring stack input
- EOF1V5450_0092 (Valid) Containing terminating opcode RETURN at the end
- EOF1V5450_0093 (Valid) Containing terminating opcode REVERT at the end
- EOF1V5450_0094 (Valid) Loop ending with unconditional RJUMP (a)
- EOF1V5450_0095 (Valid) Loop ending with unconditional RJUMP (b)
- EOF1V5450_0096 (Valid) Functions ending with RETF
- EOF1V5450_0097 (Valid) Stack is not required to be empty on terminating instruction RETURN
- EOF1V5450_0098 (Valid) Stack is not required to be empty on terminating instruction REVERT
- EOF1V5450_0099 (Valid) RETF returning maximum number of outputs (127)
- EOF1I5450_0001 (Invalid) Stack height mismatch for different paths (a)
- EOF1I5450_0002 (Invalid) Stack height mismatch for different paths (b)
- EOF1I5450_0003 (Invalid) Calls returning different number of outputs
- EOF1I5450_0004 (Invalid) Pushing loop
- EOF1I5450_0005 (Invalid) Popping loop
- EOF1I5450_0006 (Invalid) Jump table with different stack heights
- EOF1I5450_0007 (Invalid) Stack underflow for all instructions requiring stack input
- EOF1I5450_0092 (Invalid) Calling function without enough stack items: Function 0 calls Function 1 without enough parameters
- EOF1I5450_0093 (Invalid) Calling function without enough stack items: Function 1 calls Function 2 with enough parameters
- EOF1I5450_0094 (Invalid) Calling function without enough stack items: Function 0 calls Function 1 without enought parameters, Function 1 calls Function 2 without enough parameers
- EOF1I5450_0095 (Invalid) Stack Overflow: Function pushing more than 1024 items to the stack
- EOF1I5450_0096 (Invalid) Stack Overflow: Function 1 when called by Function 0 pushes more than 1024 items to the stack
- EOF1I5450_0097 (Invalid) Function ending with non-terminating instruction (a)
- EOF1I5450_0098 (Invalid) Function ending with non-terminating instruction (b)
- EOF1I5450_0099 (Invalid) Function ending with non-terminating instruction (c)
- EOF1I5450_0100 (Invalid) Function containing unreachable code after RETURN
- EOF1I5450_0101 (Invalid) Function containing unreachable code after REVERT
- EOF1I5450_0102 (Invalid) Unreachable code after RJUMP
- EOF1I5450_0103 (Invalid) Unreachable code after infinite loop

**CALLF Execution**

- Valid functions reaching max stack size (1024)


winsvega:
please look at EOFTests its a whole new suite to develop the tests for EOF code. 
basically its just a state test format but simplified to focus only on EOF code deployment

hugo-dc:
@winsvega I have moved the validation tests to a [new file](https://github.com/ethereum/tests/blob/0b6245e593c4722c00b7b800755edb08268a37d0/src/EOFTestsFiller/EIP5450/validInvalidFiller.yml) using the validation format.

winsvega:
Just a question. Do we need eof code validation exception checks?
So that you can specify that this eof is invalid because of EOF header field has incorrect byte to verify that client t8n return expected error code?

winsvega:
Btw. Put the test vectors both in the test and /src/Readme.md

hugo-dc:
> Just a question. Do we need eof code validation exception checks? So that you can specify that this eof is invalid because of EOF header field has incorrect byte to verify that client t8n return expected error code?

Yes, that would be very helpful to really validate the invalid test are failing for the expected reason. Had an issue before when the test was failing but it was due to a  malformed EOF code instead of a different reason I was testing.

src/EOFTestsFiller/EIP5450/validInvalidFiller.yml
```diff
@@ -3,9262 +3,6166 @@ validInvalid:
...
+            80 # DUP1 
+            a2 # LOG2 
+            00 # STOP 
+               # Data section (empty)
+      expectException:
+        "Shanghai" : "Invalid Code"
+
+    # Data: 82
+    # EOF1V5450_0083 (Valid) Validate input for LOG3 opcode
+    # Code:
```
winsvega: `>=Cancun` to autoenable exception on all future forks

winsvega (changes requested):
needs rebase on develop.

winsvega (approved):

PR #1169, opened 2023-02-26
Multi suicide

qbzzt:
call happens within one transaction to contractA, that has different scenarios around suicide
contractA has balance=3

SUC000  - contractA is called to perform suicide to dest1 (3 must immidiately(check that it is immidiately and not by the end of tx) go to dest1)
contractA is called again in the same tx, but this time

SUC001 - does not suicide, just receive more funds via call. check the balance of contractA (must be 0, the second transfer funds erased)
SUC002 - does another suicide to dest1 (so dest1 now have 3 + x) 
SUC003 - does another suicide to dest2 (so both dest1 and dest2 has some balance)
SUC004 - tries to make value transfer but fails because initial suicide 0ed the balance (answering this (check that it is immidiately and not by the end of tx))
SUC005 - receives more funds and try to transder it (same as SUC004, but this time work because it received funds via 2nd call)


src/GeneralStateTestsFiller/stSystemOperationsTest/multiSelfdestructFiller.yml
```diff
@@ -0,0 +1,196 @@
...
+             // SUC000
+             mstore8(0, 0xFF)
+             mstore8(1, 0x10)
+             mstore8(2, 0x00)
+             delme := call(gas(), 0xdead, 0, 0,3, 0,0)
+             sstore(0x00, balance(0x1000))
+             sstore(0x10, delme)
+
+             // Don't do anything, just receive a wei
+             // SUC001
```
winsvega: No, each test vector assumes the transaction finalization happens.
And then the scenario repeats again.

Otherwise they can entangle and affect each other 

Anyway it is better to put eachvvector in a separate transaction anyway

winsvega (changes requested):
Each vector assumes a transaction finalization happens. 1 tx - one vector.

Suc000 is the base. It happens in the beginning of each following vector.

So  1. Call regular suicide. 2. Call again selfdesteuct contract in the same tx. But different call. 3. Tx finishes. 4 . Check the post state.

Then go to the next vector that do the same. But different scenario in step 2.

src/GeneralStateTestsFiller/stSystemOperationsTest/multiSelfdestructFiller.yml
```diff
@@ -0,0 +1,278 @@
...
+        # Now (post tx) 0xDEAD no longer exists, so balance=0
+        000000000000000000000000000000000000dead:
+          shouldnotexist: 1
+
+
+
+    # SUC002
+    - indexes:
+        data: !!int 1
+        gas:  !!int -1
```
winsvega, line 185: gas and value can be omited if it is -1

winsvega (approved):
looks very good now. @marioevz  any more ideas for suicide double call ?
maybe more scenarios with revert

GeneralStateTests/stSystemOperationsTest/multiSelfdestruct.json
```diff
@@ -0,0 +1,231 @@
+{
+    "multiSelfdestruct" : {
+        "_info" : {
+            "comment" : "Implements: SUC000, SUC001, SUC002, SUC003, SUC004, SUC005\n",
```
winsvega, line 4: is this described in /src/README.md?

qbzzt: No, I thought you did - I'll add them.

PR #1176, opened 2023-03-06
feat(diffPlaces tests): Update tests to include Shanghai

qbzzt:
The final part of #1138


qbzzt:
@winsvega , no src because no src changed. It's the same src, just with the filled including Shanghai. I wouldn't have done it, but I saw that #1138 is still on the list despite #1166 being already merged - so I assumed there's some kind of problem.

winsvega:
no just the diffPlaces commit

winsvega:
will refill after

PR #1177, opened 2023-03-07
Enable ecrecover tests on all forks

axic:
The precompile has not changed since genesis, but many tests were only enabled starting a later fork (like London).

Note that some test can't be enabled before Byzantium because they use `staticcall`.

cc @gumb0 @chfast


winsvega:
What about LegacyTests?
there must be a version of this test from frontier there

axic:
> What about LegacyTests?
> there must be a version of this test from frontier there

I am not sure, but this worked well for our purposes. Any strong reason not to merge this?

winsvega:
need to remove filled versions. rebase on the develop and fill again

winsvega:
no reason no, everything for the devs. however if you need tests for all the networks like this there is a legacy tests subrepo there we have all the history vectors.\

https://github.com/ethereum/tests/pull/1190

PR #1178, opened 2023-03-07
Lower gas limit of underflow stack test 80M → 8M

axic:
These test don't actually require 80M gas to succeed.


src/GeneralStateTestsFiller/stStackTests/underflowTestFiller.yml
```diff
@@ -4529,7 +4530,7 @@ underflowTest:
         gas:  !!int -1
         value: !!int -1
       network:
-        - '>=Istanbul'
+        - 'Berlin'
```
winsvega, line 4533: why only Berlin?

winsvega:
https://github.com/ethereum/tests/pull/1191

PR #1185, opened 2023-03-10
feat(refundReset): Fix the gas costs, all the way up to Shanghai

qbzzt:


winsvega:
is there a way to make it yul independent?

PR #1192, opened 2023-03-13
move EIPTests into separate suite -t EIPTests

winsvega:
both blockchain and state versions


winsvega:
any objections? @hugo-dc

PR #1206, opened 2023-04-05
Update nonce behaviour tests. Move EOF ori test to Cancun

rodiazet:
-Change nonce behaviour to "always bump" according to https://github.com/ethereum/EIPs/blob/master/EIPS/eip-3540.md
-Move ori EOF tests to `>=Cancun` as EOF was moved to Cancun or later.
-Regenerate state and blockchain tests using evmone.


qbzzt (approved):

winsvega:
need rebase

PR #1216, opened 2023-05-09
feat(docs): Add that `!!int -1` is the default in the indexes

qbzzt:


docs/state-transition-tutorial.rst
```diff
@@ -522,15 +522,40 @@ Multiple Tests, Same Result
...
 
-::
+  .. list-table::
+   :header-rows: 0
+
+   * -
+       ::
+
+         expect:
+            - indexes:
```
winsvega, line 543: and if you remove - indexes it is gona be all set to default -1

PR #1219, opened 2023-05-17
EOF: Forbid DELEGATECALL EOF → legacy

chfast:
According to EIP-3540 updated by
https://github.com/ethereum/EIPs/pull/6359
a DELEGATECALL from EOF to legacy code is disallowed.

This updates expectations in the single test case of DELEGATECALL EOF → legacy.


winsvega:
need rebase on develop

PR #1224, opened 2023-05-24
Cancun tests [Example]

winsvega:
example of Cancun tests and EIPTests with Cancun forkname


src/EIPTestsFiller/StateTests/stEIP5656-MCOPY/MCOPY_copy_costFiller.yml
```diff
@@ -252,6 +252,7 @@ MCOPY_copy_cost:
           - :label src32_size31
           - :label src32_size32
           - :label src32_size33
+          - :label src0_size44769
```
chfast, line 255: This is because the yul code is now compiled with `PUSH0` (1 gas cheaper than `PUSH1`). I will add the destroyed test case back.

winsvega: Then push0 must not be compiled for legacy networks also

PR #1228, opened 2023-06-19
Re-assign opcodes proposed for Cancun (+EOF)

chfast:
Applies cleanup to opcodes ordering specified in https://github.com/ethereum/execution-specs/pull/764.


winsvega:
`[src/Templates/EOF1/EOF1-generator.js](https://github.com/ethereum/tests/pull/1228/files#diff-a698585921b09957be089db27093403d50c4e4a15d3a89f839b9c08e6271ce83)`
to parametrize the tests better we now use python tests and to avoid other generator scripts. 

https://github.com/ethereum/execution-spec-tests

retesteth remain supported for simple Filler.json Filler.yml  test fillers

chfast:
I only updated existing tests. You can remove the script I'm not the author.

qbzzt:
Do you want me to convert the template tests to Python? IIRC, I wrote most of them.

winsvega:
if the test works it is fine for now, but eventually parametrized tests are better to be moved to python. idk if its a good idea to rewrite it at the moment since might be double work. 

some fillers are not generated

Error: Test /home/travis/build/ethereum/tests/EOFTests/EIP5450/validInvalid.json is outdated. Filler hash is different! ('aa65' != '011b')

chfast:
It was easy to update the script this time. I think this is fine at this point (the modification is because of the spec change).

shemnon:
Should these be targeting another fork other than Cancun?  Or use the +EIP notation?  (but which EIP?)

gumb0:
Opcodes change is now included in #1238, this can be closed.

PR #1229, opened 2023-06-19
Add MCOPY tests

chfast:
The MCOPY test file focus on properly copying memory bytes.
The MCOPY_copy_cost test file focus on properly calculating the MCOPY instruction copy gas cost.
The MCOPY_memory_expansion_cost test file focus on properly calculating potential memory expansion cost in the MCOPY instruction.


chfast:
This is ready from my side.

shemnon:
Besu passes these tests.

jochem-brouwer:
EthereumJS passes these tests.

However, there is a case missing (I think). 

We had a bug in the gas calculation where the src > dst, and we are going to read out-of-bounds of current memory. The memory should thus be expanded to `src + length`. We do not do this, but this is not tested. (I believe this should be the case, right?)

jochem-brouwer:
So, what should be done in that test, is perform the exact same MCOPY again, to see if memory is properly expanded or not (now memory expansion costs are 0). In our case, memory expansion costs will not be zero.

jochem-brouwer:
Actually, never mind, we do extend, but if uncovered it should maybe be added.

MarekM25:
Nethermind passes these tests :)

chfast:
> So, what should be done in that test, is perform the exact same MCOPY again, to see if memory is properly expanded or not (now memory expansion costs are 0). In our case, memory expansion costs will not be zero.

All tests in `MCOPY.json` and `MCOPY_copy_cost.json` have memory already expanded so the memory expansion cost in `MCOPY` should be 0.

> However, there is a case missing (I think).
> 
> We had a bug in the gas calculation where the src > dst, and we are going to read out-of-bounds of current memory. The memory should thus be expanded to `src + length`. We do not do this, but this is not tested. (I believe this should be the case, right?)

In `MCOPY_memory_expansion_cost.json` the following test cases have `src > dst`:

- `dst0_src31_size706`
- `dst31_src62_size706`
- `dst0_src64_size1344`
- `dst33_src64_size1344`
- `dst1_src33_size1344`
- `huge_src0_size1`
- `huge_dst0_size_n256`

However, I'm not sure how effective these are for finding bugs out-of-bound or uninitialized-memory bugs. We can add more cases if you have precise values of `MCOPY` arguments.

jochem-brouwer:
I just saw that MSIZE is stored, but the idea would be to, after storing MSIZE, perform the exact same MCOPY operation again. This would detect a case that internally the MSIZE is set correctly, but somehow when resizing it, it takes the wrong MSIZE (for instance if it did not properly expand the memory with extra zeros). However upon writing this, this seems like a super exotic bug. But you could edit the test code to MCOPY again, to ensure that the gas costs are also correct if you call it again (hence now expansion cost is zero). I'll add a comment, but I doubt this will find bugs.

src/EIPTestsFiller/StateTests/stEIP5656-MCOPY/MCOPY_memory_expansion_costFiller.yml
```diff
@@ -0,0 +1,221 @@
...
+          function mcopy(dst, src, size) { verbatim_3i_0o(hex"5e", dst, src, size) }
+        
+          // Take most of the SSTORE cost before MCOPY. 
+          sstore(0, 1)
+        
+          // MCOPY using parameters from CALLDATA.
+          mcopy(calldataload(0), calldataload(32), calldataload(64))
+        
+          // Put MSIZE in storage.
+          sstore(0, msize())
```
jochem-brouwer, line 30: After this line, do

`mcopy(calldataload(0), calldataload(32), calldataload(64))`

again

chfast: I will add one more test suite for it.

jochem-brouwer: Ok great :smile:

chfast: Added in https://github.com/ethereum/tests/pull/1234.

PR #1231, opened 2023-06-21
Replace invalid JSON strings

shekhirin:
`Nimbus-t8n 0.1.2\n[0m` is an invalid json string

image

I noticed it when Reth state tests started failing: https://github.com/paradigmxyz/reth/actions/runs/5337139311/jobs/9672776651?pr=3307#step:7:105


winsvega:
ah no, I need to fix parsing nimbus --version command so it wont happen again.

PR #1234, opened 2023-06-25
Add more MCOPY tests

chfast:


jochem-brouwer:
Thanks for these tests, EthereumJS passes as well :+1: :smile:

chfast:
For everyone trying these tests, please take a look at detailed code coverage of your implementation if possible. If anything is not covered it means a test case is missing.

shemnon:
Besu passes. I'm not seeing any uncovered cases in Besu's code.

PR #1237, opened 2023-06-27
refill transient storage tests with updated specs on Cancun

winsvega:


chfast:
All passed by evmone.

Demuirgos:
do these tests use the updated opcodes as well ?

shemnon:
Yes, the opcode change is what triggered this refill.

Demuirgos:
> Yes, the opcode change is what triggered this refill.

Ah okey, are all the tests refilled? because 21 of them for some reason invoke a precompile (with address 0x0a), or is that by design?

shemnon:
Yea, the tests need to be updated, and then refilled.  `0xA` is now a precompile address so these tests should fail spectacularly once we get more clients implementing it.

winsvega:
idk what was the reason to use 0x0A as contract address in this test. I think we can change it to a different address. but lets see if it fails, the precompiles should not be redefined in pre section

holiman:
>  I think we can change it to a different address. but lets see if it fails, the precompiles should not be redefined in pre section

Dunno about other clients, but for geth, it doesn't matter if a precompile "has code", we will execute the _precompile_, not the code associated with the account. So it's not possible to redefine them in the `pre`.

shemnon:
That's the problem Besu ran into as well.  We check registered precompiles before checking for code, so when the new precompile landed tests broke.

PR #1267, opened 2023-08-13
stRefundTest :yul fix

qbzzt:
Also a couple of gas cost changes, probably due to optimization


src/GeneralStateTestsFiller/stRefundTest/refundMaxFiller.yml
```diff
@@ -76,12 +80,14 @@ refundMax:
         value: !!int -1
 
       network:
-        - '>=London'
```
winsvega, line 79: the test is now missing expect section for future networks

Issue #1286, opened 2023-08-24
Add some state tests with transaction from non-existing account


chfast:
This is untested territory in state tests. A transaction's sender may be non-existing account. Surely, such transaction will be invalid if base fee is non-zero. But we still want to make sure such transaction is properly validated.

winsvega:
hm, we have bc tests for this. since t8n is used it will test the same logic. 
you need such case for a custom test runner?

chfast:
No, I need this for my statetest runner.

PR #1290, opened 2023-09-14
Try to send transactions from a non-existent account

qbzzt:
@chfast , is this what you want? This PR checks txn types 0, 1, and 2, both a normal transaction and a contract creation transaction, and with different gas prices, data sizes, and values sent.

Closing: #1286


chfast (commented):
Yes, this is more than enough. Thanks!

PR #1204, opened 2023-04-03
EOFv2 - CREATE3 tests

gumb0:

PR #1307, opened 2023-10-17
EOF - simple CREATE3 test

gumb0:
This is more up to date with the spec than https://github.com/ethereum/tests/pull/1204


BertsLifee (approved):

chfast:
Outdated (I don't even remember what `CREATE3` was).

PR #1311, opened 2023-10-18
refill all tests with cancun

winsvega:


shemnon:
Besu agrees with these tests, after two minor fixes to Besu.

https://github.com/hyperledger/besu/pull/6054

jangko:
Nimbus also agrees with these tests after a stupid bug fix. 😄 

status-im/nimbus-eth1#1822

PR #1326, opened 2023-12-19
pytest test

winsvega:


src/GeneralStateTestsFiller/Python/doubleSuicide.py
```diff
@@ -0,0 +1,208 @@
...
+
+@pytest.mark.valid_from("Berlin")
+@pytest.mark.parametrize("firstSuicide", ["call", "callcode", "delegatecall"])
+@pytest.mark.parametrize("secondSuicide", ["call", "callcode", "delegatecall"])
+def test_doubleSuicide(
+    env: Environment,
+    yul: YulCompiler,
+    fork: Fork,
+    firstSuicide: Any,
+    secondSuicide: Any,
```
marioevz:
```suggestion
    firstSuicide: Op,
    secondSuicide: Op,
```

src/GeneralStateTestsFiller/Python/doubleSuicide.py
```diff
@@ -0,0 +1,208 @@
...
+        difficulty=0x020000,
+        gas_limit=71794957647893862,
+        number=1,
+        timestamp=1000,
+    )
+
+
+
+@pytest.mark.valid_from("Berlin")
+@pytest.mark.parametrize("firstSuicide", ["call", "callcode", "delegatecall"])
```
marioevz: We normally use underscore variable names in Python, so `first_suicide` would be more appropriate.

marioevz: We could also parametrize to do:
```suggestion
@pytest.mark.parametrize("first_suicide", [Op.CALL, Op.CALLCODE, Op.DELEGATECALL])
```

src/GeneralStateTestsFiller/Python/doubleSuicide.py
```diff
@@ -0,0 +1,208 @@
...
+        R reverts (including the effects of the second selfdestruct).
+        It is expected the S is selfdestructed after the transaction.
+    """
+
+    addressS = "0x1000000000000000000000000000000000000001"
+    addressR = "0x1000000000000000000000000000000000000002"
+
+    callS = ""
+    if firstSuicide == "callcode" or firstSuicide == "call":
+        callS = f"""{firstSuicide}(gas(), {addressS}, 0, 0, 0, 0, 0)"""
```
marioevz: Using Opcodes we could rewrite to:
```python
call_s_args = [Op.GAS, Op.PUSH20(address_s)]
if first_suicide in [Op.CALLCODE, Op.CALL]:
  call_s_args += [100]
call_s_args += [0] * 4
```
Then when using the Opcode you can do:
```python
Op.SSTORE(0, first_suicide(*call_s_args))
```

src/GeneralStateTestsFiller/Python/doubleSuicide.py
```diff
@@ -0,0 +1,208 @@
...
+            code=yul(
+                """
+                    {
+                        let success := """f"""{callS}""""""
+                        sstore(1, success)
+
+                        let contractR := """f"""{addressR}""""""
+                        success := call(gas(), contractR, 0, 0, 0, 0, 0)
+                        sstore(2, success)
+                        if iszero(success)
```
marioevz: The `if` is tricky in python Opcodes, but I think it's unnecessary?

src/GeneralStateTestsFiller/Python/doubleSuicide.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+    if firstSuicide == "callcode" or firstSuicide == "delegatecall":
+        if is_fork(fork, Cancun):
+            # On Cancun even callcode/delegatecall does not remove the account, so the value remains
+            post["0x095e7baea6a6c7c4c2dfeb977efac326af552d87"] = Account(
+                storage={
+                    "0x01" : "0x01",    # First call to contract S->suicide success
+                    "0x02" : "0x00",    # Second call to contract S->suicide reverted
+                    "0x03" : "15",      # Reverted value to check that revert really worked
```
marioevz:
```suggestion
                    0x01 : 0x01,    # First call to contract S->suicide success
                    0x02 : 0x00,    # Second call to contract S->suicide reverted
                    0x03 : 15,      # Reverted value to check that revert really worked
```

src/GeneralStateTestsFiller/Python/doubleSuicide.py
```diff
@@ -0,0 +1,208 @@
...
+        # Second caller unchanged as call gets reverted
+        post[addressR] = Account(
+            balance=5000000000000000000,
+            storage={}
+        )
+        if is_fork(fork, Cancun):
+            # On Cancun suicide does not remove the account, just sends the balance
+            post[addressS] = Account(
+                balance=0,
+                code="0x6103e8ff",
```
marioevz, line 159: If we omit the code in the `post` is ok, it just won't be compared, but if we use python opcodes you can copy the same value we use in the `pre`.

marioevz (commented):
Some comments :+1:

src/GeneralStateTestsFiller/Python/doubleSuicide.py
winsvega: Here I think make sense to check only string

winsvega:
https://github.com/ethereum/execution-spec-tests/pull/372

PR #1329, opened 2024-01-05
EIP-7523: Remove empty account from post-merge tests

petertdavies:
This PR removes every empty account from the prestate of post merge tests in order to comply with [EIP-7523](https://eips.ethereum.org/EIPS/eip-7523). It should not be merged until that EIP leaves Draft status. I have not included filled tests.

* Tests about empty accounts have been marked `["Istanbul", "Berlin", "London"]`.
* Tests that contain empty accounts only incidentally have had them removed.
* Some test files have had a post merge variant added that removes the empty accounts.
* Shanghai tests about empty accounts and withdrawals have been deleted.

Some obscure empty account edgecases (e.g. what if coinbase is an empty account) have only been ever tested by accident in unrelated tests. I don't seem any value in preserving these tests. If there is a desire to preserve them, new tests should be written.


src/BlockchainTestsFiller/InvalidBlocks/bcStateTests/ZeroValue_TransactionCALL_ToEmpty_OOGRevertFiller.json
```diff
@@ -23,7 +23,7 @@
 	    "sealEngine" : "NoProof",
         "expect" : [
             {
-                "network" : [">=Istanbul"],
+                "network" : ["Istanbul", "Berlin", "London"],
```
winsvega, line 26: why remove tests for nets after London?

winsvega:
if it is only after merge, then the tests must be split. empty account cases must be covered before merge, after merge vectors change into not having empty accounts

winsvega:
we really reduce test coverage by doing that in my opinion. even if we fetch the state not to have empty accounts, L2 remain vulnerable to this cases I think.

src/BlockchainTestsFiller/InvalidBlocks/bcStateTests/ZeroValue_TransactionCALL_ToEmpty_OOGRevertFiller.json
petertdavies: This test is specifically test empty account behaviour. Removing the empty account turns it into a different test, which is already implemented as `ZeroValue_TransactionCALL_OOGRevertFiller`.

petertdavies:
> we really reduce test coverage by doing that in my opinion. even if we fetch the state not to have empty accounts, L2 remain vulnerable to this cases I think.

I am confident that empty accounts do not exist in the current state of any production EVM chain anywhere.

The creation of new empty accounts became impossible in the Spurious Dragon hardfork which forked on Mainnet in November 2016. This is the same hard fork that introduced chain ids. It is unlikely that anyone would have launched an EVM after this without including Spurious Dragon since they would have wanted to use chain ids.

The only chains that are anywhere near old enough to not have launched with Spurious Dragon enabled are Ethereum, Ethereum Classic and some Ethereum Classic testnets. Ethereum is dealt with by EIP-7523 and Ethereum Classic is not going to merge so the point is moot. In addition, a transaction was sent on Ethereum Classic to clear all empty accounts.

chfast:
I'm very much in favor of this change. I can help reviewing the PR if that's needed.

PR #1339, opened 2024-02-14
EOF - validation tests

gumb0:
These tests are generated by evmone from unit tests written in C++, so they don't have a regular source / filler files.

The source of these tests can be found at:
https://github.com/ethereum/evmone/blob/eof-create4/test/unittests/eof_validation_test.cpp
https://github.com/ethereum/evmone/blob/eof-create4/test/unittests/eof_validation_stack_test.cpp
Look for the test with the same name as json test file there.


gumb0:
Regenerated the tests from `eof-create4` branch, they include now some container / subcontainer validation tests and should account for new creation instructions.

rakita (approved):
I have run these tests inside revm and they are good to go.

Would advice to merge it, so we can start cleaning up other PRs

EOFTests/eof_validation/EOF1_truncated_section.json
```diff
@@ -0,0 +1,49 @@
...
+      "EOF1_truncated_section_2": {
+        "code": "0xef000101000402000100020400000000800000fe",
+        "results": {
+          "Prague": {
+            "exception": "EOF_InvalidSectionBodiesSize",
+            "result": false
+          }
+        }
+      },
+      "EOF1_truncated_section_3": {
```
shemnon, line 31: Besu tags this one as failed `Incomplete data section`

https://github.com/ethereum/EIPs/blob/8f43e4cb0fd2fd4385d680e43f7b28e6a9e142ad/EIPS/eip-3540.md#container-validation-rules. - rule 4 state that the size _must_ match.  

I am presuming all of these are deployed or deployable, so the data section limitations in https://github.com/ipsilon/eof/blob/main/spec/eof.md#data-section-lifecycle do not apply.

EOFTests/eof_validation/EOF1_truncated_section.json
```diff
@@ -0,0 +1,49 @@
...
+        "code": "0xef000101000402000100010400020000800000fe",
+        "results": {
+          "Prague": {
+            "result": true
+          }
+        }
+      },
+      "EOF1_truncated_section_4": {
```
shemnon, line 39: Besu tags this one as failed `Incomplete data section`

https://github.com/ethereum/EIPs/blob/8f43e4cb0fd2fd4385d680e43f7b28e6a9e142ad/EIPS/eip-3540.md#container-validation-rules. - rule 4 state that the size _must_ match.  

I am presuming all of these are deployed or deployable, so the data section limitations in https://github.com/ipsilon/eof/blob/main/spec/eof.md#data-section-lifecycle do not apply.

EOFTests/eof_validation/EOF1_eofcreate_invalid.json
```diff
@@ -0,0 +1,51 @@
...
+            "exception": "EOF_InvalidContainerSectionIndex",
+            "result": false
+          }
+        }
+      },
+      "EOF1_eofcreate_invalid_4": {
+        "code": "0xef0001010004020001000c03000100160400000000800004600060ff60006000ec005000ef000101000402000100010400030000800000feaabb",
+        "results": {
+          "Prague": {
+            "exception": "EOF_EofCreateWithTruncatedContainer",
```
shemnon, line 44: I do not see how we can detect this at validation time.  EOF Create with a truncated container should only fire if auxdata is insufficient, and that could change at runtime.  Although in this instance it is staticly true, we would have to add symbolic execution to the spec to prove it.

```
010004 # Types length ( 4 )
020001 # Total code sections ( 1 )
  000c # Code section 0 , 12 bytes
030001 # Total subcontainers ( 1 )
  0016 # Sub container 0, 22 byte
040000 # Data section length(  0 )
    00 # Terminator (end of header)
       # Code 0 types
    00 # 0 inputs 
    80 # 0 outputs  (Non-returning function)
  0004 # max stack:  4
       # Code section 0
  6000 # [0] PUSH1(0)
  60ff # [2] PUSH1(-1)
  6000 # [4] PUSH1(0)
  6000 # [6] PUSH1(0)
  ec00 # [8] EOFCREATE(0)
    50 # [10] POP
    00 # [11] STOP
           # Subcontainer 0 starts here
    ef0001 # Magic and Version ( 1 )
    010004 # Types length ( 4 )
    020001 # Total code sections ( 1 )
      0001 # Code section 0 , 1 bytes
    040003 # Data section length(  3 ) (actual size 2)    
        00 # Terminator (end of header)
           # Code 0 types
        00 # 0 inputs 
        80 # 0 outputs  (Non-returning function)
      0000 # max stack:  0
           # Code section 0
        fe # [0] INVALID
      # Data section length ( 3 ) actual length ( 2 )
      aabb
           # Subcontainer 0 ends
       # Data section (empty)
       ```

File: EOFTests/eof_validation/stack/exchange_stack_validation.json [hunk 4]
```diff
@@ -0,0 +1,201 @@
...
+      },
+      "exchange_stack_validation_11": {
+        "code": "0xef00010100040200010017040000000080000a6001600160016001600160016001600160016001e86100",
+        "results": {
+          "Prague": {
+            "result": true
+          }
+        }
+      },
+      "exchange_stack_validation_12": {
```
shemnon, line 36: Besu has exchange_stack_validation_12 through exchange_stack_validation_18 as all valid.  For each case there are 10 items on the stach, and the excanges are (9, 9), (1, 9), (8, 9) (2, 9), (5, 9), (6, 9), and (4, 9).  All work with 10 stack items.

EOFTests/eof_validation/EOF1_embedded_container.json
```diff
@@ -0,0 +1,55 @@
...
+      "EOF1_embedded_container_0": {
+        "code": "0xef00010100040200010006030001001404000000008000016000e0000000ef000101000402000100010400000000800000fe",
+        "results": {
+          "Prague": {
+            "result": true
+          }
+        }
+      },
+      "EOF1_embedded_container_1": {
+        "code": "0xef00010100040200010006030001001404000200008000016000e0000000ef000101000402000100010400000000800000fe",
```
shemnon, line 13: The top level container is missing 2 data bytes.  Same issue I ahd with a prior test, do these represent deployed contracts?  i.e. the ones EOFCREATE / TXCREATE would see?

shemnon:
Other than the issues I have listed (incomplete data at root container and exchange stack depth) Besu agrees with these tests.

EOFTests/eof_validation/EOF1_eofcreate_invalid.json
gumb0: Auxdata is appended to deploy container, it's not relevant here. 

Initcontainer (the one targeted by EOFCREATE) must have actual data size equal to declared size, otherwise this EOFCREATE is invalid.

In the spec:

> `EOFCREATE` the subcontainer pointed to by initcontainer_index must have its `len(data_section)` equal `data_size`, i.e. data section content is exactly as the size declared in the header (see [Data section lifecycle](https://github.com/ipsilon/eof/blob/main/spec/eof.md#data-section-lifecycle))

EOFTests/eof_validation/EOF1_truncated_section.json
gumb0: We consider these tests to run general validation that is applied to both deployed and initcontainers, so truncated is allowed.

We apply additional check for full data section at the point of EOFCREATE validation, as I described in the comment below, and at the point of TXCREATE execution for its target top-level container.

EOFTests/eof_validation/EOF1_eofcreate_invalid.json
shemnon: First, I didn't consider these may be the inputs in an InitContainer section, subject to AUXDATA from a TXCREATE opcode.  I was focused on contract creation transactions which have no such auxdata.

The spec appears to be the opposite of what was intended and needed then for EOFCREATE.  Consider a swap contract that includes the two contract addresses loaded by DATALOADN.  Those would need to point into the area that would be auxdata.  So for validation the subcontainer would need to have a data section length that would allow for the DATALOADN operation to validate (current length + 64 at least) but could not be set yet.  So the subcontainer can be "short" but after aux data is applied must not be "short" but can grow because of aux data.

gumb0: I think maybe you're missing one level of creation process nesting.

During creation this swap contract will be a subcontainer of initcontainer, this subcontainer is targeted by RETURNCONTRACT, truncated data is allowed for it, and is appended at run-time. Parent initcontainer is targeted by outer factory container EOFCREATE, truncated data is not allowed for it. 

Factory gets those constructor arguments from calldata probably, passes them to EOFCREATE, inside initcontainer execution they are forwarded to RETURNCONTRACT, and appended to deploy container data section.

EOFTests/eof_validation/stack/exchange_stack_validation.json
shemnon: The correct action may be to change the spec of EXCHANGE, so it swaps `n+2` and `n+m+3`, where 1 is the top of the stack.  So `e900` would swap the second and third item, and would also result in no combos such as (9,9) which are pointless.  This would prevent any operation from swapping with the top of the stack, which SWAPN covers, as well as swapping the same two items.

With that re-definition the pairs are (10, 11), (2, 11), (9, 11), (3, 11), (6, 11), (7, 11), and (5, 11), all of which bust the stack and make these tests valid.

winsvega:
Is this ready to be converted from the draft ? 
please make a filler version or copy all the files as is with Copier.json extension. 
this way we won't loose or forget it because tests are tracked. 
Copier.json means the test will be copied from src folder to the generated test folder as is (Copier suffix will be removed from its name)

EOFTests/eof_validation/stack/exchange_stack_validation.json
pdobacz: Per megaspec:

```
    - read uint8 operand `imm`
    - `n = imm >> 4 + 1`, `m = imm & 0x0F + 1`
    - `n + 1`th stack item is swapped with `n + m + 1`th stack item (1-based).
    - Stack validation: `stack_height >= n + m + 1`
```

This first case `imm = 0x80, n = 9, m = 1`, so the swap pair is (10, 11) like you wrote (one based), and should underflow.

EOFTests/eof_validation/EOF1_embedded_container.json
pdobacz: Andrei's out for 2 days, I'll try to respond, even though I'm not 100% sure. The test looks ok and represents a not-yet-deployed contract. The description in the generator is:

```
    // no data section in container, but anticipated aux_data
```

Such container can be an argument to RETURNCONTRACT, provided you give it 2 bytes of aux_data, and cannot be an argument to EOF/TXCREATE.

EOFTests/eof_validation/stack/exchange_stack_validation.json
shemnon: Ok, that every term has a +1 is where my brain took shortcuts and presumed eveything had an off by one error.  they should be folded into one final use.

EOFTests/eof_validation/EOF1_eofcreate_invalid.json
shemnon: These two statements said different things...

> Initcontainer (the one targeted by EOFCREATE) must have actual data size equal to declared size, otherwise this EOFCREATE is invalid.

> During creation this swap contract will be a subcontainer of initcontainer, this subcontainer is targeted by RETURNCONTRACT, truncated data is allowed for it

The latter statement is what I was advocating for, so I will presume the first statement was made in error.  Even though it is different ops the InitContainer will be using AuxData to communicate the needed contract IDs as calldata may not be available as this may be a deeply nested create and not a top level create.  So initcode will need to be truncated similar to the deployable subcontainer.

pdobacz: I think what Andrei wrote is correct, give me a moment I'll prepare an example to demonstrate this on the swap contract example

pdobacz: @shemnon posted here, let me know if this clarifies things https://gist.github.com/pdobacz/27930c535fe891e51e77aa1baeee1572

EOFTests/eof_validation/EOF1_embedded_container.json
shemnon: Given that top level EOF will need to be validated with truncated data, this seems to be a valid test then.  What we will need are execution tests where return data has truncated, exact, and excess data for both EOFCREATE and TXCREATE.

shemnon:
Can this also be rebased off of main, or at least after https://github.com/ethereum/tests/commit/5f77c442bf862276e72f67084bdf143c7b24db65 ?  There were some issues with Cancun tests fixed by that PR.

EOFTests/eof_validation/EOF1_eofcreate_invalid.json
shemnon: I think I see the disconnect.  Neither megaeof.md nor RIP-7620 specify what `data_offset` and `data_size` are used for.  The worked example drew that out:

* I had presumed they had the same semantics as `aux_data_offset` and `aux_data_size` and that the subcontainers are re-written with the data appended.
* Based on your example it appears that the data is copied from memory into calldata.

Neither one of these interpertations have any basis from either spec other than conventions of names.  Something like this needs to be added to https://eips.ethereum.org/EIPS/eip-7620#initcontainer-execution

> * Memory starting at `data_offset` of `data_size` length is copied into the call data

We also need verbiage in the opcode sections stating that 

> * Deduct gas for possible memory expansion for data starting at `data_offset` and going for `data_size` length

shemnon:
Besu is now in line with the tests in this PR

EOFTests/eof_validation/EOF1_eofcreate_invalid.json
pdobacz: Ah, yes, indeed. I think this was skipped in megaspec, because it tacitly extrapolates CREATE/CREATE2, and then copied like that into 7620. Thanks for pointing this out and sorry about the confusion :bow: 

Definitely worth emphasizing, that **neither** EOFCREATE **nor** TXCREATE rewrite the initcontainers they reference. They only do so on deployed containers referenced on RETURNCONTRACT they end with. So `data_offset` & `data_size` from EOFCREATE & TXCREATE are a very different beast from `aux_data_offset` & `aux_data_size` from RETURNCONTRACT.

To make this distinction clear, let's use `input_offset` and `input_size` for EOFCREATE / TXCREATE, which coincides with EXT*CALL naming from EIP-7069, and not with `aux_data` stuff.

EOFTests/eof_validation/EOF1_embedded_container.json
pdobacz: > Given that top level EOF will need to be validated with truncated data, this seems to be a valid test then. What we will need are execution tests where return data has truncated, exact, and excess data for both EOFCREATE and TXCREATE.

I see this comment predates the one you've posted about the disconnect. Has it become obsolete now? If not, I'm not sure what kind of tests you're proposing, what do you mean by "return data" in this context?

winsvega:
moved to this branch https://github.com/ethereum/tests/pull/1364

PR #1344, opened 2024-02-21
fix cancun beacon contract

winsvega:


winsvega:
ok so from EIP it looks like 

```
if no code exists at BEACON_ROOTS_ADDRESS, the call must fail silently
it kind of looks like having the state without beacon contract is valid too
```

so regardless if you make a call or write storages to non existent account. 
the result state will have no accounts as empty account (having only storage) are not allowed and removed.
so we dont need to put this contract in all of the tests

PR #1346, opened 2024-02-22
Fix typos

rex4539:


rex4539:
@winsvega Could you please take a look?

winsvega:
hm. can you split it to multiple PRs? 
1) docs and legacy tests (the one that does not need a refill)
2) EIPTests  (tricky one beacuse to refill it need banches of WIP )
3) EOFTests  
4) BC/State tests that need a refill.

rex4539:
I have no idea how to do this as I'm not very familiar with the git command line args.

If you can type out the exact commands I need to type, I would be happy to do it right now.

rex4539:
That's what I figured.

PR #1349, opened 2024-03-05
Fix and refill EOF Tests according to default retesteth configuration

hugo-dc:
Fix and refill EOF Tests according to default retesteth configuration proposed in this PR: https://github.com/ethereum/retesteth/pull/225


gumb0 (approved):

gumb0:
@winsvega This looks good to go, please check.

PR #1350, opened 2024-03-06
 Update EOF1_Calls to EXT*CALL from EIP-7069

pdobacz:
With https://github.com/ethereum/evmone/pull/630, EOF will call using revamped EXT*CALL instructions, so here we adapt the tests to use them

- gas limit bumped because the 1/64th of it wasn't enough for the SSTORE instructions, in case the callee aborted
- 32 bytes returned from EOF call targets in order to use RETURNDATALOAD conveniently (instead of output buffer)

https://github.com/ethereum/tests/pull/1349 needs to be merged first


gumb0:
Can you rebase this? My stack validation update commits were merged into develop.

pdobacz:
closed in favor of https://github.com/ethereum/tests/pull/1349

PR #1358, opened 2024-03-20
Add more ecpairing test vectors

chfast:


winsvega:
please consider using .py format or make a test into pyspecs directly as it is unlikely that there will be retesteth support for the next fork

chfast:
Ok. I could've submitted it there if I knew.

PR #1384, opened 2024-06-13
fix: wrong json format issue

forcodedancing:
`# noqa: E501",` is not invalid in the json files.


IaroslavMazur:
Well done, @forcodedancing!

winsvega:
hmm, I didn't catch this with my light json parser.

winsvega:
ok the issue was in my parser. should be fixed now

PR #1390, opened 2024-06-21
fix: file names and references

IaroslavMazur:
Fixed the incorrect file names, as well as the references to them inside the other files in the repo.


winsvega:
that require test regeneration. do you want me to do it?

IaroslavMazur:
> that require test regeneration. do you want me to do it? 

At this point, I don't (yet) know to do that, so, yes, please!

Also, it'd be awesome if you could briefly explain:
- how the regeneration should be done
- and why it's required in this case, in the first place 🙏🏻

winsvega:
in this repo I track that test source  filler files were used to generate the result test files. 
if you change the filler, hash verification will fail.

winsvega:
https://github.com/ethereum/tests/pull/1393

IaroslavMazur:
> in this repo I track that test source filler files were used to generate the result test files. if you change the filler, hash verification will fail.

Thank you for regenerating the tests! 🙏

PR #1395, opened 2024-07-29
EOF: remove tests duplicated in EEST

pdobacz:
https://github.com/ethereum/execution-spec-tests/pull/571 added tests which seem to be exactly the same as tests removed here. I'm not sure why we didn't remove back then, but I stumbled upon this and figure maybe we can remove now.

CC @shemnon


shemnon:
These were ported over from the filler into EEST, prior to the existing old-style tests were brought into the repo.  So it was a bit of a timing issue.

shemnon (approved):

chfast (approved):

PR #1414, opened 2024-11-06
Correct comments

### LINKED ISSUE #1407: Misleading comments in test filler
https://github.com/ethereum/tests/blob/9201075490807f58811078e9bb5ec895b4ac01a5/src/GeneralStateTestsFiller/VMTests/vmTests/calldatacopyFiller.yml#L171

The comments claim a 0x10 byte long value is going into memory, but then the value is 0x11 bytes. I think this may be related to another misleading comment earlier that says an mload should produce 0 when it in fact produces 0x12.

dzhalaevd:
### What was wrong?
Misleading comments in test fille

Related to Issue: #1407  
Closes: #1407


winsvega:
need now to regenerate the test so it is updated with the filler hash

winsvega:
https://github.com/ethereum/tests/pull/1425

PR #1420, opened 2024-12-13
Fix returndatacopy_after_failing_delegatecallFiller

### LINKED ISSUE #1417: Invalid `to` in  returndatacopy_after_failing_delegatecallFiller
This line:

https://github.com/ethereum/tests/blob/a0e848269dd2af2b061fe9a4749d40963c2bdb5c/src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_after_failing_delegatecallFiller.json#L67C21-L67C63

should use `0x1000000000000000000000000000000000000001` as `to` address.

0xVolosnikov:
Fixes https://github.com/ethereum/tests/issues/1417

### DIFF
--- a/src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_after_failing_delegatecallFiller.json
+++ b/src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_after_failing_delegatecallFiller.json
@@ -8,12 +8,6 @@
             "currentTimestamp" : "0x03e8"
         },
         "pre" : {
-            "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6" : {
-                "balance": "0x100000",
-                "code": "",
-                "nonce": "0x00",
-                "storage": {}
-            },
             "0x1000000000000000000000000000000000000001": {
                 "balance" : "0x00",
                 "code": "{ (DELEGATECALL 10000 0x1000000000000000000000000000000000000002 0 0 0 0) (RETURNDATACOPY 0x0 0x0 32) ( SSTORE 0 (MLOAD 0))}",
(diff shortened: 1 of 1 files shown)


winsvega:
rebase
need to refill the tests and I would guess then expect section needs a correction as well. 
thanks!

winsvega:
or can just plain convert it into pyspecs

src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_after_failing_delegatecallFiller.json
```diff
@@ -8,12 +8,6 @@
             "currentTimestamp" : "0x03e8"
         },
         "pre" : {
-            "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6" : {
```
chfast, line 11: What was this account for and why have you removed it?

winsvega: the test apparently had a typo. and was calling this empty account instead of account with the actual code

chfast: I understand the main bug, but not sure why this account was in the test.

chfast:
This also affects legacy tests.

src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_after_failing_delegatecallFiller.json
winsvega: copy paste error. the test was not doing anything

src/GeneralStateTestsFiller/stReturnDataTest/returndatacopy_after_failing_delegatecallFiller.json
```diff
@@ -8,12 +8,6 @@
...
-                "balance": "0x100000",
-                "code": "",
```
winsvega, line 13: so I think here would have been the call to 100...1 which fails. and then sstore[1] = 1 to indicate that the test worked

winsvega:
`Detected error: return data out of bounds` this is the vm error on such tests. 
I assume if the test is incorrect we can as well just remove it from legacy tests too. 
when we port on python it should be better checked.

PR #1422, opened 2024-12-24
Use new EIP2537 tests

Marchhill:
Use new tests from devnet-5 release https://github.com/ethereum/execution-spec-tests/releases/tag/pectra-devnet-5%40v1.0.0


Marchhill:
Realised these shouldn't be changed directly

PR #1425, opened 2025-01-14
remove calldatacopy tests, they have been moved to EEST

pacrob:
The calldatacopy tests have been moved to EEST here:

https://github.com/ethereum/execution-spec-tests/pull/1056

so should be removed.


winsvega:
blockchain version of the test needs to be removed as well @pacrob

pacrob:
> blockchain version of the test needs to be removed as well @pacrob

Got it, thanks! Anything else needed here?

PR #1431, opened 2025-01-22
Remove migrated precompiled blake2B Tests

reedsa:
Precompiled Blake2B Tests:
- https://github.com/ethereum/execution-spec-tests/pull/1067


reedsa:
@winsvega This should cover the cases, I assume these should be merged after the conversions are complete?

winsvega:
Yes. This way, we migrate old style tests

PR #1443, opened 2025-02-05
fix: removed dead links, fixed typos, etc.

felix314159:
removed dead links (replaced them with working links when possible), fixed typos and other formatting issues


docs/retesteth-install.rst
```diff
@@ -12,7 +12,7 @@ Linux running anywhere else with an Internet connection.
 
       sudo apt install -y wget docker docker.io
 
-#. Download the latest **retesteth** `docker image <https://retesteth.ethdevops.io/dretesteth.tar>`_
```
winsvega, line 15: not https. the link exist

felix314159: Yes the link works, but when I click the link in the browser I get 'unable to connect' and since the next line shows you the command to download the file anyways I felt like it was redundant to have the text hrefd.

winsvega: the link works without unable to connect, the download works. need to remove https to http
it is not dead

felix314159: Ah, I see now what you mean. If you want we can keep it as http, but we can also consider that browsers these days are likely to show warnings due to https-only modes etc. I have no strong opinion about this

PR #1467, opened 2025-03-11
remove pyspec tests to avoid confusion

winsvega:
currently pyspec vectors exist in 2 versions. in this repo and in eest releases which rises confusion to which are the latest vectors to run and trust. and Prague version of it is not supported by retesteth. 

to avoid confusion I remove all this vectors from here. as we migrate the tests to eest.
please refer to pyspecs releases here: https://github.com/ethereum/execution-spec-tests/releases
 
the reason to have it here was to provide .json files to everyone from a repo rather than tar ball
and to verify the validity of .json schemas generated by pyspecs.


spencer-tb (approved):
LGTM! Purge away

